PR #777, opened 2014-07-30
sys-kernel: enable selinux support

gregkh:
Note, this enables SELinux automatically, in "permissive" mode, so things will still work properly.  You need to then "disable" SELinux in the boot sequence to keep things "fast".  We can't do it the other way around (default to off, and then enable after booting) as SELinux will not work that way (selinuxfs is not present, other things do not work, etc.)  This is by design...

Only merge this if you really want this, it should obsolete pull request #642


ramschmaerchen:
Don't want to be pushy, but having SE-Linux is crucial to survive in this highly competitive field of "micro OS"s.

gregkh:
@ramschmaerchen what do you mean by "micro OS's"?  and how does SELinux pertain to it?  Do you have a specific SELinux profile for your application that you want to use today?

dksh:
> Do you have SELinux rules for your application / container already written today?
No but I anticipate needing to write one eventually due to PCI DSS requirements.
Is having a sample rule set necessary for implementing SELinux in CoreOS?

> what do you mean by "micro OS's" and how does SELinux pertain to it?
I might be wrong but I think he means CoreOS and other minimal operating systems like it where the OS is so small that not even a working SELinux comes with it. This is probably the security advantage of a minimal install of a traditional Linux flavor has over CoreOS at the moment.=

adrianrogers:
I've been following this repo specifically to find out when SELinux support happens for CoreOS.

gregkh:
@dksh if you don't have any rule set for your application, then SELinux support in the base OS isn't going to do anything.  I wasn't aware that PCI requirements now required SELinux, do you have a pointer to that requirement somewhere I could read about it?

ramschmaerchen:
No need for my application to support SElinux. Look at how project atomic solves it. SELinux wraps around docker. That's how it should be done. No need for "my" application to support SELinux. Even Canonical was able to get en par with Project Atomic by pushing apparmor into their Ubuntu Core. Docker does not contain the container good enough, so there is indeed a huge need for SE-Linux or similar.

It is bad news if the attacker is able to reach outside your container.

Links:
http://www.projectatomic.io
http://www.ubuntu.com/cloud/tools/snappy

dksh:
> @dksh if you don't have any rule set for your application, then SELinux support in the base OS isn't going to do anything
As ramsch says, the solution which the Atomic project employs is what I'm going for, in which you enforce SELinux at the OS<->Container layer as opposed to the Container<->Application layer.

> wasn't aware that PCI requirements now required SELinux, do you have a pointer to that requirement somewhere I could read about it?
They aren't currently required, but I'm predicting that SELinux will become a requirement for those who want to use container technology in PCI environments.

PierreKircher:
doubt it .. may be for pci dss level 1 .. but that brings a other set of problems with it  and who wants to store full cc data unless you are a financial services agency

dksh:
> who wants to store full cc data unless you are a financial services agency
Telcos, e-commerce corporations, and any other enterprise that sets up an "ecosystem" like Apple, Google etc.

Working in an enterprise like those described above, I want to keep my OS configuration as streamlined as possible. If I can provide satisfactory security with CoreOS then I can have all servers use the same setup. Until then, I have to say no to anybody who wants to use containers even if they don't need PCI level security because it costs too much of my time to manage exceptions.

PierreKircher:
security comes in layers .. it always will .. SELinux is just 1 part of it.. and pci will NOT enforce a specific software

adrianrogers:
PCI is a set of guidelines and "best practices" that one has to comply
with. It is important in its own right, but it isn't the main reason why SE
Linux should be available in CoreOS.

I think the security and privacy issues of the last few years highlights
the necessity for another "layer" in how we implement security.
Specifically, Docker/LXC containers and CoreOS are relatively new tech.
Much is yet to be discovered or reported by those who spend their day job
hacking software to find bugs (a number of my friends who do this sort of
thing simply haven't gotten to docker yet for instance).

I've noticed the trend with containers to start applying SE Linux, amongst
other things, to secure containers on the host OS level. The most reason I
can see for this tactic is to limit the reach of anything "going rogue" or
heaven forbid a "red pill" type vulnerability that allows for arbitrary
processes to break out of the containers and execute code in the host
environment or in other containers. Something like SE Linux would allow us
to limit the fallout from such a scenario since it can enforce mandatory
access. Even network interfaces can be locked down.

The point is...we don't fully understand the true attack surface of
containers...at least not in the way that we do other things that have been
broken and fixed repeatedly. This is simply what happens with new tech.
Some of us want to have the benefits of the new tech while limiting our
exposure to the unknown risks that we believe are inherent in their use.
That's ultimately one of the reasons why SE Linux was built in the first
place. To hedge against certain unknowns behaving in unknown ways on your
system.

It isn't the only tool we have, but it is important. There is just cause
for SE Linux support in CoreOS. And as I recall, user interest was a big
factor in continuing SE Linux development for CoreOS.

I hope the number of posts in this thread (and the length of mine) have
gotten the point across.

We are interested in SE Linux support. We want it. Badly.

jhulten:
I want to use CoreOS as the base OS for the company I am working for. But without SELinux I expect I will end up using Atomic because I only have so much fight in me and taking on security folks who know a perfectly good way to secure things that is an option elsewhere is not in my best interests.

+1 SELinux, and soon...

stela5:
FYI, https://developer.ubuntu.com/en/snappy/ has AppArmor.  It might be easier to implement than SELinux...  https://www.suse.com/support/security/apparmor/features/selinux_comparison.html

c4milo:
+1 for using apparmor instead.

Tonnu:
I'm also in favour of adding SELinux to CoreOS. Which other tools are there in place now to help harden docker containers?

marineam:
Superseded by https://github.com/coreos/coreos-overlay/pull/1311

PR #791, opened 2014-08-09
sys-apps/rng-tools: Add rng-tools

crawford:
Pulled the systemd unit into rng-tools-4-r5.ebuild and dropped the confd and initd files.


mischief:
any reason not to refresh and merge this?

crawford:
Depends on https://github.com/coreos/portage-stable/pull/204.

crawford:
This will _include_ rng-tools in the image, but it will not enable it by default. We will need a systemd-generator that detects if a TMP is present, /dev/hwrng is present, or the CPU supports RDRAND and enables the service.

marineam:
[LGTM](http://lgtm.in/i/etjbIlrCn)

PR #816, opened 2014-08-23
Add systemd-docker ebuild

bcwaldon:


bassamtabbara:
Any update on this issue? 

I think CoreOS is likely the most affected by the friction between systemd and docker. Starting docker containers from systemd just doesn't work well right now.

+1 for adding systemd-docker to CoreOS as a workaround until this is fixed in a better way. 

CoreOS team, what are you thoughts on this?

PR #844, opened 2014-09-08
Use --convert-netconf=debian when using Openstack config-drive

dmsimard:
When booting CoreOS in an Openstack environment with config-drive,
nova pushes a debian-like network interface configuration file for
cloud-init to read. See:
https://github.com/openstack/nova/blob/master/nova/virt/interfaces.template

This commit modifies the Openstack overlay behavior by adding
--convert-netconf=debian to the cloudinit parameters.


dmsimard:
Just recording that I've also posted on the mailing list about this with all the details: https://groups.google.com/forum/#!topic/coreos-user/ui35wEqqx0E

PR #855, opened 2014-09-16
Added glusterfs

asiragusa:
This adds glusterfs 3.4.2 to CoreOS


Inflatablewoman:
Any news?  Or is this going to be dropped?

crawford:
We've decided against including GlusterFS in the base CoreOS image. That being said, it is relatively easy to get GlusterFS running in a container. Its a similar situation with Ceph. We include the LKMs but the userland tools must be included in a container.

asiragusa:
Alex if many people ask for GlusterFS in CoreOS is because we need to access the DFS via the host system and actually Docker doesn't allow to do that. Why to impose such limitation?

crawford:
@asiragusa We definately don't want to prevent people from running GlusterFS on CoreOS. If it's not possible to run GlusterFS effectively in a container, then there is a problem. What makes you think that Docker wouldn't be capable of running GlusterFS? I have seen some demos of it working.

dw33z1lP:
@crawford  You can use Gluster in a container but can't use a DFS (NFS and GlusterFS) against a --volume-from that is annoying a lot of people.

See https://github.com/docker/docker/issues/4213 and https://github.com/docker/docker/issues/8152 .

asiragusa:
@crawford the problem is not running GlusterFS inside a container, the problem is accessing the files from the host machine. There are heaps of issues opened on the Docker Git because of that.

crawford:
@dw33z1lP It looks like this behavior is a regression in Docker. Let me dig around and see if I can make this work.

@asiragusa Although I haven't tested it myself, according to https://github.com/docker/docker/issues/8152#issue-43335048, it sounds like you should be able to access the files from the host system just fine. The real issue, as @dw33z1lP pointed out, in the --volume-from bug.

dw33z1lP:
@crawford Yes, @cpuguy83 make it work in 0.9 max I think.

asiragusa:
@crawford sorry for the late answer. If you read carefully, the -v flag allows to export an host directory into the container, not the opposite, thus if you do -v /somepath:/shared and then you mount the DFS on the container into /shared or /shared/somesubpath it will be empty into the host.

In few words: it's not possible to share the mount namespace between a container and the host machine, thus there is no way for the host to access the DFS mounted by the container. This behavior is the same for every DFS, you can test it even with NFS.

Can you reopen this issue please?

cpuguy83:
You should actually be able to mount nfs/gluster/<whatever> from the host into the container since Docker is using recursive bind mounting.

However, container-to-container is a little more tricky.

asiragusa:
@cpuguy83 Are you able to give a working example? If I list the mount points, then a docker run gluster, then I list the mount points again I can't see any difference.

By the way this would be, by far, the best solution.

asiragusa:
@cpuguy83 according to https://github.com/torvalds/linux/blob/f96fe225677b3efb74346ebd56fafe3997b02afa/Documentation/filesystems/sharedsubtree.txt I can do the following:
```
$ mkdir a
$ mkdir b
$ sudo mount --bind a a
$ sudo mount --make-rshared a
$ mkdir a/test
$ sudo mount --bind a/test b
$ touch b/ok
$ ls a
test
$ ls a/test
ok
$ docker run -d -v ~/a:/a --name mount ubuntu
5b48f1c2e4fb91a73e9869d0fd0bfaeb32673f22ab24607d2ff4c1259b11dceb
$ docker exec -ti mount /bin/bash
# ls /a
test
# ls /a/test
ok
```

So basically I can make the directory a recursively shared and any mount to its subtree can be seen from inside the container, but if I do the following, continuing the example:

```
# mkdir /a/test2
# mount --bind /usr /a/test2
# mount --make-rshared /a/test2 # has no effect
# mount --make-rshared / # no effect either
# ls /a/test2
bin  games  include  lib  local  sbin  share  src
# exit
$ ls a/test2
$
````

The directory a/test2 is empty. I guess that I can't see the contents of the bind mount in a/test2 in the host because docker does a mount --make-rslave when mounting the root image filesystem https://github.com/torvalds/linux/blob/f96fe225677b3efb74346ebd56fafe3997b02afa/Documentation/filesystems/sharedsubtree.txt#L182. If I'm right, I'll hardly circumvent this problem.

Any advice on this?

Thanks

cpuguy83:
@asiragusa If you have a mount on the host, you can bind-mount it into the container.
You can't have a mount in a container, and share it over a volume since the mount lives in the container, and you'd be writing to the host volume dir.

cpuguy83:
btw, I think the solution to this is somewhat discussed in #8484, where there will be more than just the two volume options we have today (manual bind-mounts volumes and "normal" volumes), and provide a strategy for providing a volume which is backed by gluster, or ceph, or whatever.

pcm32:
We really need GlusterFS support on CoreOS, to be able to use Kubernetes Persistent Volume support for GlusterFS. It requires nodes (running CoreOS in our case) to be able to mount GlusterFS drives. So if at least the gluster client could be made available, it would be great. We currently mount Gluster through NFS in CoreOS (for the Kubernetes Persistent Volume), but it is a pain, because it gives a single point of failure (NFS connection is done to one of the gluster nodes, compared to Gluster direct which connects to the pool of nodes).

I haven't found a way to mount GlusterFS from CoreOS nodes directly (to be able to use the Kubernetes PV for Gluster), but if someone could point me to a way to be able to do this, I would really appreciate it. Thanks!

shenhequnying:
hello:
    if you want to mount glusterfs(fuse) in coreos, you can try this
https://hub.docker.com/r/jsafrane/glusterfs-mounter/
    i have try it can run.

2016-08-21 21:06 GMT+08:00 Pablo Moreno <notifications@github.com>:

pcm32:
Thanks @shenhequnying. Looks like an interesting idea, yet it requires recompiling of kubelet :-(, at least a year ago. Do you know if this is still the case?

PR #914, opened 2014-10-11
coreos-kernel: enable X32 ABI

errordeveloper:
This only enables kernel support for the [X32 ABI](http://lwn.net/Articles/456731/), which is a useful optimisation. A program compiled to this ABI will run in native 64-bit mode, but with 32-bit pointers and data values.


martynd:
Id like to reopen the case for enabling the x32 ABI. 

We actively utilize it in a custom CoreOS build (with no other changes than having CONFIG_X86_32 set and increment the ebuild number for the kernel)

I have seen savings over the 40% memory usage described previosuly, especially in processes which use many pointers such as Redis (They even actively encourage using 32bit builds where possible for this exact reason http://redis.io/topics/faq#what39s-the-redis-memory-footprint)

Especially given this and how CoreOS is designed with containers in mind, I think it is a good fit as it further refines the ability to optimally use resources.

Understandably you would likely not want to provide a full set of binaries to support this on the host machine as this can close to double the libraries you need to have built for CoreOS users.

With that said, simply enabling it on the kernel is enough to allow anyone who has gone to that effort (or has downloaded a working x32 toolchain) to utilize it within containers

PR #983, opened 2014-12-04
sys-apps/systemd: mark /var/log/journal as no copy-on-write

soellman:
This should set the nocow flag on the journald log dir (/var/log/journal) as per https://groups.google.com/forum/#!topic/coreos-dev/X4RyCEuHp18

### DIFF
--- a/sys-apps/systemd/systemd-215-r16.ebuild
+++ b/sys-apps/systemd/systemd-215-r16.ebuild
@@ -491,6 +491,9 @@ pkg_postinst() {
 	# Migrate 80-net-name-slot.rules -> 80-net-setup-link.rules
 	migrate_net_name_slot
 
+	# Mark journald log directory as no copy-on-write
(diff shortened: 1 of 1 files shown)


daurnimator:
This was fixed in the systemd release today (219)

From http://lists.freedesktop.org/archives/systemd-devel/2015-February/028447.html
> journald now sets the special FS_NOCOW file flag for its
          journal files. This should improve performance on btrfs, by
          avoiding heavy fragmentation when journald's write-pattern
          is used on COW file systems. It degrades btrfs' data
          integrity guarantees for the files to the same levels as for
          ext3/ext4 however. This should be OK though as journald does
          its own data integrity checks and all its objects are
          checksummed on disk. Also, journald should handle btrfs disk
          full events a lot more gracefully now, by processing SIGBUS
          errors, and not relying on fallocate() anymore.

PR #1037, opened 2015-01-01
systemd: copy symlink-usr handling from old systemd-sysv-utils ebuild

marineam:
While moving from systemd-sysv-utils to systemd w/ USE=sysv-utils I
forgot that USE=symlink-usr needed special handling to ensure the
symlinks were created correctly.


crawford:
LGTM

PR #1038, opened 2015-01-01
Cleanup duplicate packages in coreos-overlay

marineam:
Depends on https://github.com/coreos/portage-stable/pull/162


crawford:
LGTM

PR #1044, opened 2015-01-05
docker: fix flag parsing in docker wrapper script

marineam:
The dockerd script incorrectly parsed --graph=/some/path, breaking the
early-docker service.

Fixes https://github.com/coreos/bugs/issues/229


eyakubovich:
LGTM. Tested and it works.

PR #1046, opened 2015-01-06
update ccache, coreos-install, motdgen

marineam:
The current ccache has some sort of locking bug that causes build
failures when the cache directory is being created for the first time.
This typically results in the first build in a fresh SDK failing.

Depends on https://github.com/coreos/portage-stable/pull/163


vcaputo:
Late to the party but LGTM

marineam:
Thanks :)

PR #1053, opened 2015-01-13
app-emulation/docker: drop insecure registry flag

crawford:
Additionally, prune Docker 1.3.3 and it's specific service files.


marineam:
Could you split the 1.3.3 prune and flag change into two commits please?

crawford:
Done

marineam:
Looks like a rebase got mixed up but otherwise LGTM

PR #1055, opened 2015-01-16
docs: include docs for coreos tools

crawford:
babby's first eclass


crawford:
Related https://github.com/coreos/scripts/pull/374

eclass/coreos-doc.eclass
```diff
@@ -0,0 +1,29 @@
+# Copyright 2015 CoreOS, Inc.
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+# @ECLASS: coreos-doc.eclass
+# @BLURB: utility function for including documentation
+
+# @FUNCTION: dodoc-coreos
+# @USAGE: [-r] <list of docs>
+dodoc-coreos() {
```
marineam: nit but coreos as a prefix would be more inline with typical namespacing than a suffix.

crawford: ok, the random eclass that I looked at used it as a suffix. It _was_ PHP though...

marineam: Well *there's* yr problem! ;-)

eclass/coreos-doc.eclass
```diff
@@ -0,0 +1,29 @@
...
+		flags="-r"
+		shift
+	fi
+
+	[[ "${#}" -lt 1 ]] && die "$0: at least one file needed"
+
+	insinto "/usr/share/coreos/doc/${P}/"
+	for f in "${@}" ; do
+		if [[ -s "${f}" ]] ; then
+			doins "${flags}" "${f}"
```
marineam: without '-r' this will pass an empty string "" as the first argument. That seems not quite right but may be harmless. If you remove the quotes an empty string in ${flags} will not become an argument.

eclass/coreos-doc.eclass
```diff
@@ -0,0 +1,23 @@
+# Copyright 2015 CoreOS, Inc.
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+# @ECLASS: coreos-doc.eclass
+# @BLURB: utility function for including documentation
+
+# @FUNCTION: coreos-dodoc
+# @USAGE: [-r] <list of docs>
+coreos-dodoc() {
```
crawford: prefix instead of suffix

eclass/coreos-doc.eclass
```diff
@@ -0,0 +1,23 @@
...
+	local flags
+	if [[ "${1}" == "-r" ]] ; then
+		flags="-r"
+		shift
+	fi
+
+	[[ "${#}" -lt 1 ]] && die "${0}: at least one file needed"
+
+	insinto "/usr/share/coreos/doc/${P}/"
+	doins $flags $@
```
crawford: simplified this (no need to loop over the files if doins is going to do it anyway)

marineam:
[LGTM](http://www.lgtm.in/i/eLQVBNNSp)

PR #1061, opened 2015-01-21
coreos: add ebtables

marineam:
Open the possibility of implementing flannel security features with a
security mechanism instead of creative configuration hacks. ;-)

Depends on https://github.com/coreos/portage-stable/pull/167


eyakubovich:
LGTM

PR #1062, opened 2015-01-21
systemd: Accept DHCP domains with a trailing dot

marineam:
GCE includes the BIND-ism trailing '.' at the end of the domain name
reported in DHCP leases. After 215 systemd started rejecting this.
Fixes a regression compared to the current CoreOS stable release.

Fixes https://github.com/coreos/bugs/issues/220


crawford:
LGTM

PR #1065, opened 2015-01-24
overlay: add rocket

polvi:
this is just an alpha for testing. Need rocket to mature more before having proper ebuild.

### DIFF
--- a/app-emulation/rocket/rocket-0.2.0-r1.ebuild
+++ b/app-emulation/rocket/rocket-0.2.0-r1.ebuild
@@ -1,0 +1,1 @@
+rocket-9999.ebuild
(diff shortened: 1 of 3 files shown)


marineam:
Before this lands in the image downloading things in the unpack stage will need to be removed but this is fine for a first pass for testing. Perhaps add a TODO?

[LGTM](http://www.lgtm.in/i/YSawp1W9q)

polvi:
I think this is the real solution: coreos/rocket#429

marineam:
https://github.com/coreos/rocket/issues/429 isn't everything, the `go get` of appc will also need to be removed.

jonboulle:
Yes, I already have a pr up against Rocket to vendor the spec.

PR #1066, opened 2015-01-27
syslinux: drop 6.03_pre9, switch to 4.07 for ISOs

marineam:
SYSLINUX is only used for creating ISO images these days so we no longer
need the 6.xx series with UEFI support. Once we get to adding UEFI
support to our images we should perhaps use GRUB to consolidate things.

This is prompted by a report that booting our ISOs hang on some Dell
servers but the older version works.


crawford:
LGTM

PR #1068, opened 2015-01-27
sys-libs/glibc: sync with upstream

crawford:
Diff from upstream:

```diff
--- glibc-2.19-r1.ebuild	2015-01-27 11:22:41.767766366 -0800
+++ glibc-2.19-r1.ebuild	2015-01-27 11:21:32.855714797 -0800
@@ -157,6 +157,9 @@
 }
 
 eblit-src_unpack-post() {
+	cd "${WORKDIR}"
+	epatch "${FILESDIR}"/locale-default-en_US.patch
+
 	if use hardened ; then
 		cd "${S}"
 		einfo "Patching to get working PIE binaries on PIE (hardened) platforms"
@@ -207,3 +210,10 @@
 		fi
 	fi
 }
+
+# CoreOS tweaks:
+# - drop host.conf and gai.conf
+# - nsswitch.conf and rpc are provided by baselayout
+eblit-src_install-post() {
+	rm -f "${D}"/etc/{gai.conf,host.conf,nsswitch.conf,rpc} || die
+}
```


philips:
lgtm

marineam:
Eegad, jumping all the way from 2.17 to 2.19? Can we just apply a patch to 2.17 for the quick fix which will be reasonable to push out to all channels? After that we can try updating libc in alpha only. Toolchain changes typically bring surprises.

crawford:
```
- A patch already exists (since May 21, 2013), and has been applied and
  tested since glibc-2.18, released on August 12, 2013:

        [BZ #15014]
        * nss/getXXbyYY_r.c (INTERNAL (REENTRANT_NAME))
        [HANDLE_DIGITS_DOTS]: Set any_service when digits-dots parsing was
        successful.
        * nss/digits_dots.c (__nss_hostname_digits_dots): Remove
        redundant variable declarations and reallocation of buffer when
        parsing as IPv6 address.  Always set NSS status when called from
        reentrant functions.  Use NETDB_INTERNAL instead of TRY_AGAIN when
        buffer too small.  Correct computation of needed size.
        * nss/Makefile (tests): Add test-digits-dots.
        * nss/test-digits-dots.c: New test.
```

I'm going to try backporting these commits.

marineam:
It may be worth peeking at what other distros have or will be shipping just in case there is something beyond that one commit we need to know about.

crawford:
OK, it looks like its just one commit (d5dd6189d506068ed11c8bfa1e1e9bffde04decd) from glibc.

crawford:
@marineam https://github.com/coreos/coreos-overlay/pull/1069

PR #1069, opened 2015-01-27
sys-libs/glibc: patch gethostbyname buffer overrun

crawford:
CVE-2015-0235


marineam:
[LGTM](http://www.lgtm.in/i/ynvDPcZOp)

Issue #1064, opened 2015-01-22
Support for SMB 2.1??


StephenOTT:
We are looking at this:
http://blog.elastocloud.org/2014/06/using-azure-file-service-on-linux.html

We are seeing that CoreOS currently does not appear to support SMB2.1

How can we get SMB2.1 support?

```
core@coreos-1 /tmp $ zcat /proc/config.gz | grep CONFIG_CIFS
CONFIG_CIFS=m
CONFIG_CIFS_STATS=y
CONFIG_CIFS_STATS2=y
CONFIG_CIFS_WEAK_PW_HASH=y
CONFIG_CIFS_UPCALL=y
CONFIG_CIFS_XATTR=y
CONFIG_CIFS_POSIX=y
CONFIG_CIFS_ACL=y
CONFIG_CIFS_DEBUG=y
# CONFIG_CIFS_DEBUG2 is not set
CONFIG_CIFS_DFS_UPCALL=y
# CONFIG_CIFS_SMB2 is not set
```

@sylus

StephenOTT:
@chanezzon

StephenOTT:
@chanezon

sylus:
:+1:

philips:
I see no reason not to enable this. /cc @gregkh @marineam

StephenOTT:
This is a major blocker for us at the moment because Azure File Service does not support smb1.x

gregkh:
Looks like a good idea to me.

StephenOTT:
Hey guys

Any potential timeline on this being added?

crawford:
......did not close this GitHub.

crawford:
@StephenOTT it will be present in the next release after 575.0.0 (should be released within a couple of days).

StephenOTT:
Thanks!!  Amazing.  Will really help us out!

crawford:
This is now in Alpha 584.0.0.

sylus:
Thanks so much you guys are awesome!

IbnSaeed:
Thanks

PR #1075, opened 2015-01-29
sys-kernel/coreos-kernel: enable support for SMB2

crawford:
Fixes https://github.com/coreos/coreos-overlay/issues/1064.


marineam:
For future record: As far as we can tell there is no reason to pull in `CONFIG_FSCACHE=m` when enabling SMB2/3, the generic fscache module is only needed if `CONFIG_CIFS_FSCACHE` is also set. Since the dependency appears to be bogus we should submit a patch to clean it up but for now we are stuck with it.

marineam:
LGTM

PR #1077, opened 2015-01-30
Transmission

polvi:
Adds transmission ebuild, and patches it to require a reduced set of libraries.


crawford:
Can you shorten that second commit message? Otherwise, LGTM.

brianredbeard:
I see that the `IUSE` is set for both QT4 and GTK. Just wanted to confirm that it's not building the corresponding GUI packages.

crawford:
`IUSE` is just the list of options. They are only set if they begin with `+`.

polvi:
rebased and fixed message, lets see if I can get this PR to update

polvi:
got it!

PR #1079, opened 2015-01-31
profiles: use glibc-2.17 for now

crawford:
Machines with glibc-2.17 had trouble upgrading to a build with glibc-2.19.
```
/tmp/au_postint_mount.M1lgpn/bin/cgpt: relocation error: /tmp/au_postint_mount.M1lgpn/lib/libc.so.6: symbol _dl_find_dso_for_object, version GLIBC_PRIVATE not defined in file ld-linux-x86-64.so.2 with link time reference
```

Let's roll this back to 2.17 for now.

### DIFF
--- a/profiles/coreos/base/package.accept_keywords
+++ b/profiles/coreos/base/package.accept_keywords
@@ -112,3 +112,5 @@ dev-util/checkbashisms
 
 # >=3.16 required by docker 1.4
 =sys-fs/btrfs-progs-3.17.1
+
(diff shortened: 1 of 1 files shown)


philips:
lgtm, I assume this was an attempt post-GHOST to try to get to a more up-to-date glibc?

crawford:
Yeah. It didn't go so well. I want to hold off for a little bit while we get some builds out. Then we can try again.

PR #1080, opened 2015-02-01
glibc: revert back to glibc-2.17

crawford:


crawford:
First attempt didn't work very well. https://github.com/coreos/coreos-overlay/pull/1079

philips:
lgtm

PR #1083, opened 2015-02-03
app-emulation/docker: bump to v1.5.0

crawford:
Depends on https://github.com/docker/docker/pull/10286.


carmstrong:
@crawford Can we :ship: now?

marineam:
The commit ID needs to be updated. Anything we need to watch out for in this release?

carmstrong:
> Anything we need to watch out for in this release?

We should confirm that Docker is choosing the correct storage driver now that we're using ext4 as a root filesystem. I think the default is devmapper until overlay is considered stable.

carmstrong:
@crawford Looks like the new commit is a8a31eff10544860d2188dddabdee4d727545796.

crawford:
Updated.

Next time, I am waiting until its _actually_ shipped. Regardless of their publicized release date.

marineam:
@carmstrong We don't depend on docker's logic for choosing a driver, a wrapper script handles it instead: https://github.com/coreos/coreos-overlay/blob/master/app-emulation/docker/files/dockerd

carmstrong:
Wondering if we should also pull in https://github.com/docker/docker/commit/eb76cb2301fc883941bc4ca2d9ebc3a486ab8e0a and https://github.com/docker/docker/commit/ae9cdbbde7824951b124a648e897ac4c291108e3.

marineam:
[LGTM](http://www.lgtm.in/i/EESJW67kn)

marineam:
Adding MountFlags=slave should be fine, not sure unrestricted core sizes are of much use since we don't provide an easy way to get the symbols outside of the SDK, also not sure what the default limit is. Should check on that.

marineam:
For reference we use the systemd coredump handler, which the kernel execs and feeds the dump over a pipe. In this mode RLIMIT_CORE is ignored by the kernel. The default limit in the systemd handler appears to be 2GB which should be plenty.... ;-)

crawford:
Added `MountFlags=slave`.

marineam:
DO IT!!!

carmstrong:
:thumbsup:

PR #1084, opened 2015-02-04
sys-apps/systemd: add patch for nspawn module load

crawford:
http://cgit.freedesktop.org/systemd/systemd/commit/?id=d0a0ccf3fecdb422d3fb7ab89646fe9042f11acd

Addresses https://github.com/coreos/bugs/issues/260.


vcaputo:
lgtm

PR #1087, opened 2015-02-07
rkt: bump to 0.3.1

polvi:


polvi:
hmm, maybe this is not working...

```
$ rkt run coreos.com/etcd
Error finding stage1 image "/usr/bin/stage1.aci": rkt only supports http or https URLs (/usr/bin/stage1.aci)
```

polvi:
had to add the stage1.aci to bin. fixed! ready to go.

crawford:
Can you unconditionally set `KEYWORDS="~amd64"` since rocket isn't stable yet? I missed this when the ebuild was introduced. Otherwise, LGTM.

jonboulle:
@crawford so this means it is available for people doing custom builds but not in any of the builds we produce yet, is that right? (do we have any public "experimental" builds or so?)

crawford:
That's correct. coreos-base/coreos does not depend on this, so it's not included. With my suggestion, it's a bit more difficult to accidentally include it (and it accurately represents the stability of the software).

We don't do any public experimental builds.

crawford:
0.3.2 was merged.

PR #1088, opened 2015-02-08
Load .bashrc.local, if present

Ulexus:
.bashrc is presently a symbolic link to the read-only /usr partition, which is a good idea.  However, there is no simple way to customize the login script with things like TERM overrides, aliases, etc.

This patch will source the `.bashrc.local` file, if present.

### DIFF
--- a/app-shells/bash/files/dot-bashrc
+++ b/app-shells/bash/files/dot-bashrc
@@ -16,3 +16,6 @@ fi
 
 
 # Put your fun stuff here.
+
+# Load local settings, if present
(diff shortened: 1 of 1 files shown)


marineam:
Just replace the symlink with a regular file. The default copy is super trivial so there isn't much need to track the original skel version.

The use of symlinks does tend to mislead people, both in the skel case and in /etc. At least for trivial stuff like skel we should probably switch to copying instead. Thoughts?

Ulexus:
Sounds good to me

PR #1091, opened 2015-02-10
sys-apps/iproute2: sync with upstream and enable IPVLAN

crawford:


marineam:
I think this was mentioned in person but not in here: please choose 3.17 or 3.18 and prune the other. If there is a release that has the IPVLAN patch lets just use that release.

crawford:
Updated to 3.19.

marineam:
[LGTM](http://www.lgtm.in/i/gksiWqlvn)

PR #1092, opened 2015-02-10
wa-linux-agent: don't start sshd.service

crawford:


marineam:
[LGTM](http://www.lgtm.in/i/Uqw1)

PR #1096, opened 2015-02-12
Update coreutils to 8.23

marineam:
The coreutils package is a bit old and some handy things have been added such as `du --inodes` in 8.22. Since this could potentially impact any number of things here is the release notes since our current version, 8.20:

```
GNU coreutils NEWS                                    -*- outline -*-

* Noteworthy changes in release 8.23 (2014-07-18) [stable]

** Bug fixes

  chmod -Rc no longer issues erroneous warnings for files with special bits set.
  [bug introduced in coreutils-6.0]

  cp -a, mv, and install --preserve-context, once again set the correct SELinux
  context for existing directories in the destination.  Previously they set
  the context of an existing directory to that of its last copied descendent.
  [bug introduced in coreutils-8.22]

  cp -a, mv, and install --preserve-context, no longer seg fault when running
  with SELinux enabled, when copying from file systems that return an error
  when reading the SELinux context for a file.
  [bug introduced in coreutils-8.22]

  cp -a and mv now preserve xattrs of symlinks copied across file systems.
  [bug introduced with extended attribute preservation feature in coreutils-7.1]

  date could crash or go into an infinite loop when parsing a malformed TZ="".
  [bug introduced with the --date='TZ="" ..' parsing feature in coreutils-5.3.0]

  dd's ASCII and EBCDIC conversions were incompatible with common practice and
  with POSIX, and have been corrected as follows.  First, conv=ascii now
  implies conv=unblock, and conv=ebcdic and conv=ibm now imply conv=block.
  Second, the translation tables for dd conv=ascii and conv=ebcdic have been
  corrected as shown in the following table, where A is the ASCII value, W is
  the old, wrong EBCDIC value, and E is the new, corrected EBCDIC value; all
  values are in octal.

      A   W   E
     041 117 132
     133 112 255
     135 132 275
     136 137 232
     174 152 117
     176 241 137
     313 232 152
     325 255 112
     345 275 241

  [These dd bugs were present in "the beginning".]

  df has more fixes related to the newer dynamic representation of file systems:
  Duplicates are elided for virtual file systems like tmpfs.
  Details for the correct device are output for points mounted multiple times.
  Placeholder values are output for inaccessible file systems, rather than
  than error messages or values for the wrong file system.
  [These bugs were present in "the beginning".]

  df now outputs all appropriate entries in the presence of bind mounts.
  On some systems, entries would have been incorrectly elided due to
  them being considered "dummy" mounts.
  [bug introduced in coreutils-8.22]

  du now silently ignores directory cycles introduced with bind mounts.
  Previously it would issue a warning and exit with a failure status.
  [bug introduced in coreutils-8.1]

  head --bytes=-N and --lines=-N now handles devices more
  consistently, not ignoring data from virtual devices like /dev/zero,
  or on BSD systems data from tty devices.
  [bug introduced in coreutils-5.0.1]

  head --bytes=-N - no longer fails with a bogus diagnostic when stdin's
  seek pointer is not at the beginning.
  [bug introduced with the --bytes=-N feature in coreutils-5.0.1]

  head --lines=-0, when the input does not contain a trailing '\n',
  now copies all input to stdout.  Previously nothing was output in this case.
  [bug introduced with the --lines=-N feature in coreutils-5.0.1]

  id, when invoked with no user name argument, now prints the correct group ID.
  Previously, in the default output format, it would print the default group ID
  in the password database, which may be neither real nor effective.  For e.g.,
  when run set-GID, or when the database changes outside the current session.
  [bug introduced in coreutils-8.1]

  ln -sf now replaces symbolic links whose targets can't exist.  Previously
  it would display an error, requiring --no-dereference to avoid the issue.
  [bug introduced in coreutils-5.3.0]

  ln -sr '' F no longer segfaults.  Now works as expected.
  [bug introduced with the --relative feature in coreutils-8.16]

  numfmt now handles blanks correctly in all unibyte locales.  Previously
  in locales where character 0xA0 is a blank, numfmt would mishandle it.
  [bug introduced when numfmt was added in coreutils-8.21]

  ptx --format long option parsing no longer falls through into the --help case.
  [bug introduced in TEXTUTILS-1_22i]

  ptx now consistently trims whitespace when processing multiple files.
  [This bug was present in "the beginning".]

  seq again generates correct output with start or end values = -0.
  [bug introduced in coreutils-8.20.]

  shuf --repeat no longer dumps core if the input is empty.
  [bug introduced with the --repeat feature in coreutils-8.22]

  sort when using multiple threads now avoids undefined behavior with mutex
  destruction, which could cause deadlocks on some implementations.
  [bug introduced in coreutils-8.6]

  tail -f now uses polling mode for VXFS to cater for its clustered mode.
  [bug introduced with inotify support added in coreutils-7.5]

** New features

  od accepts a new option: --endian=TYPE to handle inputs with different byte
  orders, or to provide consistent output on systems with disparate endianness.

  configure accepts the new option --enable-single-binary to build all the
  selected programs in a single binary called "coreutils".  The selected
  programs can still be called directly using symlinks to "coreutils" or
  shebangs with the option --coreutils-prog= passed to this program.  The
  install behavior is determined by the option --enable-single-binary=symlinks
  or --enable-single-binary=shebangs (the default).  With the symlinks option,
  you can't make a second symlink to any program because that will change the
  name of the called program, which is used by coreutils to determine the
  desired program.  The shebangs option doesn't suffer from this problem, but
  the /proc/$pid/cmdline file might not be updated on all the platforms.  The
  functionality of each program is not affected but this single binary will
  depend on all the required dynamic libraries even to run simple programs.
  If you desire to build some tools outside the single binary file, you can
  pass the option --enable-single-binary-exceptions=PROG_LIST with the comma
  separated list of programs you want to build separately.  This flag
  considerably reduces the overall size of the installed binaries which makes
  it suitable for embedded system.

** Changes in behavior

  chroot with an argument of "/" no longer implicitly changes the current
  directory to "/", allowing changing only user credentials for a command.

  chroot --userspec will now unset supplemental groups associated with root,
  and instead use the supplemental groups of the specified user.

  cut -d$'\n' again outputs lines identified in the --fields list, having
  not done so in v8.21 and v8.22.  Note using this non portable functionality
  will result in the delayed output of lines.

  ls with none of LS_COLORS or COLORTERM environment variables set,
  will now honor an empty or unknown TERM environment variable,
  and not output colors even with --colors=always.

** Improvements

  chroot has better --userspec and --group look-ups, with numeric IDs never
  causing name look-up errors.  Also look-ups are first done outside the chroot,
  in case the look-up within the chroot fails due to library conflicts etc.

  install now allows the combination of the -D and -t options.

  numfmt supports zero padding of numbers using the standard printf
  syntax of a leading zero, for example --format="%010f".
  Also throughput was improved by up to 800% by avoiding redundant processing.

  shred now supports multiple passes on GNU/Linux tape devices by rewinding
  the tape before each pass, avoids redundant writes to empty files,
  uses direct I/O for all passes where possible, and attempts to clear
  inode storage used for small files on some file systems.

  split avoids unnecessary input buffering, immediately writing input to output
  which is significant with --filter or when writing to fifos or stdout etc.

  stat and tail work better with HFS+, HFSX, LogFS and ConfigFS.  stat -f
  --format=%T now reports the file system type, and tail -f now uses inotify,
  rather than the default of issuing a warning and reverting to polling.


* Noteworthy changes in release 8.22 (2013-12-13) [stable]

** Bug fixes

  df now processes the mount list correctly in the presence of unstatable
  mount points.  Previously it may have failed to output some mount points.
  [bug introduced in coreutils-8.21]

  df now processes symbolic links and relative paths to special files containing
  a mounted file system correctly.  Previously df displayed the statistics about
  the file system the file is stored on rather than the one inside.
  [This bug was present in "the beginning".]

  df now processes disk device nodes correctly in the presence of bind mounts.
  Now df shows the base mounted file system rather than the last one mounted.
  [This bug was present in "the beginning".]

  install now removes the target file if the strip program failed for any
  reason.  Before, that file was left behind, sometimes even with wrong
  permissions.
  [This bug was present in "the beginning".]

  ln --relative now updates existing symlinks correctly.  Previously it based
  the relative link on the dereferenced path of an existing link.
  [This bug was introduced when --relative was added in coreutils-8.16.]

  ls --recursive will no longer exit with "serious" exit code (2), if there
  is an error reading a directory not specified on the command line.
  [Bug introduced in coreutils-5.3.0]

  mkdir, mkfifo, and mknod now work better when creating a file in a directory
  with a default ACL whose umask disagrees with the process's umask, on a
  system such as GNU/Linux where directory ACL umasks override process umasks.
  [bug introduced in coreutils-6.0]

  mv will now replace empty directories in the destination with directories
  from the source, when copying across file systems.
  [This bug was present in "the beginning".]

  od -wN with N larger than 64K on a system with 32-bit size_t would
  print approximately 2*N bytes of extraneous padding.
  [Bug introduced in coreutils-7.0]

  rm -I now prompts for confirmation before removing a write protected file.
  [Bug introduced in coreutils-6.8]

  shred once again uses direct I/O on systems requiring aligned buffers.
  Also direct I/O failures for odd sized writes at end of file are now handled.
  [The "last write" bug was introduced in coreutils-5.3.0 but masked
   by the alignment bug introduced in coreutils-6.0]

  tail --retry -f now waits for the files specified to appear.  Before, tail
  would immediately exit when such a file is initially inaccessible.
  [This bug was introduced when inotify support was added in coreutils-7.5]

  tail -F has improved handling of symlinks.  Previously tail didn't respond
  to the symlink target (re)appearing after being (re)created.
  [This bug was introduced when inotify support was added in coreutils-7.5]

** New features

  cp, install, mkdir, mknod, mkfifo and mv now support "restorecon"
  functionality through the -Z option, to set the SELinux context
  appropriate for the new item location in the file system.

  csplit accepts a new option: --suppressed-matched, to elide the lines
  used to identify the split points.

  df --output now accepts a 'file' field, to propagate a specified
  command line argument through to the output.

  du accepts a new option: --inodes to show the number of inodes instead
  of the blocks used.

  id accepts a new option: --zero (-z) to delimit the output entries by
  a NUL instead of a white space character.

  id and ls with -Z report the SMACK security context where available.
  mkdir, mkfifo and mknod with --context set the SMACK context where available.

  id can now lookup by user ID, in addition to the existing name lookup.

  join accepts a new option: --zero-terminated (-z). As with the sort,uniq
  option of the same name, this makes join consume and produce NUL-terminated
  lines rather than newline-terminated lines.

  uniq accepts a new option: --group to print all items, while separating
  unique groups with empty lines.

  shred accepts new parameters to the --remove option to give greater
  control over that operation, which can greatly reduce sync overhead.

  shuf accepts a new option: --repeat (-r), which can repeat items in
  the output.

** Changes in behavior

  cp --link now dereferences a symbolic link as source before creating the
  hard link in the destination unless the -P,--no-deref option is specified.
  Previously, it would create a hard link of the symbolic link, even when
  the dereferencing options -L or -H were specified.

  cp, install, mkdir, mknod and mkfifo no longer accept an argument to the
  short -Z option.  The --context equivalent still takes an optional argument.

  dd status=none now suppresses all non fatal diagnostic messages,
  not just the transfer counts.

  df no longer accepts the long-obsolescent --megabytes option.

  stdbuf now requires at least one buffering mode option to be specified,
  as per the documented interface.

** Improvements

  base64 encoding throughput for bulk data is increased by about 60%.

  md5sum can use libcrypto hash routines where allowed to potentially
  get better performance through using more system specific logic.
  sha1sum for example has improved throughput by 40% on an i3-2310M.
  This also affects sha1sum, sha224sum, sha256sum, sha384sum and sha512sum.

  stat and tail work better with EFIVARFS, EXOFS, F2FS, HOSTFS, SMACKFS, SNFS
  and UBIFS.  stat -f --format=%T now reports the file system type, and tail -f
  now uses inotify for files on all those except SNFS, rather than the default
  (for unknown file system types) of issuing a warning and reverting to polling.

  shuf outputs subsets of large inputs much more efficiently.
  Reservoir sampling is used to limit memory usage based on the number of
  outputs, rather than the number of inputs.

  shred increases the default write block size from 12KiB to 64KiB
  to align with other utilities and reduce the system call overhead.

  split --line-bytes=SIZE, now only allocates memory as needed rather
  than allocating SIZE bytes at program start.

  stty now supports configuring "stick" (mark/space) parity where available.

** Build-related

  factor now builds on aarch64 based systems [bug introduced in coreutils-8.20]


* Noteworthy changes in release 8.21 (2013-02-14) [stable]

** New programs

  numfmt: reformat numbers

** New features

  df now accepts the --output[=FIELD_LIST] option to define the list of columns
  to include in the output, or all available columns if the FIELD_LIST is
  omitted.  Note this enables df to output both block and inode fields together.

  du now accepts the --threshold=SIZE option to restrict the output to entries
  with such a minimum SIZE (or a maximum SIZE if it is negative).
  du recognizes -t SIZE as equivalent, for compatibility with FreeBSD.

  timeout now accepts the --preserve-status option to always propagate the exit
  status, useful for commands that can run for an indeterminite amount of time.

** Bug fixes

  cp --no-preserve=mode now no longer exits non-zero.
  [bug introduced in coreutils-8.20]

  cut with a range like "N-" no longer allocates N/8 bytes.  That buffer
  would never be used, and allocation failure could cause cut to fail.
  [bug introduced in coreutils-8.10]

  cut no longer accepts the invalid range 0-, which made it print empty lines.
  Instead, cut now fails and emits an appropriate diagnostic.
  [This bug was present in "the beginning".]

  cut now handles overlapping to-EOL ranges properly.  Before, it would
  interpret "-b2-,3-" like "-b3-".  Now it's treated like "-b2-".
  [This bug was present in "the beginning".]

  cut no longer prints extraneous delimiters when a to-EOL range subsumes
  another range.  Before, "echo 123|cut --output-delim=: -b2-,3" would print
  "2:3".  Now it prints "23".  [bug introduced in 5.3.0]

  cut -f no longer inspects input line N+1 before fully outputting line N,
  which avoids delayed output for intermittent input.
  [bug introduced in TEXTUTILS-1_8b]

  factor no longer loops infinitely on 32 bit powerpc or sparc systems.
  [bug introduced in coreutils-8.20]

  install -m M SOURCE DEST no longer has a race condition where DEST's
  permissions are temporarily derived from SOURCE instead of from M.

  pr -n no longer crashes when passed values >= 32.  Also, line numbers are
  consistently padded with spaces, rather than with zeros for certain widths.
  [bug introduced in TEXTUTILS-1_22i]

  seq -w ensures that for numbers input in scientific notation,
  the output numbers are properly aligned and of the correct width.
  [This bug was present in "the beginning".]

  seq -w ensures correct alignment when the step value includes a precision
  while the start value does not, and the number sequence narrows.
  [This bug was present in "the beginning".]

  seq -s no longer prints an erroneous newline after the first number, and
  outputs a newline after the last number rather than a trailing separator.
  Also seq no longer ignores a specified step value when the end value is 1.
  [bugs introduced in coreutils-8.20]

  timeout now ensures that blocking of ALRM signals is not inherited from
  its parent, which would cause timeouts to be ignored.
  [the bug dates back to the initial implementation]

** Changes in behavior

  df --total now prints '-' into the target column (mount point) of the
  summary line, accommodating the --output option where the target field
  can be in any column.  If there is no source column, then df prints
  'total' in the target column.

  df now properly outputs file system information with bind mounts present on
  the system by skipping duplicate entries (identified by the device number).
  Consequently, df also elides the early-boot pseudo file system type "rootfs".

  cut -d$'\n' no longer outputs lines identified in the --fields list,
  to align with other implementations and to avoid delayed output of lines.

  nl no longer supports the --page-increment option, which has been
  deprecated since coreutils-7.5.  Use --line-increment instead.

** Improvements

  readlink now supports multiple arguments, and a complementary
  -z, --zero option to delimit output items with the NUL character.

  stat and tail now know about CEPH.  stat -f --format=%T now reports the file
  system type, and tail -f uses polling for files on CEPH file systems.

  stty now supports configuring DTR/DSR hardware flow control where available.

** Build-related

  Perl is now more of a prerequisite.  It has long been required in order
  to run (not skip) a significant percentage of the tests.  Now, it is
  also required in order to generate proper man pages, via help2man.  The
  generated man/*.1 man pages are no longer distributed.  Building without
  perl, you would create stub man pages.  Thus, while perl is not an
  official prerequisite (build and "make check" will still succeed), any
  resulting man pages would be inferior.  In addition, this fixes a bug
  in distributed (not from clone) Makefile.in that could cause parallel
  build failure when building from modified sources, as is common practice
  for a patched distribution package.

  factor now builds on x86_64 with x32 ABI, 32 bit MIPS, and all HPPA systems,
  by avoiding incompatible asm.  [bug introduced in coreutils-8.20]

  A root-only test predicate would always fail.  Its job was to determine
  whether our dummy user, $NON_ROOT_USERNAME, was able to run binaries from
  the build directory.  As a result, all dependent tests were always skipped.
  Now, those tests may be run once again.  [bug introduced in coreutils-8.20]
```


crawford:
LGTM

PR #1105, opened 2015-02-16
flannel: load iptables kmod prior to start

eyakubovich:
flannel execs iptables which will try to load the
proper kernel mod if needed but that doesn't work
from container.

Fixes https://github.com/coreos/bugs/issues/270


crawford:
LGTM

philips:
lgtm

marineam:
ip_tables alone is sufficient? You don't need any match extensions, targets, or nat support?

eyakubovich:
@marineam ip_tables is sufficient (that's what iptables loads). The rest of the modules are pulled in as needed via request_module() mechanism. I was also stunned at first as to why only ip_tables.ko gets loaded but with Jed's help, we figured it out.

marineam:
Ah, I had trouble with that in the past but was using nspawn which places completely different restrictions on containers.

PR #1107, opened 2015-02-18
app-emulation/waagent: correct ebuild

crawford:
The SSH patch wasn't actually being applied because the ebuild silently
failed with "epatch: command not found".


marineam:
[LGTM](http://www.lgtm.in/i/IoFnmbgQs)

PR #1111, opened 2015-02-20
profiles: enable lzma on systemd

crawford:
This will allow journald to compress the journal. More interestingly, it will give journalctl the ability to read both COMPRESSED-XZ and non-compressed journals. This is useful if you want to read, for example, a Fedora 21 journal from CoreOS.


crawford:
Depends on https://github.com/coreos/portage-stable/pull/168.

marineam:
Ohhhhh, maybe that is why journalctl cannot deal with out Debian based access point management container too!

LGTM

crawford:
I can read the Debian container's journal with this patch.

PR #1113, opened 2015-02-21
sys-apps/systemd: bump to v219

crawford:
Depends on https://github.com/coreos/coreos-overlay/pull/1112.

### DIFF
--- a/profiles/coreos/base/package.use
+++ b/profiles/coreos/base/package.use
@@ -39,7 +39,7 @@ net-analyzer/nmap ncat -lua
 app-admin/sudo -sendmail
 
 # avoid pulling in gnutls, disable gentoo-only bits, enable journal upload
-sys-apps/systemd -ssl curl vanilla lzma gcrypt
+sys-apps/systemd -ssl curl vanilla -lz4 lzma gcrypt
 
 # disable kernel config detection and module building
 net-firewall/ipset -modules
--- a/sys-apps/systemd/systemd-9999.ebuild
+++ b/sys-apps/systemd/systemd-9999.ebuild
@@ -1,6 +1,6 @@
-# Copyright 1999-2014 Gentoo Foundation
+# Copyright 1999-2015 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
-# $Header: /var/cvsroot/gentoo-x86/sys-apps/systemd/systemd-9999.ebuild,v 1.153 2014/12/26 19:58:55 floppym Exp $
+# $Header: /var/cvsroot/gentoo-x86/sys-apps/systemd/systemd-9999.ebuild,v 1.160 2015/02/20 16:13:22 floppym Exp $
 
 EAPI=5
 
@@ -35,11 +35,13 @@ LICENSE="GPL-2 LGPL-2.1 MIT public-domain"
 SLOT="0/2"
 KEYWORDS="~alpha ~amd64 ~arm ~ia64 ~ppc ~ppc64 ~sparc ~x86"
 IUSE="acl apparmor audit cryptsetup curl doc elfutils gcrypt gudev http
-	idn introspection kdbus +kmod lz4 lzma pam policykit python qrcode +seccomp
-	selinux ssl sysv-utils terminal test vanilla xkb"
+	idn importd introspection kdbus +kmod +lz4 lzma nat pam policykit python
+	qrcode +seccomp selinux ssl sysv-utils terminal test vanilla xkb"
 
 # Gentoo removed the nls use flag, we'll keep it for now
 IUSE+=" nls symlink-usr"
+
+REQUIRED_USE="importd? ( curl gcrypt lzma )"
 
 MINKV="3.8"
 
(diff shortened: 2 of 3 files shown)


sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -173,6 +180,13 @@ fi
 	# Bug 463376
 	sed -i -e 's/GROUP="dialout"/GROUP="uucp"/' rules/*.rules || die
 
+	# missing in tarball
+	cp "${FILESDIR}"/217-systemd-consoled.service.in \
+		units/user/systemd-consoled.service.in || die
+
+	# https://github.com/coreos/bugs/issues/220
+	epatch "${FILESDIR}"/218-0001-networkd-accept-a-trailing-.-on-the-end-of-domains.patch
```
marineam: A different fix went in so this patch can be dropped.

https://github.com/systemd/systemd/commit/f50f01f4b738f2f00b30d0e02e8cf54ab99a9f27

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -173,6 +180,13 @@ fi
 	# Bug 463376
 	sed -i -e 's/GROUP="dialout"/GROUP="uucp"/' rules/*.rules || die
 
+	# missing in tarball
+	cp "${FILESDIR}"/217-systemd-consoled.service.in \
```
marineam: Also, seriously? Not a blocker but we should find out what is up with this.

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -100,20 +107,19 @@ DEPEND="${COMMON_DEPEND}
...
 	>=sys-devel/gcc-4.6
 	>=sys-kernel/linux-headers-${MINKV}
+	ia64? ( >=sys-kernel/linux-headers-3.9 )
 	virtual/pkgconfig
 	doc? ( >=dev-util/gtk-doc-1.18 )
 	python? ( dev-python/lxml[${PYTHON_USEDEP}] )
+	terminal? ( media-fonts/unifont[utils(+)] )
 	test? ( >=sys-apps/dbus-1.6.8-r1:0 )"
 
-# Only required if patches touch man page source xml, which is usually.
```
marineam: These dependencies kept biting us over and over as soon as we added just the right patches. Please leave them as it was.

profiles/coreos/base/package.use
```diff
@@ -39,7 +39,7 @@ net-analyzer/nmap ncat -lua
 app-admin/sudo -sendmail
 
 # avoid pulling in gnutls, disable gentoo-only bits, enable journal upload
-sys-apps/systemd -ssl curl vanilla lzma
+sys-apps/systemd -ssl curl vanilla -lz4 lzma gcrypt
```
marineam: why -lz4 here but +lz4 in the ebuild?

sys-apps/systemd/systemd-9999.ebuild
crawford: dropped

sys-apps/systemd/systemd-9999.ebuild
crawford: reverted

profiles/coreos/base/package.use
crawford: I was hoping to make few changes to the upstream ebuild. Gentoo adopted lz4, while everyone else went with lzma.

sys-apps/systemd/systemd-9999.ebuild
crawford: Woops. That should have left with the other patches. Removed.

profiles/coreos/base/package.use
marineam: On Feb 21, 2015 8:38 AM, "Alex Crawford" <notifications@github.com> wrote:
>
> In profiles/coreos/base/package.use:
>
> > @@ -39,7 +39,7 @@ net-analyzer/nmap ncat -lua
> >  app-admin/sudo -sendmail
> >
> >  # avoid pulling in gnutls, disable gentoo-only bits, enable journal
upload
> > -sys-apps/systemd -ssl curl vanilla lzma
> > +sys-apps/systemd -ssl curl vanilla -lz4 lzma gcrypt
>
> I was hoping to make few changes to the upstream ebuild. Gentoo adopted
lz4, while everyone else went with lzma.

Is it reasonable to support both or are you avoiding a new dependency?

crawford: We can do both, but we'll have to pull in app-arch/lz4. I don't feel strongly either way. What do you think?

crawford: I noticed that if I enable both, journald compresses with lz4 by default. Fedora, because it only has lzma, uses XZ by default. So, CoreOS will be able to read Fedora journals, but not the other way around.

marineam: Ok, let's stick with lzma.

marineam:
LGTM

PR #1118, opened 2015-02-26
baselayout: create default passwd/group files in the SDK

marineam:
This has always been done in the past by scripts/sdk_lib/make_chroot.sh
and is required to support adding new users and groups (the add commands
will not create these files themselves). Most users ebuilds add on the
fly are covered by our static group file here so this doesn't come up.
However dnsmasq was just added to the SDK and its ebuild adds a dnsmasq
user. We have no need for that user in production CoreOS images so no
point in adding it to the static list.

Without this `bootstrap_sdk` fails. Nothing else really changes.


crawford:
LGTM

Issue #1120, opened 2015-02-28
Feature Request: Add drivers/block/brd module to kernel


cnelson:
Is it possible to add the bdr module to the kernel?  Under certain workloads, having a block device in ram is extraordinarily useful.

The config option is:

CONFIG_BLK_DEV_RAM

marineam:
Done, should land in alpha next Thursday :)

PR #1121, opened 2015-02-28
coreos-kernel: upgrade to 3.19

marineam:


marineam:
Added ram disk module: https://github.com/coreos/coreos-overlay/issues/1120

crawford:
LGTM

PR #1123, opened 2015-03-03
net-libs/libtirpc: import from portage-stable

mischief:


crawford:
LGTM

marineam:
Could you prune this down to just the ebuild we are currently using any files it needs? No need to for the rest and it makes it harder to identify what our local changes are. So ChangeLog, metadata, can all go. Manifest will stay but should be pruned. `ebuild foo.ebuild digest` will regenerate it.

mischief:
PTAL

marineam:
LGTM

PR #1124, opened 2015-03-03
dev-db: use LimitNOFILE and LimitNPROC for etcd

carmstrong:
Docker and flannel already use these, and we've seen issues where
etcd runs out of file handles.

### DIFF
--- a/dev-db/etcd/files/etcd.service
+++ b/dev-db/etcd/files/etcd.service
@@ -6,6 +6,7 @@ User=etcd
 PermissionsStartOnly=true
 Environment=ETCD_DATA_DIR=/var/lib/etcd
 Environment=ETCD_NAME=%m
-ExecStart=/usr/bin/etcd 
+ExecStart=/usr/bin/etcd
 Restart=always
 RestartSec=10s
+LimitNOFILE=40000


carmstrong:
ping @marineam - we discussed this in IRC

dev-db/etcd/files/etcd.service
```diff
@@ -6,6 +6,8 @@ User=etcd
 PermissionsStartOnly=true
 Environment=ETCD_DATA_DIR=/var/lib/etcd
 Environment=ETCD_NAME=%m
-ExecStart=/usr/bin/etcd 
+ExecStart=/usr/bin/etcd
 Restart=always
 RestartSec=10s
+LimitNOFILE=1048576
```
marineam: This value could very easily be higher than the limit for the entire system. Does it need to be that high?

dev-db/etcd/files/etcd.service
```diff
@@ -6,6 +6,8 @@ User=etcd
...
+LimitNPROC=1048576
```
marineam: There should only be one processes, etcd, right?

dev-db/etcd/files/etcd.service
carmstrong: > Does it need to be that high?

It's the same value we're using elsewhere, so I'm just copying that - 

https://github.com/coreos/coreos-overlay/blob/342ea64a4d1bfdc7aeb842c3583b9399ec9c16ff/app-emulation/docker/files/docker.service#L11

https://github.com/coreos/coreos-overlay/blob/342ea64a4d1bfdc7aeb842c3583b9399ec9c16ff/app-emulation/docker/files/early-docker.service#L10

https://github.com/coreos/coreos-overlay/blob/df3eb13e36eda7cc2c680cc714ac5c14f08d8268/app-admin/flannel/files/flanneld.service#L16

marineam: Hm, which in turn we coppied from upstream docker which I'm sure picked the value out of thin air. Maybe something like `40000` would be a safer value so etcd leaking fds is less likely to harm the system as a whole. We should similarly reduce the value for flanneld.

gvilarino:
:+1:

carmstrong:
@marineam I addressed your comments. I'll lower the limit for flanneld too.

gvilarino:
@carmstrong should I drop the limits to `40000` as well?

marineam:
Great, thanks! LGTM

marineam:
Oh, the commit description is now out of date since PROC got dropped.

marineam:
Merged via https://github.com/coreos/coreos-overlay/issues/1131

PR #1125, opened 2015-03-04
net-libs/libtirpc: bump to 0.2.5

mischief:


marineam:
[LGTM](http://www.lgtm.in/i/G1ujSYf2q)

PR #1127, opened 2015-03-04
etcd: add 2.0.4 and make which versions to install selectable at build-time

marineam:
Default remains to only pull in etcd 0.4.7 but exporting `ETCD_PROTOCOLS=2` will switch to using 2.0 and `ETCD_PROTOCOLS=1 2` will install both in parallel.


crawford:
LGTM

PR #1128, opened 2015-03-04
app-admin/flannel: lower file descriptor limit

carmstrong:
See https://github.com/coreos/coreos-overlay/pull/1124#discussion_r25737006

ping @marineam


marineam:
LGTM, thanks!

PR #1129, opened 2015-03-04
sys-apps/systemd: mark systemd-219 stable on amd64

crawford:


marineam:
[LGTM](http://www.lgtm.in/i/w3c2)

PR #1130, opened 2015-03-04
net-libs/libtirpc: move netconfig file into /usr

mischief:
working toward coreos/bugs#147.


marineam:
Please add a -r1 to the ebuild version to indicate that what it installs has changed. Otherwise local SDKs will not update the package.

marineam:
[LGTM](http://www.lgtm.in/i/a7fiSLkqq)

PR #1132, opened 2015-03-05
net-fs/nfs-utils: import from portage-stable

mischief:


crawford:
Can you prune the unused patches from "files/"?

crawford:
Woops, nevermind.

crawford:
We don't need the initd files.

mischief:
@crawford: although i did remove the unused patches, removing the initd files will require modification of the ebuild. since this is the initial import from portage-stable, ok to do that in another commit?

crawford:
@mischief That's fair. Can you add another commit that prunes the unneeded files?

mischief:
woops.

mischief:
PTAL

mischief:
@marineam can nfs.confd and nfsmount.confd be nuked, if we only use the newest revs of the service files? according to coreos/baselayout#14 confd files are no longer needed.

marineam:
Yes, conf.d should not be used. If the version we were shipping still used them but a newer ebuild version has switched then we should use the newer revision. If there are any lingering references they should be pruned

mischief:
@marineam, ok, done. PTAL

net-fs/nfs-utils/files/nfs-utils.conf
```diff
@@ -0,0 +1,3 @@
+d /var/lib/nfs/rpc_pipefs
```
marineam: Is this sufficient for tmpfiles? I always write them with the `-` place holder for default/unused fields.

mischief:
that's how it was in the upstream package. https://github.com/coreos/portage-stable/blob/52dfc138f545f29971a4fcab2fe6794119b20925/net-fs/nfs-utils/files/nfs-utils.conf

i can make it explicit if that's better.

marineam:
Ok, that's fine then. LGTM

PR #1134, opened 2015-03-06
sys-apps/coreutils: dont copy DIR_COLORS into etc

mischief:
unfortunately, dircolors(1) is still installed in /usr/bin, but this is one less file in /etc.

### DIFF
--- a/sys-apps/coreutils/coreutils-8.20-r102.ebuild
+++ b/sys-apps/coreutils/coreutils-8.20-r102.ebuild
@@ -126,9 +126,6 @@ src_install() {
 	emake install DESTDIR="${D}" || die
 	dodoc AUTHORS ChangeLog* NEWS README* THANKS TODO
 
-	insinto /etc
-	newins src/dircolors.hin DIR_COLORS || die
-
 	if [[ ${USERLAND} == "GNU" ]] ; then
 		cd "${D}"/usr/bin
 		dodir /bin
--- a/sys-apps/coreutils/coreutils-8.23-r1.ebuild
+++ b/sys-apps/coreutils/coreutils-8.23-r1.ebuild
@@ -132,9 +132,6 @@ src_test() {
 src_install() {
 	default
 
-	insinto /etc
-	newins src/dircolors.hin DIR_COLORS
-
 	if [[ ${USERLAND} == "GNU" ]] && ! use symlink-usr; then
 		cd "${ED}"/usr/bin
 		dodir /bin


mischief:
whoops. should be in /usr/share according to coreos/bugs#295

marineam:
So `/etc/DIR_COLORS` is used in the default system bashrc to set `LS_COLORS` in the environment. The logic seems incomplete because even though it falls back to the builtin settings in `dircolors` to detect if the terminal supports color or not it does not use the builtin settings in `dircolors` to set LS_COLORS if /etc/DIR_COLORS is not present. I cannot tell if this was intentional or not.

mischief:
it seems like ls has its own builtin colors, if LS_COLORS isn't set, *and* dircolors has a builtin config for setting LS_COLORS.

marineam:
LGTM but before merging this lets simplify the bashrc to let dircolors do its thing.

```patch
diff --git a/app-shells/bash/files/bashrc b/app-shells/bash/files/bashrc
index a3773d6..41e916f 100644
--- a/app-shells/bash/files/bashrc
+++ b/app-shells/bash/files/bashrc
@@ -33,45 +33,25 @@ case ${TERM} in
 		;;
 esac
 
-use_color=false
-#BSD#@# BSD doesn't typically come with dircolors so we need
-#BSD#@# to hardcode some terminals in here.
-#BSD#@case ${TERM} in
-#BSD#@	xterm*|rxvt*|Eterm|aterm|kterm|gnome*|screen|cons25) use_color=true;;
-#BSD#@esac
-
 # Set colorful PS1 only on colorful terminals.
-# dircolors --print-database uses its own built-in database
-# instead of using /etc/DIR_COLORS.  Try to use the external file
-# first to take advantage of user additions.  Use internal bash
-# globbing instead of external grep binary.
-safe_term=${TERM//[^[:alnum:]]/?}   # sanitize TERM
-match_lhs=""
-[[ -f ~/.dir_colors   ]] && match_lhs="${match_lhs}$(<~/.dir_colors)"
-[[ -f /etc/DIR_COLORS ]] && match_lhs="${match_lhs}$(</etc/DIR_COLORS)"
-[[ -z ${match_lhs}    ]] \
-	&& type -P dircolors >/dev/null \
-	&& match_lhs=$(dircolors --print-database)
-[[ $'\n'${match_lhs} == *$'\n'"TERM "${safe_term}* ]] && use_color=true
-
-if ${use_color} ; then
-	# Enable colors for ls, etc.  Prefer ~/.dir_colors #64489
-	if type -P dircolors >/dev/null ; then
-		if [[ -f ~/.dir_colors ]] ; then
-			eval $(dircolors -b ~/.dir_colors)
-		elif [[ -f /etc/DIR_COLORS ]] ; then
-			eval $(dircolors -b /etc/DIR_COLORS)
-		fi
+# Enable colors for ls, etc.  Prefer ~/.dir_colors #64489
+if type -P dircolors >/dev/null ; then
+	if [[ -f ~/.dir_colors ]] ; then
+		eval $(dircolors -b ~/.dir_colors)
+	elif [[ -f /etc/DIR_COLORS ]] ; then
+		eval $(dircolors -b /etc/DIR_COLORS)
+        else
+		eval $(dircolors -b)
 	fi
-
+fi
+if [[ -n "${LS_COLORS}" ]] ; then
 	if [[ ${EUID} == 0 ]] ; then
 		PS1='\[\033[01;31m\]\h\[\033[01;34m\] \W \$\[\033[00m\] '
 	else
 		PS1='\[\033[01;32m\]\u@\h\[\033[01;34m\] \w \$\[\033[00m\] '
 	fi
 
-	#BSD#@export CLICOLOR=1
-	#GNU#@alias ls='ls --color=auto'
+	alias ls='ls --color=auto'
 	alias grep='grep --colour=auto'
 else
 	if [[ ${EUID} == 0 ]] ; then
@@ -81,6 +61,3 @@ else
 		PS1='\u@\h \w \$ '
 	fi
 fi
-
-# Try to keep environment pollution down, EPA loves us.
-unset use_color safe_term match_lhs
```

marineam:
@mischief missed your comment on `ls` not needing LS_COLORS. If that's the case the bashrc is correct but that doesn't seem to work for me.

mischief:
on the image i generated with this change, there's no /etc/DIR_COLORS, nor is LS_COLORS set by the system bashrc, but i still get colors when running 'ls --color=auto'

marineam:
Alrighty then, not sure what I was hitting. Maybe just a different behavior in a different version.

LGTM

PR #1135, opened 2015-03-06
rocket: general cleanup and install stage1.aci

marineam:
This freshens up the current rocket ebuild, using emerge to download the
PXE image for proper caching between builds and fixing dependencies.

For reference this is how big the build is as-is:

    6.8M /build/amd64-usr/usr/bin/rkt
    12M  /build/amd64-usr/usr/share/rkt/stage1.aci

We may be able to improve that size but for now this will do.


crawford:
LGTM

PR #1137, opened 2015-03-11
coreos-base/coreos: add rocket ebuild to coreos

pbx0:
@marineam image builds and rocket seems to run fine. Also, is there an actool ebuild?


marineam:
[LGTM](http://www.lgtm.in/i/86LEpXXUp)

PR #1138, opened 2015-03-11
Import keyutils

mischief:
import keyutils from portage-stable. move conf files into /usr and setup symlinks with tmpfiles.d.


sys-apps/keyutils/files/tmpfiles.d/keyutils.conf
```diff
@@ -0,0 +1,3 @@
+L	/etc/request-key.conf	-	-	-	-	../usr/share/keyutils/request-key.conf
+L	/etc/request-key.d	-	-	-	-	../usr/share/keyutils/request-key.d
```
marineam: Please create this as a directory and add symlinks inside it as needed. This may require adjusting other ebuilds to use the /usr/share location instead of relying on the symlink.

sys-apps/keyutils/keyutils-1.5.9-r2.ebuild
```diff
@@ -0,0 +1,89 @@
...
+multilib_src_test() {
+	# Execute the locally compiled code rather than the
+	# older versions already installed in the system.
+	LD_LIBRARY_PATH=${BUILD_DIR} \
+	PATH="${BUILD_DIR}:${PATH}" \
+	emake test
+}
+
+multilib_src_install() {
+	# rewrite etc path so conf files are installed to /usr.
```
marineam: couldn't you call `emake install ETCDIR=/usr/share/keyutils DISTDIR="${D}"` instead? (replacing the call to `default` below)

mischief: sounds good, i did not realize you could override the makefile variable when its set with ":="

sys-apps/keyutils/files/tmpfiles.d/keyutils.conf
mischief: if `/etc/request-key.d` is a directory, i dont see any reason for `/usr/share/keyutils/request-key.d` to be created by the make install. should that be patched out of the makefile?

marineam:
The /usr/share dir doesn't hurt and is going to wind up with a file for NFS anyway so just leave it.

mischief:
PTAL

marineam:
LGTM, please squash down the commits a bit before merging though.

marineam:
[LGTM](http://www.lgtm.in/i/2FGiyeuSp)

PR #1142, opened 2015-03-13
systemd: add UseNTP DHCP option for network configs

marineam:
Updates both 218 and 219, for now only 218 has been tested.


crawford:
Can you add the patch to the 219 ebuild as well?

marineam:
@crawford fixed, split the 219 ebuild copy from the patch commit too

marineam:
@crawford updated again, adding a second patch to allow timesyncd to start in virtual machines. Both patches are upstream.

crawford:
LGTM

PR #1143, opened 2015-03-13
README: document that ebuilds are licensed under the GPLv2

marineam:
This differs from a lot of our code which is Apache or BSD licensed.


crawford:
LGTM

PR #1150, opened 2015-03-19
sys-apps/systemd: fix networkd-wait-online interface checks to prevents ...

mischief:
...hangs on systems with multiple NICs


mischief:
this patch addresses coreos/bugs#279. i tested manually in qemu using the monitor to run `set_link <if> off`to simulate no-carrier. it would be nice if mantle supported sending monitor commands to each machine in the cluster so that this could be programatically tested in kola.

marineam:
Could you provide the patch as a complete git commit generated by `git format-patch` please? That way it is trivial to import into a git branch. Ideally the patch here is the exact same email that gets sent to systemd-devel@

mischief:
@marineam ptal

crawford:
@mischief please rebase

marineam:
LGTM

PR #1157, opened 2015-03-23
etcd: use etcd-starter and bump to latest releases

crawford:
etcd-starter needs some work, but the rest of this should be good.


crawford:
Depends on https://github.com/coreos/manifest/pull/60.

dev-db/etcd-starter/etcd-starter-9999.ebuild
```diff
@@ -0,0 +1,37 @@
...
+# Distributed under the terms of the GNU General Public License v2
+# $Header:$
+#
+
+EAPI=5
+CROS_WORKON_PROJECT="coreos/etcd-starter"
+CROS_WORKON_LOCALNAME="etcd-starter"
+CROS_WORKON_REPO="git://github.com"
+COREOS_GO_PACKAGE="github.com/coreos/etcd-starter"
+inherit coreos-doc coreos-go toolchain-funcs cros-workon
```
marineam: Unless I'm missing something coreos-doc and toolchain-funcs are unused.

crawford: Don't need coreos-doc

crawford: Good timing.

crawford: Fixed

crawford:
Using etcd-starter 0.0.2.

kelseyhightower:
@crawford lgtm

dev-db/etcd/etcd-0.4.8.ebuild
```diff
@@ -37,14 +37,14 @@ src_compile() {
 }
```
marineam: `!etcd_protocols_2? ( !dev-db/etcd:2 )` needs to be updated to conflict with etcd-starter instead.

crawford: good catch

dev-db/etcd-starter/etcd-starter-9999.ebuild
```diff
@@ -0,0 +1,37 @@
...
+
+DESCRIPTION="etcd-starter"
+HOMEPAGE="https://github.com/coreos/etcd-starter"
+SRC_URI=""
+
+LICENSE="Apache-2.0"
+SLOT="0"
+
+DEPEND=">=dev-lang/go-1.2"
+
```
marineam: needs something like this:
```sh
RDEPEND="!dev-db/etcd:2[etcd_protocols_1]
    !dev-db/etcd:1[etcd_protocols_2]"
```

Basically this ebuild installs /usr/bin/etcd so it needs to conflict with the situations in which the other two ebuilds may install /usr/bin/etcd. Alternatively (if etcd-starter is ready) always install etcd-starter as /usr/bin/etcd so we can drop all this weird use flag logic to determine who installs /usr/bin/etcd

dev-db/etcd/etcd-0.4.8.ebuild
marineam: On second thought putting the conflicts in etcd-starter would be simpler.

crawford:
@marineam updated

dev-db/etcd/etcd-9999.ebuild
```diff
@@ -24,7 +24,6 @@ SRC_URI=""
 
 LICENSE="Apache-2.0"
 SLOT="2"
-IUSE=""
```
marineam: This line can be left alone.

crawford: good to go

marineam:
LGTM

PR #1158, opened 2015-03-23
sys-apps/systemd: nspawn Map seccomp filters to caps

jayofdoom:
http://cgit.freedesktop.org/systemd/systemd/commit/?id=9a71b1122c6e49dd9227f82b2f53837c7ea13019

Fixes coreos/bugs#293


crawford:
You'll also need to bump the file name to sys-apps/systemd/systemd-219-r2.ebuild

crawford:
Can you also add this to the 218 ebuild? We aren't currently shipping 219.

crawford:
It builds... LGTM!

marineam:
[LGTM](http://www.lgtm.in/i/ier1)

PR #1162, opened 2015-03-25
systemd: build from git instead of release tarballs

marineam:
To compare the patch list see: https://github.com/coreos/systemd/tree/v218-coreos and https://github.com/coreos/systemd/tree/v219-coreos


sys-apps/systemd/systemd-218-r10.ebuild
```diff
@@ -4,22 +4,22 @@
...
-inherit git-r3
+CROS_WORKON_PROJECT="coreos/systemd"
+CROS_WORKON_REPO="git://github.com"
+
+if [[ "${PV}" == 9999 ]]; then
+	# Use ~arch instead of empty keywords for compatibility with cros-workon
+	KEYWORDS="~amd64 ~arm ~x86"
+else
+	CROS_WORKON_COMMIT="85fa71efc8554c20f7886fbf9ec40e47dc4fcb57"
+	KEYWORDS="~amd64 ~arm ~x86"
```
crawford: amd64?

marineam: oops, fixed. Left the other KEYWORDS line in by mistake

crawford:
LGTM

PR #1170, opened 2015-03-30
sys-kernel/coreos-kernel: enable ZRAM

### LINKED ISSUE #1169: Enable ZRAM as kernel module
[ZRAM](http://en.wikipedia.org/wiki/Zram) provides compressed memory as a block device. Its primary use is to increase the available RAM  on memory-constrained systems, but it can also be used as a compressed version of tmpfs.

Disk usage from the kernel module is minimal and user-space configuration makes it highly suitable for use with systemd. It would provide a useful alternative to disk-based swap when running CoreOS in memory-constrained environments. (Particular use-case in mind is low memory cloud VMs for etcd clusters.)

mischief:
closes #1169


marineam:
LGTM

PR #1171, opened 2015-03-30
sys-apps/systemd: remove SRC_URI from ebuild

mischief:
since systemd is checked out from git, we dont download a tarball any
more.


marineam:
Does the 218 ebuild have the same issue?

marineam:
Perfect, you can also drop `Manifest`.

marineam:
LGTM

PR #1173, opened 2015-03-30
dev-db/etcd: remove starter

crawford:
This will end up installing etcd and etcd2 side-by-side, with their corresponding service units and documentation.


dev-db/etcd/files/etcd2.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=etcd
+
+[Service]
+User=etcd
+PermissionsStartOnly=true
+Environment=ETCD_DATA_DIR=/var/lib/etcd
+ExecStart=/usr/bin/etcd
```
mischief: should this be /usr/bin/etcd2?

crawford: That's a good question. @philips, what is the recommended strategy here? Should we use a separate data-dir for etcd2 and ask the user to move their data (or update the service file) if they want to migrate to etcd2? Or just use the same directory?

mischief: sorry, i did not mean the data dir, but thats a good point too. i was referencing `newbin "${WORKDIR}/gopath/bin/${PN}" etcd2` in the ebuild. this service file is called 'etcd2.service' so i thought it would call /usr/bin/etcd2 installed by the etcd2 ebuild.

dev-db/etcd/files/etcd2.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=etcd2
+
+[Service]
+User=etcd
+PermissionsStartOnly=true
+Environment=ETCD_DATA_DIR=/var/lib/etcd
```
marineam: `/var/lib/etcd2`

Data directories need to be different. That also means etcd2 needs its own tmpfiles to create the new data dir.

crawford: Ah, I didn't realized that was a tmpfiles config. Updated.

dev-db/etcd/files/etcd2.conf
```diff
@@ -0,0 +1,2 @@
+d    /var/lib/etcd2 0755 etcd etcd - -
+d    /var/run/etcd2 0755 etcd etcd - -
```
marineam: Is run actually used?

crawford: @philips says it is not. Removed.

dev-db/etcd/files/etcd2.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=etcd2
+Conflicts=etcd.service
+
+[Service]
+User=etcd
+PermissionsStartOnly=true
```
marineam: There isn't an ExecStartPre these days so this doesn't do anything as far as I know.

marineam:
LGTM, just clean up `PermissionsStartOnly=true`

philips:
lgtm, after fixing mike's thing.

crawford:
Updated.

PR #1177, opened 2015-04-01
app-emulation/rkt: update rkt

crawford:


app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -32,7 +32,7 @@ DEPEND=">=dev-lang/go-1.2
 RDEPEND=""
```
jonboulle: can we move to go-1.3?

crawford: The SDK currently uses go-1.3.3. This is just forcing the version to be _at least_ go-1.2.

marineam: We are actually using go 1.3 in the SDK, but if that is a strict requirement it is worth documenting in the ebuild.

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -2,24 +2,24 @@
...
 fi
 
-# Must be in sync with stage1/rootfs/usr/cache.sh
+# Must be in sync with stage1/rootfs/usr_from_coreos/cache.sh
 IMG_RELEASE="444.5.0"
 IMG_URL="http://stable.release.core-os.net/amd64-usr/${IMG_RELEASE}/coreos_production_pxe_image.cpio.gz"
 
-DESCRIPTION="rocket"
-HOMEPAGE="https://github.com/coreos/rocket"
+DESCRIPTION="rkt"
```
marineam: Maybe worth putting a real description there? :-)

app-emulation/rkt/rkt-9999.ebuild
crawford: OK, @jonboulle says 1.3 is a requirement. I'll update the ebuild.

marineam: Unrelated to the 1.3 question, but either add a conflicts with the old rocket name so it gets uninstalled or add an entry to coreos-overlay/profiles/updates to inform portage that the name of the installed package has changed.

app-emulation/rkt/rkt-9999.ebuild
crawford: Done.

marineam:
LGTM

app-emulation/rkt/rkt-9999.ebuild
crawford: TIL. Updated.

PR #1180, opened 2015-04-03
oem-gce: bind-mount /var/run/docker.sock to gcloud container

GeertJohan:
The command `gcloud preview docker` requires access to the host system docker.

I will create a PR at https://github.com/GoogleCloudPlatform/cloud-sdk-docker to add docker to the container.

See https://github.com/GeertJohan/dockerfiles/tree/master/google-cloud-sdk-with-docker for more information.


mischief:
LGTM

PR #1181, opened 2015-04-04
rkt: add rocket conflict dependency

marineam:
Portage is not processing the rocket->rkt package move for me in
/build/amd64-usr for some reason. Not sure what trick we are missing. As
an alternative to the package move use a conflict to remove the old
package instead.


crawford:
:/ LGTM

PR #1182, opened 2015-04-05
app-admin/fleet: remove dependency on etcd

crawford:
We have to think about this one. We've done a pretty good job telling people to explicitly start etcd and fleet from their cloud-configs, but this will still break backwards-compatibility.


philips:
It seems OK to me.

marineam:
Re: c9b38cd21e8d441ce24ed0f6df1bdbf79f978e49 you should keep the after rule and include etcd2, just drop the wants.

We will need to call this one out loudly in the change log as anyone who is used to starting fleet but not etcd will be surprised by this. Fortunately we've always listed both in our examples as far as I know.

crawford:
Updated.

marineam:
@crawford need to add etcd2.service

crawford:
;_; Too much context switching.

marineam:
[LGTM](http://www.lgtm.in/i/M8YNeuxoq)

crawford:
Holding off until the Alpha is built.

PR #1183, opened 2015-04-06
update_engine: remove dependency on libchromeos

marineam:
Magic of dead code removal. See https://github.com/coreos/update_engine/pull/61


marineam:
As a bonus, we are down by one dbus binding library!

crawford:
LGTM

PR #1184, opened 2015-04-07
Add basic support for arm64 

glevand:
Adds basic support for arm64 platforms.  Please consider.

-Geoff


profiles/profiles.desc
```diff
@@ -4,3 +4,10 @@ amd64   coreos/amd64/generic/prod           dev
 amd64   coreos/amd64/sdk                    dev
 amd64   default/linux/amd64/10.0            dev
 amd64   default/linux/amd64/10.0/coreos     dev
+
+arm64   coreos/arm64/generic                dev
+arm64   coreos/arm64/generic/dev            dev
+arm64   coreos/arm64/generic/prod           dev
+arm64   coreos/arm64/sdk                    dev
+arm64   default/linux/arm64/10.0            dev
```
marineam: The 10.0 profiles are old, they only exist for amd64 because I hadn't deleted it yet.

glevand:
I pushed out a new set that omits the first patch, which added the 10.0 profile, and removes the reference to '10.0' in profiles.desc.  Also rebased to the latest master.

marineam:
LGTM, thanks!

PR #1187, opened 2015-04-08
eclass: strip down cros-workon, switch to git-r3

marineam:
The cros-workon eclass was a mix of all sorts of magic tricks from
ChromeOS but the only ones we care about are the abilities to clone from
local git repos and manage live ebuilds. Many of the extra config
options are unused now in CoreOS and the whole "common.mk" build system
is completely irrelevant. Gutting it all will hopefully make problems we
encounter with it easier to understand.

Also, by moving from the git-2 to the git-r3 eclass for fetching remote
repositories we can use the new EGIT_CLONE_TYPE=mirror option to fetch
all refs of remote repositories, hopefully avoiding the difficulty of
managing packages with multiple development branches like systemd.

### DIFF
--- a/eclass/cros-workon.eclass
+++ b/eclass/cros-workon.eclass
@@ -2,30 +2,11 @@
 # Distributed under the terms of the GNU General Public License v2
 
 # @ECLASS: cros-workon.eclass
-# @MAINTAINER:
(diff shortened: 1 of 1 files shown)


crawford:
LGTM

PR #1190, opened 2015-04-09
profiles: mask cross compiler flags that are masked on target

marineam:
The hardened and sanitizer options are masked on arm64 so the arm64
cross compiler should also not enable them instead of inheriting the
amd64 default to enable them. Fixes building cross gcc.


mischief:
lgtm

crawford:
LGTM

PR #1200, opened 2015-04-14
dev-db/etcd: set ETCD_NAME to machine ID

crawford:
Each node's name is required to be unique during bootstrapping.


xiang90:
LGTM

marineam:
LGTM, not sure how we forgot that.

crawford:
@marineam I was under the impression that it wasn't needed. This is why we test!

PR #1201, opened 2015-04-14
flannel: remove wants etcd and add requires etcd2

eyakubovich:


crawford:
We'll need to make some noise about this one. If users haven't explicitly started etcd, this might cause the node to fail when it's updated.

crawford:
LGTM

crawford:
LGTM

PR #1204, opened 2015-04-16
update_engine: convert to autotools and make

marineam:
scons has been replaced by a standard autotools build system. Beyond
that a few things are different:

 - remove GCC 4.8 workaround, the build system handles this now.
 - instead of trying to make executing code in a cross-compiled $SYSROOT
   work just avoid running the test suite for board builds. The tests
   can still run for the native SDK build.
 - `make install` now handles at least part of the install process.


crawford:
LGTM

PR #1205, opened 2015-04-17
update_engine: Add arm64 keyword to ebuild

glevand:


glevand:
I didn't test this yet.  Do you want to merge?

marineam:
This alone isn't going to work until its dependencies also have the right keyword but I'm fine with landing these out of order since there isn't any negative impact. So LGTM

PR #1207, opened 2015-04-17
profiles: add new eselect package names to package.provided

marineam:
Upstream has renamed the eselect modules, add the new names in
preparation for updating portage-stable.


crawford:
LGTM

PR #1213, opened 2015-04-21
mantle: bump to use commit with fixed build

pbx0:
Fixes build issue from last mantle bump from missing path rewrite.


crawford:
LGTM

PR #1216, opened 2015-04-22
config_wrapper: Fix tool path

glevand:
Create a new variable TOOLPATH to use when executing host tools.

When executing host tools the path must only include directories of executables
for the host.  In its operation config_wrapper modifies the path to include the
board build directory which has board executables.

Fixes build errors like these when the host and board architectures differ:

  /build/arm64-usr/build/bin/aarch64-cros-linux-gnu-gpg-error-config: /build/arm64-usr/usr/bin/sed: No such file or directory

---
We could also have something like:

SED=$(which sed)
PATH="${SYSROOT}/usr/bin:${PATH}"
...
exec ${cfg} "$@" | ${SED} -r "s:(-[IL])(${SYSROOT})?:\1${SYSROOT}:g"

-Geoff


marineam:
Gah, I swar I fixed this but must have never merged it. :( I'll dig up that fix to see if there was anything else it handled.

glevand:
Maybe your patch moved config_wrapper from coreos-overlay to scripts?

marineam:
Found it in an experimental branch from January, it drops the $PATH magic altogether. Let me know if that works for you: https://github.com/coreos/coreos-overlay/pull/1218

glevand:
Sure, either patch seems fine.

PR #1217, opened 2015-04-22
Add secure boot code

mjg59:
Add shim and update grub to incorporate required functionality


sys-boot/shim/shim-9999.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	default_src_unpack
+}
+
+src_compile() {
+	make EFI_PATH=/usr/lib shim.efi
+}
+
+src_install() {
+	insinto /usr/lib/shim
+	newins "shim.efi" shim.efi
```
marineam: When the name doesn't change you can use doins

sys-boot/shim/shim-9999.ebuild
```diff
@@ -0,0 +1,34 @@
...
+
+DESCRIPTION="UEFI Shim loader"
+HOMEPAGE="https://github.com/rhinstaller/shim"
+
+LICENSE="BSD"
+SLOT="0"
+IUSE=""
+
+RDEPEND=""
+DEPEND="sys-boot/gnu-efi"
```
marineam: Doesn't shim depend on some other stuff to build like sb-sign or... I forgot

mjg59: Ah, yeah, it depends on openssl. I'll add that.

sys-boot/shim/shim-9999.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2014 Gentoo Foundation
```
marineam: Yeah, vim does this by default but we are CoreOS, Inc. :)

marineam:
one last header nit and then LGTM

PR #1220, opened 2015-04-23
app-emulation/docker: bump to v1.6.0

mischief:
this commit also removes a patch by brandon philips which was applied to
upstream docker.

it also brings the ebuild closer in sync to the build in tianon/docker-overlay.


marineam:
LGTM

PR #1222, opened 2015-04-24
update_engine: fix build with USE=delta_generator

marineam:
Pull in https://github.com/coreos/update_engine/pull/69


mjg59:
LGTM

PR #1223, opened 2015-04-24
oem-*: fix etcd2 peer election option

crawford:


marineam:
Aww damn, does etcd2 use the old value but it is just deprecated or is this entirely broken?

LGTM

crawford:
I think this was just entirely broken. I haven't heard much complaint about it though. I only noticed the issue because I saw it in a user's log.

PR #1225, opened 2015-04-24
flannel: support for aws-vpc backend and bump to 0.4.0

eyakubovich:
aws-vpc backend expects AWS credentials to be passed via
env vars. Propogate them into the container via --env.


crawford:
LGTM

PR #1228, opened 2015-04-28
app-emulation/docker: turn off contrib, turn on btrfs & overlay again

mischief:
i goofed when updating the docker 1.6.0 ebuild.


marineam:
LGTM

PR #1229, opened 2015-04-28
Portage 2.2.18

marineam:
Been a long while since we've synced with upstream.

The most interesting thing in this update you are likely to notice is that portage will ignore binary packages built with different dependencies, causing things to get rebuilt more often but likely avoiding some problems.


mischief:
lgtm

crawford:
LGTM

Issue #1234, opened 2015-04-29
Missing mvsas module in kernel


dmitryint:
Please, enable mvsas module for Supermicro AOC-SASLP-MV8.

PR #1235, opened 2015-04-29
coreos-kernel: bump to 4.0.1, enable mvsas driver

marineam:
Fixes https://github.com/coreos/coreos-overlay/issues/1234


mischief:
lgtm

PR #1239, opened 2015-05-04
Enable dell RBU and DCDBAS kernel modules.

zapient:
Closes: https://github.com/coreos/bugs/issues/346

PR #1240, opened 2015-05-04
Enable dell RBU and DCDBAS kernel modules

marineam:
Minor revision to https://github.com/coreos/coreos-overlay/pull/1239


mischief:
lgtm

PR #1241, opened 2015-05-07
spidermonkey: fix improper error handling

marineam:
Mixing `local` and `|| die` will never work because the exit status
being tested is that of local, not any subcommands that may be involved.


glevand:
LGTM

crawford:
LGTM

PR #1248, opened 2015-05-11
sys-app/ignition: add initial ebuild and service

crawford:


sys-apps/ignition/files/ignition@.service
```diff
@@ -0,0 +1,17 @@
...
+After=initrd-fs.target
+ConditionPathExists=/etc/initrd-release
+ConditionPathExists=!/sysroot/etc/machine-id
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+Environment=OEM_NAME=pxe
+EnvironmentFile=-/usr/share/oem/environment
+# TODO(crawford): /sysroot should be a variable
```
marineam: eh, /sysroot is hard coded into systemd ;-)

sys-apps/ignition/files/ignition@.service
```diff
@@ -0,0 +1,17 @@
...
+ExecStart=/bin/ignition --root=/sysroot --oem=${OEM_NAME} --stage=%i
```
marineam: Dracut doesn't put it in /usr/bin like it is on the filesystem?

crawford: Fixed :)

marineam:
Drop the pointless TODO, /sysroot is always the same in the initrd. Otherwise LGTM

[LGTM](http://lgtm.in/i/KFdvznpiq)

PR #1249, opened 2015-05-11
oem-ec2-compat: blacklist xen_fbfront

jedsmith:
To complete the fix started in 480fb78, we ban xen_fbfront from loading on EC2.
Testing missed a case where this would still happen.

(Testing this fix now.)

### DIFF
--- a/coreos-base/oem-ec2-compat/files/grub-ec2.cfg
+++ b/coreos-base/oem-ec2-compat/files/grub-ec2.cfg
@@ -1,0 +1,4 @@
+# CoreOS GRUB settings for EC2
+
+# Blacklist the Xen framebuffer module so it doesn't get loaded at boot
+set linux_append="modprobe.blacklist=xen_fbfront"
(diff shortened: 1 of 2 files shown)


crawford:
Is there any reason not to yank out that module entirely?

jedsmith:
I can't imagine a scenario where it's needed. Nonetheless, the fix works (just tested).

coreos-base/oem-ec2-compat/files/grub.cfg
```diff
@@ -0,0 +1,4 @@
+# CoreOS GRUB settings for EC2
```
crawford: Can you give this file an ec2-specific name? (in case we get other grub configs in the future)

jedsmith:
Does `doins` take an overriding filename as the next parameter? I'm not very familiar with ebuild.

crawford:
You can use `newbin`, the second parameter being the target file name.

jedsmith:
Like so?

crawford:
Almost. Use `newins` instead of `newbin` (I realize I gave you the wrong command before).
https://devmanual.gentoo.org/function-reference/install-functions/

jedsmith:
Fixed. :rage:

crawford:
LGTM. Go ahead and squash those commits.

jedsmith:
Done.

PR #1250, opened 2015-05-12
sys-kernel/coreos-kernel: mount root read/write

crawford:


marineam:
LGTM, we should test mixing this kernel config and booting with `usr=` instead of `mount.usr=` just to be sure adding mount.usrflags doesn't cause the systemd generator to kick in and conflict with the old dracut shell script handling usr=

crawford:
Booting with the following seems to work fine:

```
init=/usr/lib/systemd/systemd rootflags=rw mount.usrflags=ro BOOT_IMAGE=/boot/vmlinuz console=ttyS0,115200n8 console=tty0 root=LABEL=ROOT usr=PARTUUID=7130c94a-213a-4e5a-8e26-6cce9662f132
```

I blew away the files in root and rebooted. All of the generators ran and recreated the filesystem. Is there anything else I should test before merging?

marineam:
Nope, if the above kernel args work then I'm good. Merge away

PR #1252, opened 2015-05-12
Swap ntpd for systemd-timesyncd

marineam:
The ntpd code base is a can of worms and timesyncd has the advantage of respecting the NTP server provided via DHCP


mischief:
lgtm

brianredbeard:
Is this going to have an affect on ntp.org?

marineam:
@brianredbeard timesyncd will default to the same timeservers. Also users can mask timesyncd and enable ntpd if they wish.

PR #1254, opened 2015-05-13
sys-kernel/coreos-{sources,kernel}: bump to 4.0.3

mischief:
see also https://www.kernel.org/pub/linux/kernel/v4.x/ChangeLog-4.0.3


crawford:
LGTM

PR #1257, opened 2015-05-14
eclass/coreos-go: Updates for arm64

glevand:
This is the first set of patches toward adding go language support for arm64 builds.

This updates the coreos-go eclass to use the GOARCH environment variable which sets the target machine architecture.  GOARCH defaults to the host machine architecture, so has been working correctly for amd64 builds, but did not generate the correct binaries for arm64.

app-admin/sdnotify-proxy needed an update to build for arm64, so I took the opportunity to convert it to use the coreos-go eclass.  It can now be used as a test case for the coreos-go eclass.

Also included in the series are a few minor enhancements to the coreos-go eclass and app-admin/sdnotify-proxy.  Please consider.

-Geoff


eclass/coreos-go.eclass
```diff
@@ -9,41 +9,52 @@
...
-[[ ${EAPI} != "5" ]] && die "Only EAPI=5 is supported"
+[[ ${EAPI} != "5" ]] && die "${ECLASS}: Only EAPI=5 is supported"
 
 inherit flag-o-matic multiprocessing toolchain-funcs
 
 DEPEND="dev-lang/go"
 
+# @FUNCTION: go_get_arch
+# @USAGE: export GOARCH=$(go_get_arch)
+go_get_arch() {
```
marineam: conveniently so far these match portage's arch names, available in $ARCH. Should we just do `GOARCH="${ARCH}"`?

glevand: If we wrap it with go_get_arch() then if an arch is added that doesn't match we only need to fix  go_get_arch() and not all the ebuilds that use $ARCH directly.  What do you think?

marineam: They may so providing go_get_arch() sounds reasonable.

marineam:
LGTM, thanks!

PR #1259, opened 2015-05-15
dev-lang/go: Updates for arm64

glevand:
This is the second and final set of patches that add go language support for arm64 builds.

Go is no longer bootstrapped through a C compiler, but is now bootstrapped by go itself.  To support this a new package dev-lang/go-bootstrap is available that installs a version of go-1.4 for bootstraping.  Unfortunately, the upstream go-bootstrap has a bug when FEATURES=splitdebug, so we need a local fix.  I'll file a gentoo bug report for this.

For dev-lang/go I've created a new temporary ebuild, go-1.5.ebuild,  that builds the go compiler from the latest git checkout.  Once go-1.5 is available upstream and merged to portage-stable this ebuild can be removed.  Should we name this local one something like go-1.4.999 to avoid problems?

Please consider.

-Geoff


profiles/coreos/base/package.accept_keywords
```diff
@@ -115,3 +115,5 @@ dev-util/checkbashisms
 # fix for https://github.com/coreos/bugs/issues/324
 =net-misc/curl-7.41
 
+# Only needed until go-1.5 is available upstream
+dev-lang/go-bootstrap ~amd64
```
marineam: Might as well just set KEYWORDS=amd64 in the ebuild since it is here in the overlay.

dev-lang/go-bootstrap/go-bootstrap-1.4.1.ebuild
```diff
@@ -0,0 +1,83 @@
+# Copyright 1999-2015 Gentoo Foundation
```
marineam: Only need to include the one version of go-bootstrap that you want to use.

dev-lang/go/go-1.5.ebuild
```diff
@@ -0,0 +1,106 @@
+# Copyright 1999-2015 Gentoo Foundation
```
marineam: Use the name `go-9999.ebuild` since this is live. If you want to snapshot a particular version you can use something like `go-1.5_pre20150515` and set EGIT_COMMIT=<hash> in the ebuild.

marineam:
When you file a bug upstream please include a note and a link in the ebuild or commit message

dev-lang/go-bootstrap/go-bootstrap-1.4.2-r1.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	cd src
+	PATH="${GOBIN}:${PATH}" \
+		./run.bash --no-rebuild --banner || die "tests failed"
+}
+
+src_install()
+{
+	dodir /usr/lib/go1.4
+	exeinto /usr/lib/go1.4/bin
+doexe bin/*
```
marineam: This indent error is in the upstream ebuild but might as well fix it :)

metadata/md5-cache/dev-lang/go-bootstrap-1.4.1
```diff
@@ -0,0 +1,10 @@
+DEFINED_PHASES=compile install postinst prepare test
```
marineam: Prune these two metadata files, both are referring to ebuilds that are not here.

glevand:
Fixed the indentation and also removed metadata/md5-cache/dev-lang/go-bootstrap-1.4.1 which I missed.

Should be pretty close now...

marineam:
LGTM, thanks!

glevand:
Filed a bug report for the dev-lang/go-bootstrap problems: https://bugs.gentoo.org/show_bug.cgi?id=549598

Did you want me to mention it somewhere else?

marineam:
@glevand a comment in the ebuild would be good.

glevand:
Added the bug number.

-Geoff

PR #1261, opened 2015-05-19
Update go program ebuilds for arm64

glevand:
This series re-works the go language program ebuilds to allow building for arm64.  I have not finished rkt, as it is has a long and involved build procedure which I'll continue to work on.  

Included in the series are some minor fixups to the coreos-go eclass.

Since coreos-go eclass has a dependency on dev-lang/go and the current stable is go-1.4 I could remove several redundant DEPENDS from the ebuilds.

Please consider.

-Geoff


app-admin/locksmith/locksmith-9999.ebuild
```diff
@@ -23,20 +23,18 @@ LICENSE="Apache-2.0"
...
 IUSE=""
 
-DEPEND=">=dev-lang/go-1.2"
-
 src_compile() {
 	# work around gentoo hardened gcc incompatibilities with cgo
 	# see https://bugs.gentoo.org/show_bug.cgi?id=493328
 	if gcc-specs-pie; then
-		GOLDFLAGS="-extldflags -fno-PIC"
+		CGO_LDFLAGS+=" -extldflags -fno-PIC"
```
marineam: The eclass already disables PIE which I think is sufficient, is disabling PIC actually required?

coreos-base/coreos-cloudinit/coreos-cloudinit-9999.ebuild
```diff
@@ -23,27 +23,24 @@ LICENSE="Apache-2.0"
...
-	export CGO_CFLAGS="-I${ROOT}/usr/include"
-	export CGO_LDFLAGS="-L${ROOT}/usr/lib"
+src_prepare() {
+	coreos-go_src_prepare
+
+	ln -s ${S}/third_party/src/github.com/guelfey \
+		${GOPATH}/src/github.com/guelfey
 
 	if gcc-specs-pie; then
 		CGO_LDFLAGS+=" -fno-PIC"
```
marineam: I think this can be dropped.

eclass/coreos-go.eclass
```diff
@@ -11,7 +11,7 @@
 
 [[ ${EAPI} != "5" ]] && die "${ECLASS}: Only EAPI=5 is supported"
 
-inherit flag-o-matic multiprocessing toolchain-funcs
```
marineam: toolchain-funcs provides gcc-specs-pie iirc

eclass/coreos-go.eclass
```diff
@@ -75,7 +79,7 @@ coreos-go_src_compile() {
 coreos-go_src_install() {
 	debug-print-function ${FUNCNAME} "$@"
 
-	dobin "${GOBIN}"/*
+	dobin ${GOBIN}/*
```
marineam: Please leave quoted. Looks funny and in this case not strictly required but I prefer to err on the side of obsessively quoting when writing shell.

app-admin/locksmith/locksmith-9999.ebuild
glevand: In this case I just converted what was already there.  On reading bug 493328, there is some mention of disabling PIC, but the problem seems to be PIE, and no other ebuilds disable PIC, so I think we should remove this, then if a problem is found, disable PIC in the eclass.

marineam: Yeah, I think it was just that no PIC implies no PIE but itself PIC doesn't actually matter one way or the other because the final product is an executable, not a library.

glevand:
I split out the eclass changes into separate patches, updated locksmith and rebased to the latest.

marineam:
LGTM, thanks!

marineam:
@glevand and FYI, rkt is looking into rejiggering their build system using autoconf so once that lands it may make sense to just make sure the standard --build/--host/$CFLAGS/$LDFLAGS are plumbed correctly instead of using the go eclass.

https://github.com/coreos/rkt/pull/908

PR #1262, opened 2015-05-21
Backport Linux 4.0.1 to 647

marineam:
This updates our current stable branch to the kernel we currently ship in beta.

Should fix at least some of the issues reported in https://github.com/coreos/bugs/issues/320


crawford:
LGTM

PR #1263, opened 2015-05-21
coreos-base/coretest: bump to latest version

crawford:
Verify that networking in Docker works
Check for systemd-timesyncd instead of ntp


marineam:
[LGTM](http://lgtm.in/i/6FhGmH08q)

PR #1266, opened 2015-05-27
Revert "update_engine: lots of updates"

marineam:
This reverts commit 4dbfe48bd20b036a11c30538e3ce5cdfb5bb1715:

     - remove call to crossystem
     - support for payloads that include kernels
     - lots of refactoring/cleanups

The support for kernel payloads broke updates on systems that did not
already have the newer /boot/coreos/vmlinuz-{a,b} kernels.

Rolling back does restore the call to /usr/bin/crossystem but that call
failing shouldn't have any adverse behavior, just it will just add yet
another useless error message to our logs.


crawford:
LGTM

PR #1269, opened 2015-05-28
app-emulation/rkt: bump to v0.5.5 and add gc timer

crawford:
WIP


crawford:
Sticking with 0.5.5 for now.

jonboulle:
Looks OK

marineam:
Please update to a less aggressive schedule, I don't think we need to be running this every 10 minutes.

crawford:
@marineam I went with a 1d grace period and 12h periodic timer.

marineam:
[LGTM](http://lgtm.in/i/Akncv6Vfq)

PR #1272, opened 2015-06-02
sys-apps/systemd: bump to v220-coreos

crawford:
WIP. The commit hash will need to be updated once @mischief lands his domain-name fix.


marineam:
Please also update the header to reflect the upstream revision of systemd-9999.ebuild you are syncing with.

marineam:
looks good, so far

crawford:
Updated the commit hash.

jonboulle:
lgtm

PR #1276, opened 2015-06-09
sys-apps/systemd: add Xen and CCISS fixes

crawford:


mischief:
lgtm

marineam:
LGTM but we need to check on cciss too https://github.com/coreos/systemd/pull/9#issuecomment-110249839

crawford:
Updated to include CCISS fix.

marineam:
LGTM, can you poke upstream too or should I?

crawford:
I submitted a PR. I don't know if you want to make any more fuss about this though.

PR #1277, opened 2015-06-10
app-emulation/wa-linux-agent: bump to v2.0.13

crawford:


crawford:
Eventually I'd like to include the udev rule that was added.

mischief:
lgtm.

PR #1279, opened 2015-06-11
add e2fsprogs to seismograph dependencies

gtank:


marineam:
You need e2fsprogs, e2fsprogs-libs just provides libcom_err.so and libss.so

marineam:
Looks good, just need to update the git hash once https://github.com/coreos/seismograph/pull/27/ lands

PR #1282, opened 2015-06-12
Packet image OEM

stresler:
Feedback appreciated!  Also, let me know if there is a better way to do this PR.

### DIFF
--- a/coreos-base/oem-packet/files/cloud-config.yml
+++ b/coreos-base/oem-packet/files/cloud-config.yml
@@ -1,0 +1,41 @@
+#cloud-config
+
+coreos:
+  units:
+    - name: oem-reset-interfaces.service
+      command: start
+      content: |
+        [Unit]
+        Description=Reset the bond.
+        [Service]
+        Type=oneshot
+        ExecStart=/usr/share/oem/bin/reset-interfaces.sh
+    - name: oem-cloudinit.service
+      command: restart
+      runtime: yes
+      content: |
+        [Unit]
+        Description=Cloudinit from Packet metadata
+        After=oem-reset-interfaces.service
+        Requires=oem-reset-interfaces.service
+
+        [Service]
+        Type=oneshot
+        ExecStart=/usr/bin/coreos-cloudinit --oem=packet
+    - name: oem-phone-home.service
+      command: start
+      content: |
+        [Unit]
+        Description=Basic setup tasks to initialize the machine.
+        Wants=sys-devices-virtual-net-bond0.device
+        After=sys-devices-virtual-net-bond0.device
+
+        [Service]
+        Type=oneshot
+        ExecStart=/usr/share/oem/bin/phone-home.sh
+  oem:
+    id: packet
+    name: Packet
+    version-id: @@OEM_VERSION_ID@@
+    home-url: https://packet.net
+    bug-report-url: https://github.com/coreos/bugs/issues
--- a/coreos-base/oem-packet/files/phone-home.sh
+++ b/coreos-base/oem-packet/files/phone-home.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -e
+
+API_URL=$(curl -s https://metadata.packet.net/metadata | jq ".api_url" | sed -e 's/^"//'  -e 's/"$//')
+
+while ! wget --inet4-only --spider --quiet "${API_URL}/build_info.txt"; do
+	echo "$0: API unavailable: $API_URL" >&2
+	sleep 2
+done
+
+logger "Making a call to tell the packet API this machine is online."
+curl -X POST $API_URL/phone-home
+logger "This device has been announced to the packet API."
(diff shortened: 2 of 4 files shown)


coreos-base/oem-packet/files/phone-home.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+curl -s -o /tmp/jq http://images.packet.net/jq
```
marineam: This is terrifying. Either we should add jq to the image or perhaps ship a little go app in the OEM partition that does this instead of a shell script.

coreos-base/oem-packet/files/reset-interfaces.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
marineam: add `set -e` to ensure a failure aborts the script and an error is reported.

marineam: @crawford is this the best way to do this? What does onmetal do right now?

marineam: nm, found it: https://github.com/coreos/coreos-overlay/blob/master/coreos-base/oem-rackspace-onmetal/files/rename-interfaces.sh so basically the same

coreos-base/oem-packet/files/reset-interfaces.sh
```diff
@@ -0,0 +1,7 @@
...
+logger "Hard reset of all network interfaces due to upstream systemd-networkd issue: https://github.com/coreos/bugs/issues/36 ."
+ip link set enp1s0f0 down
+ip link set enp1s0f1 down
+ip link del bond0
+systemctl restart systemd-networkd
+sleep 15
```
marineam: is this sleep really required?

coreos-base/oem-packet/files/cloud-config.yml
```diff
@@ -0,0 +1,41 @@
+#cloud-config
+
+coreos:
+  units:
+    - name: oem-reset-interfaces.service
```
crawford: Should this run after the oem-cloudinit service? Otherwise, it will reconfigure with the original configs.

stresler: Well, no, this is actually a limitation of our system that we will eventually need to correct.  When the machine comes back online post-reboot the machine doesn't have network at all until this runs. SO trying to run oem-cloudinit.service  will fail because it cannot reach our metadata service.  

The problem here is that this ends up running twice. Would it be better to place this .service file directly with ebuild so that it is disconnected from cloud-init altogether?

coreos-base/oem-packet/files/reset-interfaces.sh
stresler: The virtual interface takes time to come up and systemctl seems to return after kicking it off. This makes sense, because systemctl has finished it's job, it is ambivalent success/failure of the the service it calls, and the subsequent job that announces the machine to our API requires network.  

I can think of two alternatives.  I can force the second job to wait until network is up, and is probably better than this sleep.  

Or I can make a target that is statisfied by the bond0 having link.

coreos-base/oem-packet/files/reset-interfaces.sh
stresler: Now that I understand 'udevadm test' I prefer this.  I'll make the alterations.

coreos-base/oem-packet/files/phone-home.sh
stresler: I'm open to either, which is preferred? I did it this way because I presumed you wouldn't want some random bash json decoder in the CoreOS codebase.

coreos-base/oem-packet/files/reset-interfaces.sh
stresler: I'm looking further into this. Just running that instead of what we have does not bring our bond configuration up properly. This could be a difference in how we're doing network over rackspace.

coreos-base/oem-packet/files/reset-interfaces.sh
marineam: Adding Wants/After=sys-devices-virtual-net-bond0.device to the next unit should be more reliable.

coreos-base/oem-packet/files/reset-interfaces.sh
marineam: yeah, I didn't mean to imply rackspace should be copied, the issue is slightly different. Just add the set -e

coreos-base/oem-packet/files/phone-home.sh
marineam: Posted https://github.com/coreos/coreos-overlay/pull/1291

The alternative would be to ship the jq binary in the OEM partition instead of the base image but the base image should be fine.

coreos-base/oem-packet/files/reset-interfaces.sh
stresler: Works in testing. Can remove the sleep 15 with the altered dependency. Update coming soon. Thanks!

marineam:
LGTM, could you squash the commits please?

stresler:
Squashed

crawford:
LGTM. Thanks!

PR #1288, opened 2015-06-15
coreos-overlay: Remove unused sys-fs/fuse package

glevand:
This PR removes dependencies on sys-fs/fuse from the image meta-packages and removes the sys-fs/fuse package.  Another PR for portage-stable will remove some packages that will become unused with the removal of sys-fs/fuse.

sys-fs/fuse is a dependency of sys-fs/libfat, and hard-host-depends-0.0.1.ebuild mentions that sys-fs/libfat is 'needed to create and sign images', but I could not find any actual use of libfat so I removed it from the hard-host-depends RDEPEND list and created a portage-stable patch to remove the sys-fs/libfat package.


marineam:
Please bump the revision in the symlink for  coreos-base/coreos-dev and coreos-base/hard-host-depends

It looks like you are hand-editing files in metadata/md5-cache/* but those should only be automatically generated and are ignored if they are stale so leaving them doesn't hurt. Over time if too many are stale emerge can start to get slower in a fresh SDK or after a repo sync, when that happens we just need to run ./update_metadata in scripts.

glevand:
@marineam Bumped:
  coreos-dev-0.1.0 -> coreos-dev-0.1.1
  hard-host-depends-0.0.1 -> hard-host-depends-0.0.2

Let me know if you think other versions would be better.

glevand:
@marineam Updated as requested.

marineam:
LGTM!

crawford:
I am reverting this change for now (https://github.com/coreos/coreos-overlay/pull/1294). As it turns out, open-vm-tools has an undeclared dependency on fuse. We need to eventually fix the open-vm-tools ebuild.

glevand:
@crawford Did you have a problem at build, or at run-time?

open-vm-tools checks for installed fuse in it's configure script using pkgconfig.  If you did not do a fresh system build or cleanup stale installed fuse files before your open-vm-tools build then that may have been your problem.  Maybe something like this:

  rm {,/build/*}{/usr/include/fuse.h,/usr/lib64/pkgconfig/fuse.pc}

I should have mentioned that a file clean up may be needed in my original PR.

crawford:
@glevand it failed to build on our [production builder](https://buildbot-master1.coreos.marineau.org/builders/packages_master/builds/412/steps/shell_2/logs/stdio), so it would have been a clean environment. This was the failure:

```
libtool: link: x86_64-cros-linux-gnu-gcc -Dvmblock_fuse -U_XOPEN_SOURCE -D_XOPEN_SOURCE=600 -DUSERLEVEL -D_FILE_OFFSET_BITS=64 -D_FILE_OFFSET_BITS=64 -I/usr/include/fuse -I/build/amd64-usr/usr/include/glib-2.0 -I/build/amd64-usr/usr/lib64/glib-2.0/include -I../modules/shared/vmblock -I. -O2 -pipe -mtune=generic -Wall -Wno-pointer-sign -Wno-unused-value -fno-strict-aliasing -Wno-unknown-pragmas -Wno-uninitialized -Wno-deprecated-declarations -Wno-unused-but-set-variable -Wl,-O1 -Wl,--as-needed -Wl,-rpath -Wl,/usr/share/oem/lib64 -o vmware-vmblock-fuse util.o fsops.o main.o block.o stubs.o -pthread  -lfuse -ldl -lglib-2.0 -pthread
/usr/libexec/gcc/x86_64-cros-linux-gnu/ld: cannot find -lfuse
collect2: error: ld returned 1 exit status
```

glevand:
@crawford looking at the build log, I'm not certain, but it looks like your problem could be stale files.  The build did 'build_packages --skip_chroot_upgrade' which seems to re-use the SDK emerged packages and then just builds any new or changed board packages.  I don't think it does an uninstall of removed packages. That would leave the host and possibly the board fuse.h and fuse.pc there I think.  Also, a common bug is for the configure script to look on the host, not the target when it does its test for fuse.h and fuse.pc, so the test for fuse could be incorrectly succeeding because of that.

I had very similar problems with stale files when testing my original patches.  When I removed the stale fuse files the open-vm-tools configure script worked OK, and built open-vm-tools without fuse support.

crawford:
@glevand I think you're right. It looks like I didn't clear out the correct build directory. Either way, we have our Alpha build. I'll go ahead and start cleaning up open-vm-tools.

glevand:
@crawford I llooked into the problem.  The SDK image (coreos-sdk-amd64-712.0.0.tar.bz2) still has fuse 'installed', so the SDK will have /usr/lib64/pkgconfig/fuse.pc.  The open-vm-tools configure is using that in its test for fuse.  We need fix the open-vm-tools ebuild to make configue use the board's fuse.pc, which would not be present with my fuse package removal patches.  We could also just remove fuse from the SDK tar file.

crawford:
@glevand thank you for looking into this. I'm just going to fix the build scripts to use the --without-fuse configure flag. @marineam fixed this in the past, but the patch must have been dropped when I updated the package.

glevand:
@crawford The --without-fuse option should be fixed in my PR https://github.com/coreos/coreos-overlay/pull/1307.

PR #1289, opened 2015-06-15
coreos-overlay: Add arm64 keywords

glevand:
Add the arm64 keyword to any ebuilds that need it.

### DIFF
--- a/app-admin/flannel/flannel-9999.ebuild
+++ b/app-admin/flannel/flannel-9999.ebuild
@@ -4,9 +4,9 @@
 EAPI=4
 
 if [[ "${PV}" == 9999 ]]; then
-	KEYWORDS="~amd64"
+	KEYWORDS="~amd64 ~arm64"
 else
-	KEYWORDS="amd64"
+	KEYWORDS="amd64 arm64"
 fi
 
 inherit systemd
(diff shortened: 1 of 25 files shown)


app-emulation/open-vm-tools/open-vm-tools-9.10.0.ebuild
```diff
@@ -17,7 +17,7 @@ EGIT_SOURCEDIR="${WORKDIR}"
 
 LICENSE="LGPL-2"
 SLOT="0"
-KEYWORDS="amd64 ~x86"
+KEYWORDS="amd64 arm64 ~x86"
```
marineam: open-vm-tools is only applicable to VMware so I doubt this makes sense. Probably should fix by using `arm64? (  app-emulation/open-vm-tools )` or similar in board-packages.

app-emulation/xenserver-pv-version/xenserver-pv-version-6.2.0.ebuild
```diff
@@ -12,7 +12,7 @@ SRC_URI=""
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="amd64"
+KEYWORDS="amd64 arm64"
```
marineam: Only applicable to XenServer deployments, unlikely to be useful on ARM I assume. Add a conditional depend to the coreos ebuild instead.

glevand:
@marineam I removed the two patches for open-vm-tools and xenserver-pv-version from this series and will include new patches with your recommended changes in another PR.

marineam:
LGTM

PR #1291, opened 2015-06-17
coreos: add jq to base image

marineam:
Needed by an OEM and it is a small/general-enough thing that it seems
better to ship in the image than try to stick it into the OEM partition.

For https://github.com/coreos/coreos-overlay/pull/1282


mischief:
lgtm

crawford:
LGTM

PR #1297, opened 2015-06-18
coreos-overlay: Updates for arm64 kernel

glevand:
Adds arm64 kernel support.

The 3rd patch adds commands to coreos-kernel.eclass that generate a kernel minimal defconfig for reference.  The trend by kernel developers is to use minimal configs over full configs as they are generally considered easier to maintain.  A minimal config has all the options that equal the kconfig default options removed.


eclass/coreos-kernel.eclass
```diff
@@ -128,6 +128,13 @@ coreos-kernel_src_prepare() {
 coreos-kernel_src_configure() {
 	# Use default for any options not explitly set in defconfig
 	yes "" | kmake oldconfig
+
+	# For convinence, generate a minimal defconfig of the build
+	local f=$(find_defconfig)
+	f=${f##/*/}
+	kmake savedefconfig
+	einfo "Saving minimal kernel defconfig as ${KBUILD_OUTPUT}/$f"
```
marineam: I'd just leave it as defconfig instead of matching the config in the ebuild tree, a number of names are searched so it isn't as predictable as "{KBUILD_OUTPUT}/defconfig"

glevand:
@marineam updated as requested to just output as defconfig.

marineam:
LGTM

PR #1298, opened 2015-06-19
replace TMPDIR with DOCKER_TMPDIR

cdwertmann:
docker 1.6 seems to ignore TMPDIR and puts tmp files into /var/lib/docker/tmp/

### DIFF
--- a/app-emulation/docker/files/docker.service
+++ b/app-emulation/docker/files/docker.service
@@ -5,7 +5,6 @@ After=docker.socket early-docker.target network.target
 Requires=docker.socket early-docker.target
 
 [Service]
-Environment=TMPDIR=/var/tmp
 EnvironmentFile=-/run/flannel_docker_opts.env
 MountFlags=slave
 LimitNOFILE=1048576


mischief:
@cdwertmann nice catch. can you change the commit message to be `app-emulation/docker: replace TMPDIR with DOCKER_TMPDIR`? otherwise lgtm.

marineam:
Well, the point of TMPDIR was to get docker to stop using `/tmp` which is tmpfs. So if the default is now under /var we can just drop this.

marineam:
and considering /var/lib/docker may be a different filesystem than /var/tmp leaving it as the docker default is probably better anyway.

crawford:
+1 for yanking it out altogether.

mischief:
removing it sounds good. if it is removed, docker will use /var/lib/docker/tmp.

cdwertmann:
I removed TMPDIR entirely now.

I came across this when looking into an issue with using Docker registry v2 and the docker daemon filling up the disk with huge temp files in /var/lib/docker/tmp. When the disk is full, etcd2 writes corrupted .WAL files and the node cannot re-join the cluster. Putting docker temp files into the tmpfs would not solve the problem of docker not cleaning up after itself, but would at least prevent etcd2 from corruption. I guess I'll open another ticket for that.

crawford:
LGTM. Please squash your commits.

cdwertmann:
Squashed.

marineam:
LGTM, thanks!

PR #1301, opened 2015-06-23
coreos-base: remove dependencies on a few unused python libraries

marineam:
As far as I can tell we do not need pydbus, pygobject, or dbus-python in
the SDK or the dev image.


crawford:
Ok.

PR #1303, opened 2015-06-23
docker: re-add kernel dependency

marineam:


crawford:
Should the 1.7.0 ebuild also be bumped?

marineam:
Doesn't change what is installed so no need.

crawford:
OK. LGTM.

PR #1305, opened 2015-06-23
Migrate to gentoo-functions

marineam:
This is a migration upstream has partially completed and it removes the need for our special efunctions package. It also fixes an issue with gcc-config: https://bugs.gentoo.org/show_bug.cgi?id=547962


mischief:
lgtm

PR #1306, opened 2015-06-23
media-libs/freetype: remove old ebuild

marineam:
This ebuild breaks with the latest upstream eclasses but we don't need
it. Drop it and the one python library that depended on it.


crawford:
LGTM

PR #1308, opened 2015-06-24
coreos: add lshw to base image

marineam:
An easy tool for reporting exactly what hardware we are dealing with.

Depends on https://github.com/coreos/portage-stable/pull/226


crawford:
LGTM

PR #1311, opened 2015-06-24
Bring in custom selinux work

mjg59:
There's a few places where we have to diverge from upstream


mjg59:
Ok, I've split those up into the unmodified code and the local changes.

marineam:
In the initial import please include the eclass so the following commit also shows the changes you make there. Also please prune the ebuilds down to just the version you intend to use/patch, otherwise it is easy to mix up which ebuild actually mattered the next time someone touches the package.

The coreos-kernel ebuild revision should be bumped along with the config update.

coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -119,6 +119,7 @@ RDEPEND="${RDEPEND}
 	net-misc/ntp
 	net-misc/rsync
 	net-misc/wget
+	sec-policy/selinux-virt
```
marineam: I missed this before, coreos needs a revision bump.

marineam:
Are there any differences we should expect from a system running with selinux enabled with the default policy from a system that doesn't have selinux in the kernel? I'm not sure what potential oddities there may be when we flip this switch.

beyond the tmpfiles rules for initializing root filesystem symlinks LGTM

mjg59:
At this point we should expect no difference other than the generation of AVCs if rkt violates an selinux constraint. We'll flip to enforcing later.

marineam:
ok :)

PR #1315, opened 2015-06-26
systemd: fix networkd bug for bond interfaces

marineam:
Pull in https://github.com/coreos/systemd/pull/12, required for https://github.com/coreos/coreos-overlay/pull/1282


mischief:
lgtm

PR #1317, opened 2015-06-29
coreos-init: update for logrotate support

marineam:
btmp and wtmp will now be properly rotated, yay!

Masking of logrotate configs has moved to just apply to boards, leaving
them in the SDK can be a useful reference.


mischief:
lgtm

PR #1318, opened 2015-06-29
coreos-overlay: Move to portage-stable

glevand:
The nmap and iperf packages were updated for arm64 support and moved to portage-stable.  This PR removes the old packages from coreos-overlay.

nmap was updated in https://github.com/coreos/portage-stable/pull/206.
iperf was updated in https://github.com/coreos/portage-stable/pull/231


marineam:
LGTM

PR #1320, opened 2015-06-29
ec2: disable new-style interface naming on Amazon

marineam:
Before we can enable ixgbevf devices by default we need to prevent the
interface names from changing names and surprising folks. On the down
side this may surprise anyone manually enabling ixgbevf on their
instances but I expect that to be the smaller population.

Required for https://github.com/coreos/scripts/pull/417 https://github.com/coreos/bugs/issues/273


crawford:
LGTM

PR #1325, opened 2015-06-30
Add selinux build fixes

mjg59:
There were a couple of build failures due to selinux packages accidentally
relying on host packages.

### DIFF
--- a/app-admin/setools/setools-3.3.8-r7.ebuild
+++ b/app-admin/setools/setools-3.3.8-r7.ebuild
@@ -87,6 +87,7 @@ src_configure() {
 		--with-java-prefix=${JAVA_HOME} \
 		--disable-selinux-check \
 		--disable-bwidget-check \
+		--with-sepol-devel=${ROOT}/usr \
 		$(use_enable python swig-python) \
 		$(use_enable java swig-java) \
 		$(use_enable X swig-tcl) \
--- a/coreos/config/env/app-admin/setools
+++ b/coreos/config/env/app-admin/setools
@@ -1,2 +1,2 @@
-export ac_cv_file__usr_lib64_libsepol_a=yes
+export ac_cv_file__build_amd64_usr__usr_lib64_libsepol_a=yes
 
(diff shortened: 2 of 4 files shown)


sys-apps/policycoreutils/policycoreutils-2.4.ebuild
```diff
@@ -100,6 +100,7 @@ src_compile() {
 			INOTIFYH="$(usex dbus)" \
 			SESANDBOX="n" \
 			CC="$(tc-getCC)" \
+			PREFIX="${ROOT}" \
```
marineam: is this prefix value only significant during build? prefix generally means the runtime path (such as /usr) so if it influences runtime there could be problems.

coreos/config/env/app-admin/setools
```diff
@@ -1,2 +1,2 @@
-export ac_cv_file__usr_lib64_libsepol_a=yes
+export ac_cv_file__build_amd64_usr__usr_lib64_libsepol_a=yes
```
marineam: hm, this seems icky and will break for different arch/boards. I suppose it is ok for now to fix the build but something cleaner would be nice.

app-admin/setools/setools-3.3.8-r7.ebuild
```diff
@@ -87,6 +87,7 @@ src_configure() {
 		--with-java-prefix=${JAVA_HOME} \
 		--disable-selinux-check \
 		--disable-bwidget-check \
+		--with-sepol-devel=${ROOT}/usr \
```
marineam: Is this only used to find the static libsepol.a? just want to be sure runtime is ok.

mjg59: Yeah, this is only used to define the search path during configure.

sys-apps/policycoreutils/policycoreutils-2.4.ebuild
mjg59: Yeah, it (and things derived from it) are only referenced in the makefile, not the code.

marineam:
Ok, LGTM to me then. Eventually I'd like to see the `ac_cv_file__build_amd64_usr__usr_lib64_libsepol_a` thing cleaned up, either by improving the configure script or by copying libsepol.a or some other scheme but no need to deal with that right now.

PR #1326, opened 2015-06-30
Enable kernel lockdowns in Secure Boot environments

mjg59:
Secure Boot is easy to work around unless the kernel restricts
userspace's ability to modify the kernel. Add kernel patches from Fedora
that do this.


marineam:
Now that we are patching the source we should drop `K_NOUSENAME="yes"` and `K_NOSETEXTRAVERSION="yes"` so `-coreos`, `-coreos-r1`, and so on get appended to the version. Also please bump the rev of coreos-sources and coreos-kernel to revision `-r1`

marineam:
actually, the coreos-kernel eclass needs to be updated to be able to find the kernel source first, I'll do that in another PR

marineam:
Need to update the sources revision in the coreos-kernel ebuild. I didn't bind them together directly because kernel configs change faster than sources.

mjg59:
Uhp. Yeah, fixed that.

marineam:
LGTM

PR #1327, opened 2015-06-30
coreos-kernel: prepare for using patched kernels

marineam:
We will be carrying some patches so the version of the source code will
no longer be simply the upstream mainline version. A -coreos or
-coreos-r1 and so forth will be appended. A new variable defining the
source revision (e.g. -r1) has been added so we can continue to bump the
coreos-kernel revision independently of coreos-sources for minor things
like config updates.

Required for https://github.com/coreos/coreos-overlay/pull/1326


mischief:
lgtm

PR #1329, opened 2015-07-01
Fix policycoreutils build

mjg59:
There was still some python leaking into this - skip building sepolicy
to avoid issues with cross-compilation.


marineam:
LGTM

PR #1331, opened 2015-07-01
Remove calls to host tools during selinux policy build

mjg59:
SELinux policies were attempting to run the host checkmodule and semodule
commands. The former is easy to fix via pointing them at the build root, the
latter we skip entirely because we don't want to install the policy at this
point - we'll do that during image build.


marineam:
Ok, if it is running commands this is another cross-compile problem and the SDK will need to provide the native checkmodule. But LGTM for the sake of getting the build fixed asap :)

PR #1332, opened 2015-07-01
sys-kernel/coreos-kernel: enable sfc and mtd drivers

mischief:
fixes coreos/bugs#400


marineam:
Just need to bump the revision to -r3 I believe

mischief:
@marineam ptal

marineam:
LGTM

PR #1335, opened 2015-07-02
selinux: make selinux optional based on USE flags, disable for now.

marineam:
Some issues still to work out with tmpfiles and logind.


glevand:
I tested this with arm64 builds and it seems to work as expected.  The selinux related packages that were failing builds, were no longer built.

LGTM

crawford:
LGTM

PR #1339, opened 2015-07-04
sys-apps/systemd: enable nat use flag for ip masquerading in networkd

mischief:
fix half of coreos/bugs#404


marineam:
No need to rev-bump for use flag changes, the use flag itself invalidates the old binary package.

marineam:
but otherwise LGTM

mischief:
fixed the rev.

PR #1343, opened 2015-07-08
ccache and systemd build cleanups

marineam:
Kernels are now zillions times happier with ccache. systemd is improved some, the majority of its cpu time is now stuck in lto (a good thing, cuts down on systemd binary sizes a lot) and libtool (ugg, seriously shell is a terrible language to write low level tools in).

Depends on https://github.com/coreos/portage-stable/pull/236

### DIFF
--- a/eclass/coreos-kernel.eclass
+++ b/eclass/coreos-kernel.eclass
@@ -92,8 +92,12 @@ update_bootengine_cpio() {
 }
 
 kmake() {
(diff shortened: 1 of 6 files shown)


crawford:
It looks like some other commits came along for the ride. The commit messages could use a little spell check but otherwise LGTM.

marineam:
Those extras are the systemd "build cleanups" ;-)

PR #1344, opened 2015-07-08
util-linux: don't bother excluding misc small utilities

marineam:
We may not *need* things like fsck.minix but it does get the systemd
test suite to pass. setarch should be there. fdformat and others are no
longer built by the ebuild.


crawford:
Are these tools also going to wind up in the image?

marineam:
Yes, we can any to the prod specific install mask if they are worth continuing to exclude.

crawford:
Does it make sense to continue to exclude these utilities in the image (but leave them in the SDK)?

marineam:
Perhaps, I don't really care one way or the other.

crawford:
Meh, LGTM.

PR #1345, opened 2015-07-08
sys-apps/systemd: include parsing fixes

crawford:


marineam:
LGTM, although maybe -r4 makes more sense since -r2 and -r3 in master pointed to versions that are now effectively older than this one.

crawford:
-r4 it is

PR #1349, opened 2015-07-09
systemd: revise python dependencies

marineam:
Dug more into the build process and python is not strictly required for
man pages, it is only used to create an index which isn't significant.
Only call setup_python when features that *do* need python are enabled.

Fixes bootstrap_sdk which builds systemd with minimal use flags.


crawford:
LGTM

PR #1350, opened 2015-07-09
profiles: remove unneeded entries for openssl

crawford:
Depends on https://github.com/coreos/portage-stable/pull/237.


mischief:
Lgtm

PR #1354, opened 2015-07-10
profiles: use compiler hash for ccache

marineam:
By default ccache checks the compiler's mtime and size but that gets
thrown off by reinstalls, including from binary packages. The
alternative mode reads the compiler binary itself instead. In theory
that may be slower but in reality both modes are effectively the same
speed. ccache will now work under catalyst


crawford:
LGTM

PR #1358, opened 2015-07-13
coreos-overlay: Use latest vim for cross compile fix

glevand:
Specify vim-7.4.712 and vim-core-7.4.712 in package.accept_keywords
to pickup needed upstream cross-compile fix.

  https://bugs.gentoo.org/show_bug.cgi?id=473372

Fix build errors like these:

  checking for tgetent()... configure: error: NOT FOUND!
  You need to install a terminal library; for example ncurses.

Depends on https://github.com/coreos/portage-stable/pull/241 for vim ebuild updates.


marineam:
LGTM

PR #1359, opened 2015-07-14
coreos-kernel: add -fstack-check=no to kernel build

marineam:
In addition to enabling PIE and stack protector, hardened compilers also
enable the -fstack-check option which I had previously overlooked,
conflating it with -fstack-protector which the kernel build already
handles properly. This is the second time I hit this trying to use
ccache, I forgot that -nopie was insufficent for kernel builds. Last
time around reverted in cd043688f09b22ccf245ffd11f8a22a5ff31e577.

At least this time I finally dug deep enough to find the cause instead
of setting myself up for repeating history in another 6 months. For
reference the issue is that the Go 1.3.x runtime can call clock_gettime
with a very small stack. If the vDSO library was built with
-fstack-check it will attempt to write 0 to a location beyond the end of
the very small stack, triggering SEGV:

    Dump of assembler code for function __vdso_clock_gettime:
       0x00007ffff7ffaa50 <+0>:     push   %rbp
       0x00007ffff7ffaa51 <+1>:     mov    %rsp,%rbp
       0x00007ffff7ffaa54 <+4>:     push   %r14
       0x00007ffff7ffaa56 <+6>:     push   %r13
       0x00007ffff7ffaa58 <+8>:     push   %r12
       0x00007ffff7ffaa5a <+10>:    push   %rbx
       0x00007ffff7ffaa5b <+11>:    sub    $0x1038,%rsp
    => 0x00007ffff7ffaa62 <+18>:    orq    $0x0,(%rsp)
       0x00007ffff7ffaa67 <+23>:    add    $0x1020,%rsp


philips:
lgtm, I am trusting you on this one :)

PR #1362, opened 2015-07-14
sys-apps/systemd: sync with coreos/systemd v222

mischief:
pulls in changes that fix coreos/bugs#408 and fix a client's problem with multiple routers over dhcp.


crawford:
LGTM

Issue #1363, opened 2015-07-14
Consider starting up rkt MDS socket unit by default


eyakubovich:
Brought up in https://github.com/coreos/rkt/pull/1141.

rkt metadata service support socket activation, we should consider enabling it by default in CoreOS.

UserTaken:
Enabled in 758.1.0 so this can be closed.

PR #1364, opened 2015-07-14
app-emulation/rkt: bump to v0.7.0

### LINKED ISSUE #1360: app-emulation/rkt: bump version and adapt to new build system

crawford:
Fixes #1360.


crawford:
We might want to fix https://github.com/coreos/coreos-overlay/issues/1363 while we are here.

crawford:
Fixes https://github.com/coreos/coreos-overlay/issues/1360 and https://github.com/coreos/coreos-overlay/issues/1363.

steveeJ:
The ebuild is missing the dependency to dev-perl/Capture-Tiny.

crawford:
@steveeJ yep. Fixed.

crawford:
This PR depends on us updating Perl and including some depenencies in the coreos-overlay/portage-stable repos.

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -28,30 +30,42 @@ IUSE=""
 
 DEPEND=">=dev-lang/go-1.3
 	app-arch/cpio
-	sys-fs/squashfs-tools"
+	sys-fs/squashfs-tools
+	dev-perl/Capture-Tiny"
```
marineam: This will need to be added to coreos-base/sdk-depends in order to include in the SDK. Sadly portage does not yet know when to build build dependencies into $ROOT and when to put them into /

crawford: https://github.com/coreos/coreos-overlay/pull/1365

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -28,30 +30,42 @@ IUSE=""
...
 RDEPEND="!app-emulation/rocket"
 
-src_unpack() {
-	local cache="${S}/stage1/rootfs/usr_from_coreos/cache"
+src_configure() {
+	local myeconfargs=(
+		--with-stage1=host
```
marineam: Shouldn't this be --with-stage1=coreos?

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -28,30 +30,42 @@ IUSE=""
...
 	insinto /usr/share/rkt
-	doins "${S}/bin/stage1.aci"
+	dobin "${S}/build-${P}/bin/stage1.aci"
 
 	systemd_dounit "${FILESDIR}"/${PN}-gc.service
 	systemd_dounit "${FILESDIR}"/${PN}-gc.timer
+	systemd_dounit "${S}"/dist/init/systemd/${PN}-metadata.service
+	systemd_dounit "${S}"/dist/init/systemd/${PN}-metadata.socket
+
+	systemd_enable_service default.target ${PN}-metadata.socket
```
marineam: sockets.target

app-emulation/rkt/rkt-9999.ebuild
crawford: When I used that option, the build fails with:

```
gpg: can't open `/build/amd64-usr/var/tmp/portage/app-emulation/rkt-0.7.0/work/rkt-0.7.0/build-rkt-0.7.0/tmp/usr_from_coreos/pxe.img.a681903e9274fc3197c6c11a09e4121c.sig': No such file or directory
```

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -5,22 +5,19 @@ EAPI=5
...
 
 if [[ "${PV}" == 9999 ]]; then
     KEYWORDS="~amd64"
 else
-    CROS_WORKON_COMMIT="40ced98c320c056e343fe9c3eaeb90a4ff248936" # v0.5.5
+    CROS_WORKON_COMMIT="9579f4bf57851a1a326c81ec2ab0ed2fdfab8d24" # v0.7.0
     KEYWORDS="amd64"
 fi
 
-# Must be in sync with stage1/rootfs/usr_from_coreos/cache.sh
```
alban: This file does not exist anymore with rkt 0.7.7. It is replaced by [stage1/usr_from_coreos/usr_from_coreos.mk](https://github.com/coreos/rkt/blob/master/stage1/usr_from_coreos/usr_from_coreos.mk#L2):
```
UFC_IMG_RELEASE := "709.0.0"
UFC_IMG_URL := "http://alpha.release.core-os.net/amd64-usr/$(UFC_IMG_RELEASE)/coreos_production_pxe_image.cpio.gz"
```
This CoreOS image and the signature will be downloaded if `./configure --with-stage1=coreos` is used.

alban: /cc @krnowak

app-emulation/rkt/rkt-9999.ebuild
alban: So there are two possibilities:
- Use `--with-stage1=host` and the build will not need to download an older version of the CoreOS image from internet. Instead, it will pick at run-time systemd and other binaries from the host. This option was added for Linux distributions such as Fedora and Debian because they forbid internet access during the build process, and forbid reusing pre-compiled binaries from third-parties (CoreOS) instead of compiling from source.
- Use `--with-stage1=coreos` (as it was before and it is still the default I think) and it will attempt to download the CoreOS image. Since ebuild don't allow internet access during the build, it should be pre-downloaded in `rkt-9999.ebuild` but the URL must be fixed as commented below.

/cc @krnowak

crawford: But doesn't that mean that rkt 1.2.3 will potentially run differently depending on your underlying CoreOS version?

marineam: @alban --with-stage1=host is probably the right path but unless I missed something it pulls from `/` while we need to give it an arbitrary $ROOT, the build host is never has anything to do with the target we are building for.

Also `host` attempts to do thing like run `systemctl --version` which it must not do in our case as the build host's architecture may not match the targets.

marineam: The quickest way forward probably would be to use the pre-built `coreos` (pairing rkt with a previously tested stage1 is perhaps safest anyway), we just need to skip the downloads and such and have the build trust the ebuild gives it the right thing. Longer term creating stage1 from an arbitrary $ROOT or at run-time is ideal.

alban: `--with-stage1=host` pulls from `/` at run-time (when commands like `rkt run ...` are executed), not at build-time. So the builder architecture or $ROOT should not be relevant.

But I agree that the pre-built `coreos` is more tested for now, so safest indeed.

@crawford yes, with `host`, rkt 1.2.3 will potentially run differently depending on the underlying CoreOS version: when the host switches to a new version of systemd, rkt containers will automatically use the new systemd even if rkt is not updated. This puts more burden on the distribution...

crawford:
Superseded by https://github.com/coreos/coreos-overlay/pull/1392.

PR #1366, opened 2015-07-14
sys-apps/ignition: update to the latest

crawford:


vcaputo:
lgtm

crawford:
@vcaputo I removed Type=ext4 from the mount.

vcaputo:
cool, lgtm, I like how the guid approach to "first boot?" turned out.

PR #1370, opened 2015-07-15
coreos-overlay: dev-util/bsdiff: Change suffix sort to sais-lite

glevand:
To avoid dependencies on dev-libs/libdivsufsort and dev-util/cmake convert bsdiff
from using the divsufsort suffix sort to using the sais-lite suffix sort from [1].

The need for an alternative suffix sort is discussed in [2] through [4].  A
dependency on dev-util/cmake is problematic as cmake cannot be cross-compiled and
a switch to sais-lite avoids this dependency.

 [1] https://sites.google.com/site/yuta256/sais
 [2] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=409664
 [3] https://github.com/sparkle-project/Sparkle/pull/515
 [4] https://chromium.googlesource.com/chromiumos/third_party/bsdiff/+/e2b85ce4ea246ca804ee2a9e18005c44e193d868

I have other patches ready that remove the libdivsufsort package the cmake dependency in hard-host-depends, and the cmake package.


marineam:
Interesting, given the impact on update_engine I'll take some time to try this out myself.

marineam:
This is triggering a segfault for me, given the command:
```
$ bsdiff /usr/libexec/git-core/git-interpret-trailers /build/amd64-usr/usr//libexec/git-core/git-interpret-trailers patch
```

GDB bt:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7975544 in __memcmp_sse4_1 () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff7975544 in __memcmp_sse4_1 () from /lib64/libc.so.6
#1  0x0000555555555057 in search (st=0, pos=<synthetic pointer>, en=1929944, 
    newsize=1905376, new=0x55555575a010 "\177ELF\002\001\001", 
    oldsize=1929944, old=0x7ffff7e14010 "\177ELF\002\001\001", 
    I=0x7ffff6964010) at bsdiff.c:74
#2  main (argc=<optimized out>, argv=0x7fffffffe0a8) at bsdiff.c:182
```

Valgrind:
```
==10436== Invalid read of size 1
==10436==    at 0x4C319AC: __memcmp_sse4_1 (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==10436==    by 0x109056: main (in /usr/bin/bsdiff)
==10436==  Address 0x19464705588688 is not stack'd, malloc'd or (recently) free'd
==10436== 
==10436== 
==10436== Process terminating with default action of signal 11 (SIGSEGV)
==10436==  General Protection Fault
==10436==    at 0x4C319AC: __memcmp_sse4_1 (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==10436==    by 0x109056: main (in /usr/bin/bsdiff)
```

glevand:
@marineam Sorry, I had a type cast error that was causing the segfault.  Fixed and rebased to the latest.

Please try again to evaluate with update_engine.

marineam:
Works for me now! It does seem to be slower than it used to be but that isn't a big deal. Can always revisit this later if needed.

LGTM

PR #1372, opened 2015-07-16
coreos-overlay: Update net-fs/nfs-utils for arm64

glevand:
Adds the updated nfs-utils-1.3.2 ebuild from upstream to get build fixes for arm64 and forward ports the local changes that were applied to nfs-utils-1.2.9.

@marineam @mischief The systemd setup in src_install() was quite different than was done in the 1.2.9 ebuild.  I am not entirely confident that everything gets installed correctly, so please review src_install() with extra scrutiny.


glevand:
@marineam Rebased to latest.  Please consider.

marineam:
LGTM

PR #1373, opened 2015-07-16
Revert "coreos-devel/sdk-depends: add Capture-Tiny"

marineam:
Reverts coreos/coreos-overlay#1365

Updating perl, which includes a build system that isn't completely naive of cross-compiling, in turn causes the cmake build system to realize it is being cross-compiled (previously it was blissfully unaware), thus causing cmake to fail because its ebuild isn't setting things up right for cross compiles.

Temporarily roll back perl to a release building.


mischief:
lgtm

marineam:
nm, this looked related but something else is wrong.

PR #1374, opened 2015-07-16
profiles: fix path to coreos-overlay

marineam:
Missed this reference to /usr/local/portage in a recent scripts change:

https://github.com/marineam/coreos-scripts/commit/174a847e362f8a1d05210750eafb6ccb4735f59c

So existing SDKs kept working but newly created SDKs would mysteriously
fail to build some things like cmake and vim. :(


marineam:
/cc @glevand this is the cause of some of your recent troubles :-/

This variable being correct pulls in fixes like:
https://github.com/coreos/coreos-overlay/blob/master/coreos/config/env/dev-util/cmake
https://github.com/coreos/coreos-overlay/blob/master/coreos/config/env/app-editors/vim
and assorted others

glevand:
@marineam I could build the latest successfully, so this seems to have fixed those problems.

PR #1378, opened 2015-07-17
rkt: fix grace period time format

marineam:
Fixes https://github.com/coreos/coreos-overlay/issues/1377

PR #1379, opened 2015-07-17
rkt: fix grace period variable format

marineam:
systemd requires the ${var} syntax when embedded in words.

Missed in https://github.com/coreos/coreos-overlay/pull/1378


crawford:
LGTM

PR #1380, opened 2015-07-17
ignition: add dependency on systemd-networkd.service

vcaputo:
This relies on https://github.com/coreos/bootengine/pull/52 introducing systemd-networkd to the initramfs.


crawford:
LGTM

PR #1383, opened 2015-07-17
coreos-base/coreos-metadata: add initial ebuild

crawford:
Depends on https://github.com/coreos/coreos-overlay/pull/1382.

### DIFF
--- a/coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
+++ b/coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
@@ -1,0 +1,39 @@
+# Copyright (c) 2015 CoreOS, Inc.. All rights reserved.
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=5
+CROS_WORKON_PROJECT="coreos/coreos-metadata"
+CROS_WORKON_LOCALNAME="coreos-metadata"
+CROS_WORKON_REPO="git://github.com"
+inherit coreos-doc cros-workon systemd
+
+if [[ "${PV}" == 9999 ]]; then
+	KEYWORDS="~amd64 ~arm64"
+else
+	KEYWORDS="amd64 arm64"
+fi
+
+DESCRIPTION="coreos-metadata"
+HOMEPAGE="https://github.com/coreos/coreos-metadata"
+SRC_URI=""
+
+LICENSE="Apache-2.0"
+SLOT="0"
+IUSE=""
+
+DEPEND="dev-lang/rust-bin"
+
+# TODO(crawford): figure out why this hack is needed.
+src_prepare() {
+	mv "${S}"/.cargo "${WORKDIR}"
+}
+
+src_compile() {
+	CARGO_HOME="${WORKDIR}/.cargo" cargo build --release --verbose || die
+}
+
+src_install() {
+	dobin target/release/coreos-metadata
+
+	systemd_dounit "${FILESDIR}"/coreos-metadata.service
+}
--- a/coreos-base/coreos-metadata/files/coreos-metadata.service
+++ b/coreos-base/coreos-metadata/files/coreos-metadata.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=CoreOS Metadata Agent
+
+[Service]
+Type=oneshot
+EnvironmentFile=-/usr/share/oem/oem-release
+ExecStart=/usr/bin/coreos-metadata --provider=${OEM_ID} --output=/run/metadata/coreos
+
+[Install]
+Required-By=metadata.target


coreos-base/coreos-metadata/files/coreos-metadata.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=CoreOS Metadata Agent
+Requires=network-online.target
+After=network-online.target
```
marineam: This is the first time we are depending on network-online.target and I am suspicious of its utility and safety... and may not be required for all OEMs anyway. Perhaps fine for now though...

crawford: What do you mean about safety?

This service and it's parent target won't be enabled by default. If the user needs metadata, they'll have to add the dependency to their service.

marineam: - Do all OEMs need networking? If there is a metadata scheme that doesn't need networking then there is no reason to assume the host even has a working network at the moment.
- Our current patched version of networkd-wait-online will block until just one interface is up, this may not be the interface required to fetch metadata so coreos-metadata needs to have a wait-retry loop anyway. If coreos-metadata can handle waiting gracefully there is no point to networkd-wait-online.
- The systemd folks regard networkd-wait-online as a hack, preferring apps to be able to handle the dynamic nature of networking natively and intend networkd-wait-online to only be used when it is impractical to fix an app.

crawford: - No. That is a good point.
 - Fair
 - TIL

coreos-base/coreos-metadata/files/coreos-metadata.target
```diff
@@ -0,0 +1,3 @@
+[Unit]
```
crawford: Is this a bad idea? My original thinking was to create a general target that people would use to install services under and depend on. If it has coreos-metadata.service by default, it means that people will have to create a drop-in to modify it or another target altogether. Maybe we should just leave this out.

crawford:
This also depends on:
 - [ ] the coreos-metadata project being added to the repo manifest
 - [ ] the code being merged into the coreos-metadata project

coreos-base/coreos-metadata/files/coreos-metadata.target
marineam: To decouple from the coreos-metadata implementaiton you can do a couple things: 

 - rename to just metadata.target, perhaps install via the init repo instead.
 - use `systemd_enable metadata.target coreos-metadata.service` to add the .wants link

Effectively the same but makes more of a logical split which will be help if we do venture into the realm of alternate metadata sources.

crawford: I like that.

coreos-base/coreos-metadata/files/coreos-metadata.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=CoreOS Metadata Agent
+
+[Service]
+Type=oneshot
+EnvironmentFile=-/usr/share/oem/oem-release
+ExecStartPre=-/usr/bin/mkdir /run/metadata
```
marineam: Shouldn't coreos-metadata just do this?

crawford: Probably. I'll update it.

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -0,0 +1,40 @@
...
+
+src_compile() {
+	CARGO_HOME="${WORKDIR}/.cargo" cargo build --release --verbose || die
+}
+
+src_install() {
+	dobin target/release/coreos-metadata
+
+	systemd_dounit "${FILESDIR}"/coreos-metadata.service
+	systemd_dounit "${FILESDIR}"/coreos-metadata.target
```
marineam: target no longer included and needs enable step to be added

crawford: Woops, removed. I purposely didn't enable the service so that the user can optionally enable it.

crawford:
This isn't building properly. Let me track down the issue and push a fix.

crawford:
OK, the issue was that git was treating the index as a submodule so the files weren't actually being committed. Forcing git to commit the files makes this build nicely.

marineam:
LGTM

crawford:
OK, the issue was that git was treating the index as a submodule so the files weren't actually being committed. Forcing git to commit the files makes this build nicely.

PR #1385, opened 2015-07-17
coreos/config: Remove outdated vim fix

glevand:
Commit 5c4e74c68c3262ee35bd039d94b0237df7381888 (profiles: Use latest vim for
cross compile fix) updated vim to vim-7.4.712, which outdates this local fix.


marineam:
LGTM

PR #1387, opened 2015-07-18
coreos-base/oem-*: add oem-release

crawford:
Quoting some variables while I'm in here.


vcaputo:
lgtm, though I know you're going to change stuff.

marineam:
So using OEM_ID means it differs from the existing /etc/oem-release format. I suppose that is OK but since then /etc/oem-release cannot be a symlink to the new file update_engine and anything else reading /etc/oem-release needs to have extra logic to read both locations

crawford:
Oh, I didn't expect us to symlink /etc/oem-release to /usr/share/oem-release. If we wanted that to work, we would need to tell cloudinit to stop writing that file and let tmpfiles handle it instead. Of course, in order for that to work, we would need to ensure that /usr/share/oem-release exists first (which won't be true on older machines).

Why do we need to read from both locations? As long as /etc/oem-release has all of the info, everyone should be happy. It's unfortunate that there is duplicated information, but we can start migrating to /usr/share/oem-release and eventually adopt it once we ship CoreOS Slim.

marineam:
@crawford sorry, I forgot the fact that cloudinit will still run the OEM cloud config even if the user configuration is ignition so /etc/oem-release will still get generated.

LGTM.

PR #1391, opened 2015-07-21
Type3 bonded changes

dlaube:


marineam:
```sh
INTERFACES=$(ip link show | gawk -F ':' '/^[0-9]+/ { print $2 }' | tr -d ' ' | sed 's/lo//')
```
from https://github.com/coreos/coreos-overlay/blob/master/coreos-base/oem-rackspace-onmetal/files/rename-interfaces.sh reads ever so slightly better. But now that I think about it this is even better:

```sh
INTERFACES=$(ls -1 /sys/class/net | grep -v -e '^lo$')`
```

dlaube:
Good call. Since we need to set link down on the physical interfaces and outright set link deleted for bond, how about something like;

```bash
INTERFACES=$(ls -1 /sys/class/net | egrep -v "(^lo$|^bond)")
```

dlaube:
@marineam , would something like this be acceptable?

marineam:
Seems fine, not sure if it matters one way or the other if the bond is downed before deleting it. If it does matter include a comment saying so.

dlaube:
Alright thank you @marineam. I'm leaving the exclude for ^bond since there is also/bonding_masters in that directory.

coreos-nettest1 bin # ls /sys/class/net/
bond0/           bonding_masters  enp1s0f0/        enp1s0f1/        lo/
coreos-nettest1 bin # sh reset-interfaces.sh
Cannot find device "bonding_masters"

Running with the following gets around that and appears to work well;
INTERFACES=$(ls -1 /sys/class/net | egrep -v "(^lo$|^bond)")

dlaube:
@marineam Is there anything else we need or are we good to go here?

crawford:
LGTM

crawford:
shoulder LGTM from @marineam

PR #1392, opened 2015-07-22
app-emulation/rkt: version bumps

### LINKED ISSUE #1360: app-emulation/rkt: bump version and adapt to new build system

steveeJ:
Fixes #1360.


mischief:
@steveeJ `cros-workon` eclass is needed so that ebuilds will use source code from repos managed by the `repo` command in the coreos sdk, which includes rkt

mischief:
@steveeJ is this supposed to go upstream to gentoo? why is the copyright assigned to gentoo in rkt-9999.ebuild?

steveeJ:
@mischief originally that was the idea, as it's necessary to swap git-r3 for cros-workon I might as well change the header too.

mischief:
it would be cool to have rkt in gentoo proper, but you will need to use cros-workon for coreos-overlay and git-r3 for the upstream ebuild.

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -1,57 +1,106 @@
...
 
-	insinto /usr/share/rkt
-	doins "${S}/bin/stage1.aci"
+	insinto /usr/share/rkt/
+	doins "${S}/${BUILDDIR}/bin/stage1.aci"
 
-	systemd_dounit "${FILESDIR}"/${PN}-gc.service
-	systemd_dounit "${FILESDIR}"/${PN}-gc.timer
+	systemd_dounit "${S}"/dist/init/systemd/${PN}-gc.service
+	systemd_dounit "${S}"/dist/init/systemd/${PN}-gc.timer
```
marineam: the gc units from the rkt repo are much more aggressive than we've been using in CoreOS so far. I'd rather stick with what we have been using unless the rkt team thinks the aggressive timing is important.

steveeJ: Reverted back. I'll investigate the values but no need to block this PR until then.

marineam:
LGTM, anything we should know about before shipping this?

steveeJ:
@marineam maybe these two things:
* actools is installed by default with this package, but already noticed that according to #1402 ;-)
* and as a reminder: #1363 is not handled by this ebuild

PR #1396, opened 2015-07-22
sys-kernel/bootengine: add ignition as dependency

crawford:


vcaputo:
lookin' real good

marineam:
LGTM

PR #1398, opened 2015-07-22
Kernel: update to 4.0.9, squelch audit noise

marineam:
Revisiting audit may be needed in the future but for now avoid flooding user logs.


crawford:
LGTM

PR #1403, opened 2015-07-23
profiles: update man pages for better light reading

marineam:
Depends on https://github.com/coreos/portage-stable/pull/247


crawford:
LGTM

PR #1408, opened 2015-07-23
coreos-overlay: Move packages from portage-stable

glevand:
Move several packages from portage-stable to coreos-overlay so we can add local cross compile fixes.
Updates subversion from from subversion-1.8.11 to subversion-1.8.13 and adds an updated xfsprogs-3.2.2 ebuild.


marineam:
Looks like the xfs patch would be a good candidate to submit upstream xfs project, have you done that yet?

LGTM

glevand:
@marineam Submitted fix upstream and added link to it in ebuild.
http://oss.sgi.com/bugzilla/show_bug.cgi?id=1114

PR #1411, opened 2015-07-28
Enable selinux

mjg59:
Include a missing package that caused some problems, and configure it so that systemd doesn't get unhappy when it attempts to use policy that doesn't exist in our configuration.


marineam:
To make it clearer which semanage ebuild is in use (and contains CoreOS modifications) please drop the unneeded ebuild versions. Otherwise we are more likely to make mistakes next time we deal with this package.

mjg59:
Done.

marineam:
if we are going to keep the live 9999 ebuild could you make the same changes to it so it remains in sync?

marineam:
9999 is still missing the `echo "handle-unknown=allow" >> "${S}/src/semanage.conf"` change.

marineam:
LGTM

marineam:
Since the audit stuff landed could you rebase this please? :)

mjg59:
Testing the rebase right now

marineam:
Ship it! :)

PR #1416, opened 2015-07-30
sys-kernel/bootengine: bump bootengine to include passwd/group utils

vcaputo:
Also introduces dependency on sys-apps/shadow.


crawford:
LGTM

PR #1422, opened 2015-08-04
add Install sections to etcd and fleet

crawford:
These are useful for use with Ignition. There is a risk that poorly written cloud-configs will accidentally enable these services and break things, but I think it's time we ripped that band-aid off.


app-admin/fleet/files/fleet.service
```diff
@@ -11,3 +11,6 @@ After=fleet.socket
 ExecStart=/usr/bin/fleetd
 Restart=always
 RestartSec=10s
+
+[Install]
+WantedBy=default.target
```
marineam: A number of our units do this but I think we should actually be using `multi-user.target` which is what upstream always does.

crawford: Hmmmmmmmm, yeah. I agree. `default.target` is incorrect here.

marineam:
LGTM

PR #1423, opened 2015-08-05
subversion: set stable keywords

marineam:
When moving this ebuild from portage-stable KEYWORDS was left all ~arch,
breaking the SDK build when the ebuilds were removed from
portage-stable.


mischief:
lgtm

PR #1424, opened 2015-08-05
coreos-overlay Enable qemu user mode emulation

glevand:
Enable building arm64 qemu user mode emulation (qemu-aarch64) in
the SDK to allow for running target utilities like ldconfig during
the image build stage.


marineam:
LGTM

PR #1425, opened 2015-08-05
app-admin/kubelet: add kubernetes kubelet

crawford:
Fixes https://github.com/coreos/bugs/issues/418.

### DIFF
--- a/app-admin/kubelet/kubelet-1.0.1.ebuild
+++ b/app-admin/kubelet/kubelet-1.0.1.ebuild
@@ -1,0 +1,45 @@
+#
+# Copyright (c) 2015 CoreOS, Inc.. All rights reserved.
+# Distributed under the terms of the GNU General Public License v2
+# $Header:$
+#
+
+EAPI=5
+
+inherit flag-o-matic toolchain-funcs
+
+DESCRIPTION="Kubernetes Container Manager"
+HOMEPAGE="http://kubernetes.io/"
+KEYWORDS="amd64"
+SRC_URI="https://github.com/GoogleCloudPlatform/kubernetes/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+S="${WORKDIR}/kubernetes-${PV}"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+IUSE=""
+
+DEPEND="dev-lang/go"
+RDEPEND="net-misc/socat"
+
+src_prepare() {
+	epatch "${FILESDIR}/0001-Plumb-linker-flags-through-from-the-Makefile.patch"
+
+	if gcc-specs-pie; then
+		append-ldflags -nopie
+	fi
+
+	export CC=$(tc-getCC)
+	export CGO_ENABLED=${CGO_ENABLED:-1}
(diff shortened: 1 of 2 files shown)


mischief:
how big is it? :)

lgtm

crawford:
23MB

app-admin/kubelet/kubelet-1.0.2.ebuild
```diff
@@ -0,0 +1,38 @@
+#
+# Copyright (c) 2015 CoreOS, Inc.. All rights reserved.
+# Distributed under the terms of the GNU General Public License v2
+# $Header:$
+#
+
+EAPI=5
+
+EGIT_REPO_URI="git://github.com/GoogleCloudPlatform/kubernetes"
+EGIT_COMMIT="e310e619fc1ac4f3238bf5ebe9e7033bf5d47ee2" # v1.0.2
```
marineam: Alternate idea, use tarballs for fetching from github when we don't really need git itself for live ebuilds and such:

https://github.com/coreos/coreos-overlay/blob/16e4eadad1415a54bd0afce45369932395cf50c1/sys-apps/efunctions/efunctions-1.ebuild

The magic in vcs-snapshot makes sure things work.

crawford:
@aaronlevy and I verified this.

app-admin/kubelet/kubelet-1.0.2.ebuild
crawford: Good idea

crawford:
@marineam updated.

app-admin/kubelet/kubelet-1.0.1.ebuild
```diff
@@ -0,0 +1,27 @@
...
+
+LICENSE="Apache-2.0"
+SLOT="0"
+IUSE=""
+
+RDEPEND="net-misc/socat"
+
+src_unpack() {
+	mkdir ${S} || die
+	unpack ${A}
```
marineam: this unpack call isn't doing anything.

app-admin/kubelet/kubelet-1.0.1.ebuild
```diff
@@ -0,0 +1,27 @@
...
+# Distributed under the terms of the GNU General Public License v2
+# $Header:$
+#
+
+EAPI=5
+
+DESCRIPTION="Kubernetes Container Manager"
+HOMEPAGE="http://kubernetes.io/"
+KEYWORDS="amd64 arm64"
+SRC_URI="https://storage.googleapis.com/kubernetes-release/release/v${PV}/bin/linux/${ARCH}/${PN}"
```
marineam: Needs `".../${PN} -> ${P}"` so the binary that is downloaded and mirrored doesn't conflict with future versions.

marineam:
Why are you skipping the build and shipping a binary?

app-admin/kubelet/kubelet-1.0.1.ebuild
marineam: also if $S is not the unpacked source you may as well set `S="${WORKDIR}"` above.

marineam:
LGTM

PR #1426, opened 2015-08-06
coreos: add strace, dig, mkpasswd

marineam:
Our image is officially Not Small™ so stop pretending it is.

For unknown reasons mkpasswd is bundled with whois.


jedsmith:
Where do I send the scotch

vcaputo:
lgtm

PR #1430, opened 2015-08-06
sys-apps/ignition: ignition-files.service before initrd-parse-etc.ser…

vcaputo:
…vice

initrd-switch-root.target was too late, we get potentially killed during
the cleanup steps before that.  We also need to write files before
reloading etc since we may influence that in ignition-files.


crawford:
LGTM

PR #1432, opened 2015-08-06
Revert "coreos-base/coreos: add kubelet"

crawford:
This reverts commit 04663d5e18ce0bdb4dd31f716fae6381b549eba5.


vcaputo:
lgtm, but I'm a bit unclear on what the game plan is here.

marineam:
LGTM

UserTaken:
I was surprised to see this added but it's not going to as useful without the kube-proxy. Adding the whole kube stack is going to increase the size of `/usr` by 50% so I think going the `Flannel` route makes sense.

```
To reduce the CoreOS image size, flannel daemon is stored in CoreOS Enterprise Registry 
as a Docker container and not shipped in the CoreOS image. For those users wishing not 
to use flannel, it helps to keep their installation minimal.
```

PR #1433, opened 2015-08-06
coreos-base/oem-rackspace-onmetal: fix race

crawford:
Rather than doing this crazy rename during boot, just use the GRUB
config to pass net.ifnames=0 to the kernel.


crawford:
This hasn't been tested yet.

crawford:
Fixes https://github.com/coreos/bugs/issues/427.

marineam:
LGTM if it works ;-)

PR #1435, opened 2015-08-07
nss-usrfiles: use proper compiler

marineam:
Without exporting CC make was defaulting to using `cc` as the compiler
which is the host compiler, not the cross compiler.


mischief:
lgtm

marineam:
@glevand FYI, this fixes reading user accounts from /usr/share/baselayout/passwd and friends

PR #1436, opened 2015-08-07
app-admin/kubelet: update with better defaults

crawford:
KUBELET_DEFAULT_OPTS is used instead of individual environment variables
and flags in preparation for kubelet accepting environment variables in
the future. We don't want to collide with their flags, so we will just
convert our flag string to a series of environment variables when the
time comes.


marineam:
I don't follow why early docker is a concern. Does the dependency actually fix a correctness issue or are you just favoring a more explicit serialized dependency chain?

crawford:
Hmmm, I'm not sure why, but I thought that those dependencies were required. I do not believe they are.

marineam:
I suppose if it works without and we leave them out, it'll be one less thing to think about when rkt support happens.

marineam:
If you are going to put all the args together just drop the var and add them to the start command. At that point a dropin setting the env var would be just as repetitive as redefining the whole start command.

crawford:
But then the user wouldn't easily be able to append onto the command line.

marineam:
The user cannot append now, they can only replace the variable in one go

crawford:
Sorry. There is another variable (`KUBELET_OPTS`) that the user can use. I could add `Environment=KUBELET_OPTS=` to make it more obvious.

marineam:
Ah, well KUBELET_DEFAULT_OPTS is still pointless

marineam:
LGTM

PR #1437, opened 2015-08-08
Fixes to make better use of ccache

marineam:
In short, use CCACHE_BASEDIR to get better use of the cache

### DIFF
--- a/eclass/coreos-kernel.eclass
+++ b/eclass/coreos-kernel.eclass
@@ -62,7 +62,7 @@ find_defconfig() {
 # As if using cpio isn't bad enough already.
 # If lib doesn't exist or isn't a symlink then nothing is returned.
 get_bootengine_lib() {
-	cpio -itv --quiet < bootengine.cpio | \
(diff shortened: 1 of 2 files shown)


eclass/coreos-kernel.eclass
```diff
@@ -93,10 +93,10 @@ kmake() {
 	if gcc-specs-pie; then
 		kernel_cflags="-nopie -fstack-check=no"
 	fi
-	emake -C "${KERNEL_DIR}" \
+	emake \
 		ARCH="${kernel_arch}" \
 		CROSS_COMPILE="${CHOST}-" \
-		KBUILD_OUTPUT="${S}" \
+		KBUILD_OUTPUT="build" \
```
vcaputo: This looks like it should be "${S}/build", but I'm not familiar with any of this machinery.

marineam: ${S} is the current directory, and I want the path given to make to be relative.

crawford:
LGTM

PR #1440, opened 2015-08-08
coreos-base/coreos-metadata: add initial ebuild

crawford:


crawford:
The other approach I've been working on is to vendor the dependencies into the coreos-metadata repo. So far, I've only gotten it to work by cloning the index, but it should be possible to drop that restriction.

crawford:
I was able to make this much simpler by vendoring all of the dependencies. Yes, it does make the repo more complex, but the upside is that we will never have a crate version mismatch.

crawford:
Depends on https://github.com/coreos/portage-stable/pull/254.

crawford:
Depends on https://github.com/coreos/coreos-overlay/pull/1441.

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -0,0 +1,38 @@
...
+CROS_WORKON_PROJECT="coreos/coreos-metadata"
+CROS_WORKON_LOCALNAME="coreos-metadata"
+CROS_WORKON_REPO="git://github.com"
+inherit cros-workon systemd
+
+if [[ "${PV}" == 9999 ]]; then
+	KEYWORDS="~amd64 ~arm64"
+else
+	CROS_WORKON_COMMIT=""
+	KEYWORDS="amd64 arm64"
```
marineam: Have you verified that cross-compiling arm64 works correctly?

marineam: If not don't include arm64 in the keywords

coreos-base/coreos-metadata/files/coreos-metadata.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=CoreOS Metadata Agent
+
+[Service]
+Type=oneshot
+EnvironmentFile=-/usr/share/oem/oem-release
```
marineam: Why make oem-release file optional? Won't the command fail if OEM_ID isn't set?

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
crawford: Removed

coreos-base/coreos-metadata/files/coreos-metadata.service
crawford: Required

marineam:
LGTM

PR #1443, opened 2015-08-09
qemu static build and grub arm64 efi support

andrejro:


sys-boot/grub/grub-9999.ebuild
```diff
@@ -119,6 +122,19 @@ QA_PRESTRIPPED="
 "
 
 pkg_pretend() {
+	if [[ ${GRUB_PLATFORMS} =~ "arm64" ]] ; then
+		ebegin "Checking for aarch64-cros-linux-gnu-gcc"
+		if type -p aarch64-cros-linux-gnu-gcc > /dev/null ; then
```
marineam: Perhaps it would be better to add `grub_platforms_arm64? ( cross-x86_64-cros-linux-gnu/gcc )` to DEPEND rather than use a run-time check.

andrejro: Ok, this way requires the overlay is already set, but at the end of the day the effect is the same.
Fix commited.

marineam:
LGTM, thanks!

PR #1445, opened 2015-08-10
sys-kernel/coreos-firmware: Fix empty list error

glevand:
Fix the condition where doinst fails because there are no firmware files to
install.


sys-kernel/coreos-firmware/coreos-firmware-99999999.ebuild
```diff
@@ -115,6 +115,10 @@ src_prepare() {
...
 
 src_install() {
-	insinto /lib/firmware/
-	doins -r *
+	local list=$(find -type f) || die
+
+	if [[ -n ${list} ]] ; then
+		insinto /lib/firmware/
+		doins -r ${list}
+	fi
```
marineam: At least for amd64 we should emit a warning or possibly even die in this case. Otherwise it is possible to silently miss the fact that `firmware-scan` didn't work correctly which has happened in the past. If arm doesn't include any firmware should we just wrap whatever is depending on coreos-firmware with `amd64? ( sys-kernel/coreos-firmware )` to skip the ebuild entirely?

glevand: @marineam The current arm64 kernel config is pretty minimal, so needs no firmware.  In the general case there will be firmware files for arm64.  I think is too much to have conditional logic, etc.  I'll try to enable something in the arm64 kernel config that needs firmware.

glevand:
@marineam I changed my coreos-firmware patch from accepting an empty list to printing an error message and failing.  Without this the user just gets 'failed (install phase)' if there are no firmware files to install.

Included is an update to arm64_defconfig that adds a driver that needs firmware and also enables  more arm64 board support.

marineam:
LGTM, thanks!

PR #1454, opened 2015-08-12
docker: downgrade all kernel config checks to warnings

marineam:
The linux-info eclass sets variables referring to the kernel source and
build trees when a binary package is built. This is problematic when
linux-info uses the default build path /lib/modules/$V/build which
includes the built kernel version. So if the kernel has changed since
the docker package was built installing docker from the binary package
will fail, breaking build_image and more. To work around the issue just
make the config checks non-fatal. They really aren't all that useful to
us anyway.


crawford:
LGTM

PR #1462, opened 2015-08-13
kublet: fix expansion of KUBELET_OPTS

marineam:
In exec commands `${VAR}` always expands to a single word, making it
equivalent to `"${VAR}"` in shell.


bcwaldon:
LGTM - confirmed this fix locally

PR #1467, opened 2015-08-14
Support ARM images in sbsigntool

mjg59:
Import the Ubuntu patchset for sbsigntool so we pick up support for
signing ARM binaries.


marineam:
LGTM

PR #1472, opened 2015-08-17
sys-kernel/bootengine: set ignition-generator exec

crawford:


crawford:
Related https://github.com/coreos/bootengine/pull/58.

marineam:
LGTM once you update the git hash

PR #1475, opened 2015-08-20
Go 1.5

marineam:
Not stable on amd64 yet, but available for testing!


marineam:
Also submitted upstream: https://bugs.gentoo.org/show_bug.cgi?id=558216

philips:
@marineam not stable for amd64?!?! did you mean aarch64/arm64 instead?

marineam:
@philips It isn't stable for amd64 because I haven't actually tested building or running CoreOS with 1.5. It *is* stable for arm64, replacing the live ebuild which was previously marked stable for arm64.

So I meant exactly what I said. ;-)

philips:
Regressions are coming in for go 1.5. We should probably hold off for awhile. For example: https://github.com/golang/go/issues/12226

marineam:
Thanks for the heads up. Not too surprising considering how much changed in this release and how long it took. So I'll assume we need to wait for at least 1.5.1 before making 1.5 stable on amd64. Still need to merge this for arm though.

marineam:
Rebased using the upstream commits to go-9999.ebuild, no significant changes.

mischief:
lgtm

glevand:
I've been building amd64 with the live ebuild for a while without problems.  Never run-time tested it though.

PR #1476, opened 2015-08-20
sys-apps/ignition: move network depdencies to ignition-disks.service

vcaputo:
Explicitly order ignition-disks after the network units, and switch
from Requires to Wants in so ignition can still function without them.


crawford:
LGTM

PR #1478, opened 2015-08-24
Fix selinux configuration file location

mjg59:
We were installing selinux configuration files in /etc which caused problems
on upgrades. Move them into /usr and ensure that systemd sets up appropriate
temporary files. Fixes https://github.com/coreos/bugs/issues/447


marineam:
LGTM

PR #1480, opened 2015-08-25
sys-apps/ignition: ignition-files requires initrd-setup-root.service

vcaputo:
ignition-files can do things like add users and ssh keys, we use
shadowutils to do this which requires a minimally setup root filesystem.

Since ignition-disks might completely wipe the root filesystem,
initrd-setup-root must run to get us boot-strapped.


crawford:
LGTM, after you bump the ebuild.

PR #1481, opened 2015-08-25
flannel: bump to 0.5.3

eyakubovich:


mischief:
any testing done so far with new etcd lib against both 0.4 and 2.x?

eyakubovich:
yes, I tested it with both the latest 0.4 and etcd master. Also ran it on latest alpha with etcd2.

PR #1482, opened 2015-08-25
coreos-firmware: do not exclude directories and symlinks

marineam:
Commit ed4660a2d9abe4583b9b327d2801703ec9f4ca18 added an explicit error
message to catch when something goes wrong and cannot find any firmware
required by the kernel. This additionally changed the call to doins to
an explicit list of files, flattening out the directory tree and
excluding symlinks. The tree *must* be preserved exactly in order for
the kernel to find the files it needs.

Fixes https://github.com/coreos/bugs/issues/450


vcaputo:
lgtm

PR #1484, opened 2015-08-26
coreos-kernel: bump to 4.1.6

marineam:


marineam:
Added patch with possible fix for https://github.com/coreos/bugs/issues/435

crawford:
LGTM

PR #1485, opened 2015-08-26
coreos-overlay: dev-lang/go: arm64 package support

glevand:
Add a new USE flag 'arm64-extras' to the go ebuild which will build and install
the needed arm64 package files that will allow the SDK to cross build for arm64
targets.

@marineam There are a lot of ways we could do this.  Let me know if you have any preferences, ideas, etc.

### DIFF
--- a/dev-lang/go/go-9999.ebuild
+++ b/dev-lang/go/go-9999.ebuild
@@ -23,7 +23,7 @@ HOMEPAGE="http://www.golang.org"
 
 LICENSE="BSD"
 SLOT="0/${PV}"
-IUSE="cros_host"
+IUSE="cros_host +arm64-extras"
 
 DEPEND="cros_host? ( >=dev-lang/go-bootstrap-1.4.1 )"
 RDEPEND=""
@@ -40,6 +40,10 @@ STRIP_MASK="/usr/lib/go/pkg/*.a
 	/usr/lib/go/src/debug/elf/testdata/*
 	/usr/lib/go/src/debug/dwarf/testdata/*
 	/usr/lib/go/src/runtime/race/*.syso"
+
+if use cros_host && use arm64-extras; then
+	STRIP_MASK+="/usr/lib/go/pkg/tool/linux_arm64/*"
+fi
 
 if [[ ${PV} != 9999 ]]; then
 	S="${WORKDIR}"/go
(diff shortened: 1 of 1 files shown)


dev-lang/go/go-9999.ebuild
```diff
@@ -123,17 +127,30 @@ src_compile()
 	export GOHOSTARCH=$(go_arch ${CBUILD})
 	export GOHOSTOS=$(go_os ${CBUILD})
 	export CC=$(tc-getBUILD_CC)
+	export GOOS=$(go_os)
+
+	cd src
+
+	# Build the target support first.
```
marineam: Why does this need to be first?

dev-lang/go/go-9999.ebuild
```diff
@@ -169,6 +186,15 @@ src_install()
 	insinto /usr/lib/go/pkg/tool
 	doins -r "pkg/tool/$(go_tuple)"
 	fperms -R +x /usr/lib/go/pkg/tool
+
+	if use cros_host && use arm64-extras; then
+		insinto /usr/lib/go/pkg
+		doins -r pkg/linux_arm64
+
+		insinto /usr/lib/go/pkg/tool
+		doins -r pkg/tool/linux_arm64
```
marineam: Why is tool required? I would have thought only the std libs under pkg would be useful for cross compiling.

glevand: It my not be needed  I don't remember if it was (I made this patch a long time ago).  I'll check it.

dev-lang/go/go-9999.ebuild
glevand: As noted in a comment below this one, the host build needs to be done last so that the go environment is correct for native builds.  If we build for the host then build for the target the go environment will be for the last target.

marineam: ah, sorry, missed that. So that's an artefact of make.bash being a bit to simple for our case but it deletes go_bootstrap so we cannot just add the call to build std for the second arch.

dev-lang/go/go-9999.ebuild
glevand: @marineam We do not need tool.  I removed it from the updated patch.

marineam:
LGTM, I'd be curious what a more generic scheme will look like but no need to deal with that right now.

PR #1486, opened 2015-08-26
Don't install /sbin/selinuxenabled

mjg59:
We don't use this for anything, but Ansible won't run if it's present and
the selinux python bindings aren't.

Fixes https://github.com/coreos/bugs/issues/449


marineam:
Hm, https://github.com/coreos/coreos-overlay/blob/master/profiles/coreos/targets/generic/make.defaults feels a bit more applicable but the result is the same.

marineam:
LGTM

PR #1487, opened 2015-08-27
sys-apps/ignition: ensure ignition-disks runs after disk-uuid

vcaputo:
There's potential for racing with sgdisk or its generated udev events
which may also open the device if we get scheduled simultaneously.  This
is problematic since ignition-disks may also do partitioning and
formatting of the same devices.


crawford:
LGTM

PR #1494, opened 2015-08-27
coreos-kernel: re-enable ccache path rewriting for kernel builds

marineam:
The current kernel build process expects ccache to be effective but
since setting CCACHE_BASEDIR globally was reverted in 35d8b34a builds
have been uncached any time the source revision changes.

Until it is safe to set CCACHE_BASEDIR globally again we should at least
make use of it for kernel builds.


crawford:
LGTM

PR #1495, opened 2015-08-27
coreos-overlay: hard-host-depends: Remove unused dev-util/cmake

glevand:


marineam:
Please bump the revision in the symlink as well.

marineam:
LGTM

PR #1497, opened 2015-08-27
coreos-overlay: eclass/coreos-go Updates to aid build debugging

glevand:


marineam:
LGTM

glevand:
Fixed a minor copy and paste error that would have only been hit in x86 builds.

PR #1498, opened 2015-08-28
Add missing package metadata.xml

glevand:
Adds a metadata.xml file to any packages that are missing it.  If there is an
upstream gentoo metadata.xml file for the package, that is used, otherwise a
minimal metadata.xml is added.

Fixes equery errors like these:

  !!! Package XXX is missing metadata.xml


marineam:
LGTM

PR #1502, opened 2015-08-30
etcd 2.1.2 and Linux 4.1.6 for 766 branch

marineam:
Proposed updates for 766.2.0 which will go to beta this week and stable at a to-be-determined time after that. There isn't enough time to bake this in beta and promote it to stable following the regular schedule this week so a new date for stable will need to be decided.

 - etcd 2.1.2 includes a regression fix, typically we would just roll back versions to deal with regressions but reportedly downgrading from 2.1 to 2.0 is not safe. https://github.com/coreos/etcd/issues/3350
 - Linux 4.1.6 includes important NMI fixes related to http://www.openwall.com/lists/oss-security/2015/08/04/8


marineam:
The final patch is merely a build-time fix for the docker ebuild.

mischief:
lgtm

PR #1510, opened 2015-09-02
sys-kernel/bootengine: revert to pre-ignition version for 766

marineam:
First boot with ignition on EC2 may fail. We've worked around this in
the current alpha by disabling ignition on EC2. For the 766 branch we
can just disable ignition entirely, the old version in this branch has
other issues anyway.


crawford:
LGTM

marineam:
Added disk-uuid backport

crawford:
LGTM

PR #1511, opened 2015-09-02
init commit of interoute oem files

xied75:
This is the initial PR for interoute OEM build of CoreOS. It mainly combines cloudstack related files with vmtoolsd since we are based on VMware hypervisor. We also modified the way we expose network information into the environment file to make utilising our MPLS VPN network easier.

### DIFF
--- a/coreos-base/oem-interoute/files/cloud-config.yml
+++ b/coreos-base/oem-interoute/files/cloud-config.yml
@@ -1,0 +1,61 @@
+#cloud-config
+
+coreos:
+    units:
+      - name: vmtoolsd.service
+        command: start
+        content: |
+          [Unit]
+          Description=VMware Tools Agent
+          Documentation=http://open-vm-tools.sourceforge.net/
+          ConditionVirtualization=vmware
+
+          [Service]
+          ExecStartPre=/usr/bin/ln -sfT /usr/share/oem/vmware-tools /etc/vmware-tools
+          ExecStart=/usr/share/oem/bin/vmtoolsd
+          TimeoutStopSec=5
+      - name: cloudstack-ssh-key.service
+        command: restart
+        runtime: yes
+        content: |
+          [Unit]
+          Description=Sets SSH key from metadata
+          Requires=coreos-setup-environment.service
+          After=coreos-setup-environment.service
+
+          [Service]
+          Type=oneshot
+          StandardOutput=journal+console
+          EnvironmentFile=/etc/environment
+          ExecStart=/usr/share/oem/bin/cloudstack-ssh-key
+      - name: cloudstack-cloudinit.service
+        command: restart
+        runtime: yes
+        content: |
+          [Unit]
+          Description=Cloudinit from CloudStack-style metadata
+          Requires=coreos-setup-environment.service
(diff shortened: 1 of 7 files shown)


coreos-base/oem-interoute/oem-interoute-0.0.2.ebuild
```diff
@@ -0,0 +1,45 @@
...
+	dobin "${FILESDIR}/cloudstack-set-guest-password"
+	dobin "${FILESDIR}/cloudstack-ssh-key"
+	dobin "${FILESDIR}/cloudstack-coreos-cloudinit"
+	dobin "${FILESDIR}/coreos-setup-environment"
+
+	insinto "/usr/share/oem"
+	doins "${T}/cloud-config.yml"
+	doins "${FILESDIR}/oem-release"
+
+        insinto "/etc/systemd/system"
```
marineam: OEM's should not install to /etc or anywhere else outside of /usr/share/oem. This service file needs to be embedded into your cloud-config.yml.

coreos-base/oem-interoute/files/cloud-config.yml
```diff
@@ -0,0 +1,49 @@
...
+          [Service]
+          Type=oneshot
+          EnvironmentFile=/etc/environment
+          ExecStart=/usr/share/oem/bin/cloudstack-coreos-cloudinit
+    oem:
+      id: interoute
+      name: Interoute 
+      version-id: @@OEM_VERSION_ID@@
+      home-url: http://interoute.com/
+      bug-report-url: https://github.com/coreos/coreos-overlay
```
marineam: This should be `https://github.com/coreos/bugs/issues` these days. If other OEMs list this url we should fix that.

coreos-base/oem-interoute/files/cloud-config.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: vmtoolsd.service
+        command: start
+        content: |
+          [Unit]
+          Description=VMware Tools Agent
+          Documentation=http://open-vm-tools.sourceforge.net/
+          ConditionVirtualization=vmware
+
+          [Service]
+          ExecStartPre=/usr/bin/ln -sfT /usr/share/oem/vmware-tools /etc/vmware-tools
```
crawford: Why is this here?

marineam: vmware-tools needs that path and cloudinit runs too late to do this with tmpfiles, iirc it is this way in the existing vmware OEM

xied75: Our hypervisor is VMware. We need the scaling stuff.

-----Original Message-----

xied75: I thought I copied from it, no?

-----Original Message-----

coreos-base/oem-interoute/files/cloud-config.yml
marineam: This still need to be updated

marineam:
LGTM, any thoughts @crawford?

crawford:
LGTM

PR #1512, opened 2015-09-03
profiles: re-enable debug symbols for board packages

marineam:
Commit f066db21 dropped generating debug symbols because at the time we
were never using them. Well we want to use them now. :)


crawford:
L-gTM

PR #1514, opened 2015-09-04
coreos-sources: Add wl18xx firmware patch

glevand:
Adds missing firmware for the Hikey developer board wireless controller.
Upstream patch from http://patchwork.kernerl.xyz/patch/7114261/


marineam:
Please update the revision of coros-sources (and the corresponding revision that points to it in coreos-kernel)

marineam:
LGTM

PR #1515, opened 2015-09-06
sys-libs/ncurses: sync with upstream

marineam:
Upstream has shuffled ncurses around slots. Previously there was only:

 ncurses-5.9:5

Now there are:

 ncurses-5.9:0/5
 ncurses-5.9:5/5
 ncurses-6.0:0/6

So the new :0 slot is the package that installs everything while the :5
slot is a minimal library-only package meant to be installed along side
the new 6.0 version. All this isn't really meaningful or helpful to us
but we have to follow suit because now oodles of upstream packages
explicitly depend on the new :0 slot.

While migrating I've synced up with upstream as of commit
9c928e3de99c4b2f01f8dfd1b4cc97ae8cd181e5.

Note: Gentoo made this transition between 5.9-r4 and 5.9-r5, we are one
revision higher so -r6 is our first ebuild in the new scheme.


mischief:
lgtm

PR #1516, opened 2015-09-06
app-emulation/rkt: bump to 0.8.1 release

steveeJ:
The stage1 flavor 'host' is not working correctly using recent systemd versions
(224, 225). For now systemd is pinned to 222 in case of host flavor.


marineam:
LGTM, although I think you can just drop the 0.7 ebuild.

@philips any opinion on going with 0.8 as is, getting a 0.8.1, or adding the port change as a temporary patch in the ebuild?

jonboulle:
I'm shooting for 0.8.1 tomorrow, so we can bump this after that.

jonboulle:
coreos/rkt#1372

jonboulle:
@steveeJ mind bumping this to 0.8.1?
https://github.com/coreos/rkt/commit/a4b17f07700afd4e24be0153ca4fc73c35e14c31

krnowak:
@steveeJ: I can't comment in the file changes directly, so trying here.

Instead of creating a dir in line 69 and doing a copy, just use `--with-coreos-local-pxe-image-path=${DISTDIR}/${PXE_FILE} --with-coreos-local-pxe-image-systemd-version=v222` instead. Just make sure that systemd version is alright.

steveeJ:
Bumped and updated with @krnowak's suggestions, thanks!

app-emulation/rkt/files/rkt-gc.service
```diff
@@ -1,7 +0,0 @@
-[Unit]
```
jonboulle: eh, why are these going?

steveeJ: They were only shipped due to too aggressive timing but these were corrected in rkt: https://github.com/coreos/rkt/commit/00c539e60a6f3b1bfb724f0db0bd1b88359239aa.

jonboulle: ack, thx

jonboulle:
lgtm

app-emulation/rkt/files/rkt-gc.service
marineam: Your PR preserves the 0.7 ebuild which needs these. Either drop the old ebuild entirely or keep these files in place for now.

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -35,7 +35,7 @@ HOMEPAGE="https://github.com/coreos/rkt"
 
 LICENSE="Apache-2.0"
 SLOT="0"
-IUSE="doc examples +rkt_stage1_coreos rkt_stage1_host rkt_stage1_src +actool"
+IUSE="doc examples +rkt_stage1_coreos rkt_stage1_host rkt_stage1_src +actool systemd"
```
marineam: Doesn't look like this new use flag does anything

PR #1517, opened 2015-09-06
profiles: upgrade to GCC 4.9

marineam:
We need this version for improved arm64 support and it has worth-while
improvements. None of the 4.9 related bugs in upstream should impact us:
https://bugs.gentoo.org/show_bug.cgi?id=495124

Depends on https://github.com/coreos/portage-stable/pull/264


mischief:
lgtm

crawford:
LGTM

PR #1522, opened 2015-09-08
Import libselinux and fix setexeccon() bug

mjg59:
setexeccon() was broken in a way that broke systemd-nspawn. Backport
fec839cf17ba3e9cc9fc5e4382b00c61aee91c80 from upstream to fix that.


marineam:
LGTM

PR #1529, opened 2015-09-11
sys-apps/baselayout: bump ebuild

mischief:
the rkt group is missing in recent alpha builds even though it was added
to coreos/baselayout a while ago.

fixes coreos/bugs#459


crawford:
This needs a commit bump as well, right? Unless I'm mistaken, it looks like the baselayout is three commits behind.

mischief:
in fact, a number of commits behind. good eye. i will fix that.

crawford:
How the heck did this even work?

mischief:
@crawford what exactly, rkt? i haven't the slightest. on my alpha system running 801.0.0, i have no rkt group as shown in the original bug report.

mischief:
@crawford PTAL

crawford:
This looks correct, but @vcaputo's change makes me nervous. It is evidently not needed.

mischief:
@crawford i can remove inclusion of @vcaputo's patch to make rkt work for now. shall i do that?

crawford:
Meh, we aren't in a hurry. Might as well wait until tomorrow to sort it out with @vcaputo.

mischief:
okay, feel free to merge if you decide to cut a new release before i respond.

vcaputo:
@mischief @crawford the movement of creating home became unnecessary when we changed bootengine/dracut to have all the coreos users and groups available in its root and simply create the / and /home files in a straightforward fashion.

I would argue however that the change should remain because the creation of /home is part of initializing /.

PR #1533, opened 2015-09-12
net-fs/nfs-utils: move to 1.3.2

mischief:
nfs-utils 1.2.9 is affected by undefined behavior that produces crashes
under our new gcc 4.9 compiler.

fixes coreos/bugs#468


mischief:
**/!\ ACHTUNG! /!\**

the nfs systemd unit names have changed! they were moved from the gentoo portage repo into upstream nfs-utils.

should we override the upstream nfs-utils unit names so they maintain compatibility with gentoo's unit names?

mischief:
i've made an additional commit to make symlinks to keep compatibility with the old unit names from the gentoo ebuild.

net-fs/nfs-utils/nfs-utils-1.3.1-r5.ebuild
```diff
@@ -12,7 +12,7 @@ SRC_URI="mirror://sourceforge/nfs/${P}.tar.bz2"
 
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="alpha amd64 arm arm64 hppa ia64 ~mips ppc ppc64 s390 sh sparc x86"
+KEYWORDS="~alpha amd64 arm ~arm64 hppa ~ia64 ~mips ppc ppc64 ~s390 ~sh sparc x86"
```
marineam: should arm64 be stable too?

mischief: arm64 is using 1.3.2.. so, i am not sure?

mischief: i lied, geoff added nfs-utils-1.3.2 but didn't mark it stable for arm64.

marineam: Hm, well it seems a bit odd to have both 1.3.1 and 1.3.2 ebuilds in the overlay. Or if they are both there both should be in sync and have the same fixes (symlinks etc) in them. Either way the arm64 keywords can remain unstable for now.

mischief:
@marineam PTAL at the latest commit, which enabled rpcbind.service for rpcbind.target.

net-fs/nfs-utils/nfs-utils-1.3.1-r5.ebuild
```diff
@@ -109,13 +112,20 @@ src_install() {
 		doins id_resolver.conf
 	fi
 
-	systemd_dotmpfilesd "${FILESDIR}"/nfs-utils.conf
```
marineam: The tmpfiles rules have been dropped, are the dirs now created by the service units or what?

mischief:
@marineam re-added tmpfiles snippet, ptal

mischief:
@marineam moved to nfs-utils 1.3.2, kola tests passing :-) ptal

net-fs/nfs-utils/nfs-utils-1.3.2-r8.ebuild
```diff
@@ -123,10 +124,14 @@ src_install() {
 	fi
 	rm "${D}$(systemd_get_unitdir)"/nfs-config.service || die
 	sed -i -r \
-		-e "/^EnvironmentFile=/s:=.*:=${EPREFIX}/etc/conf.d/nfs:" \
```
marineam: should this change to `"/^EnvironmentFile=/d"`?

mischief: yes, that would get rid of `EnvironmentFile=-/run/sysconfig/nfs-utils`

mischief:
@marineam fixed the sed, ptal

marineam:
LGTM

PR #1537, opened 2015-09-15
Fix up SELinux support

mjg59:
Moving to running SELinux from systemd-nspawn rather than the stage1 directly showed up a few issues, namely:

1) Our policy didn't grant the full set of needed permissions
2) Our policy granted a bunch of permissions that were then unnecessary
3) Overlayfs was just broken as far as SELinux went

(3) is the most annoying. There are patches being worked on upstream to rectify this, but they don't work cleanly with 4.1 due to some VFS changes. 4.2 is out now, so I've moved us forward to that, forward-ported the secure boot changes and added the overlayfs fixes.


crawford:
Can you provide a diff between the two defconfigs? It also looks like you haven't updated the kernel sources.

crawford:
Oops, scratch that second comment. I spoke too soon.

mjg59:
There's no change in the defconfig - my initial check didn't suggest that there was anything interesting in the new options, but I don't know what our policies are there.

marineam:
If nothing changed that's fine.

marineam:
Since you are dropping the 4.1 ebuilds go ahead and rm the 4.1 defconfigs too. Otherwise LGTM

PR #1538, opened 2015-09-15
coreos-kernel: bump to 4.1.7 for the 766 branch

marineam:
15-cpuset-use-trialcs-mems_allowed-as-a-temp-variable.patch has been
dropped as it was included in the 4.1.7 release.


crawford:
LGTM

PR #1539, opened 2015-09-15
 app-emulation/docker: bump to v1.8.2

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/472.

marineam:
Ebuild should be `1.8.2` not `docker-1.8.1-r2`

crawford:
Updated.

marineam:
LGTM

vaijab:
How do you override this now? The only I can see going back to `json-file` logging is to replace entire `ExecStart` line.

The issue I have is that now kubernetes cannot read logs from docker.

crawford:
It will be fixed. https://github.com/coreos/bugs/issues/483

PR #1544, opened 2015-09-16
flannel: allow env file path to be overridden

aaronlevy:
The flannel options.env file isn't easily overridden in the default unit. This was an issue in https://github.com/coreos/coreos-kubernetes/pull/42 where the options file is not written via cloud-config at each boot, so we needed to symlink from /etc

Not sure this is the right approach, but ultimately hoping for an easier way to override this setting.

### DIFF
--- a/app-admin/flannel/files/flanneld.service
+++ b/app-admin/flannel/files/flanneld.service
@@ -13,12 +13,13 @@ Environment="TMPDIR=/var/tmp/"
 Environment="DOCKER_HOST=unix:///var/run/early-docker.sock"
 Environment="FLANNEL_VER={{flannel_ver}}"
 Environment="ETCD_SSL_DIR=/etc/ssl/etcd"
+Environment="FLANNEL_ENV_FILE=/run/flannel/options.env"
 LimitNOFILE=40000
 LimitNPROC=1048576
 ExecStartPre=/sbin/modprobe ip_tables
 ExecStartPre=/usr/bin/mkdir -p /run/flannel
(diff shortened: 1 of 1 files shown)


app-admin/flannel/files/flanneld.service
```diff
@@ -13,12 +13,13 @@ Environment="TMPDIR=/var/tmp/"
...
 Environment="FLANNEL_VER={{flannel_ver}}"
 Environment="ETCD_SSL_DIR=/etc/ssl/etcd"
+Environment="FLANNEL_ENV_FILE=/run/flannel/options.env"
 LimitNOFILE=40000
 LimitNPROC=1048576
 ExecStartPre=/sbin/modprobe ip_tables
 ExecStartPre=/usr/bin/mkdir -p /run/flannel
 ExecStartPre=/usr/bin/mkdir -p ${ETCD_SSL_DIR}
-ExecStartPre=/usr/bin/touch /run/flannel/options.env
+ExecStartPre=-/usr/bin/touch ${FLANNEL_ENV_FILE}
```
marineam: Why is this allowed to fail? In case the file is read-only?

aaronlevy: My thinking was if changed from default (/run/flannel/options.env) the path may not exist and touch would fail. Figured it would be better that docker errors with 'file doesn't exist', rather than the touch command failing. But the read-only case is valid too.

crawford:
LGTM

eyakubovich:
LGTM

PR #673, opened 2014-06-10
Open iscsi

ppickfor:
my first pull request so any advise welcome :)

open iscsi userspace tools

sample cloud config

```
#cloud-config
coreos:
  units:
    - name: systemd-sysctl.service
      command: restart
    - name: iscsid.socket
      command: start
    - name: open-iscsi-discoverydb@.service
      content: |
        [Unit]
        Description=Open-iSCSI discoverydb for sendtargets portal %I
        Documentation=man:iscsiadm(8)
        After=network-online.target iscsid.service 
        [Service]
        Type=oneshot
        ExecStart=/usr/sbin/iscsiadm -m discoverydb -t st -p %I -D
        RemainAfterExit=true
    - name: open-iscsi-discoverydb@station30.netremedies.ca.service
      command: start
    - name: open-iscsi-target@.path
      content: |
        [Unit]
        Description=Watch for Open-iSCSI target %i
        Documentation=man:iscsiadm(8)
        [Path]
        PathExistsGlob=/etc/iscsi/send_targets/*/*%i*
    - name: open-iscsi-target@.service
      content: |
        [Unit]
        Description=Open-iSCSI target %i
        Documentation=man:iscsiadm(8)
        After=remote-fs.target iscsid.service 
        [Service]
        Type=oneshot
        ExecStart=/usr/sbin/iscsiadm -m node -T %i -l
        ExecStop=/usr/sbin/iscsiadm -m node -T %i -u
        RemainAfterExit=true
    - name: open-iscsi-target@iqn.2012-10.ca.netremedies:athlon64.netremedies.ca.data.docker.path
      command: start
    - name: var-lib-docker.mount
      command: start
      content: |
        [Unit]
        Description=Mount to /var/lib/docker
        Before=docker.service
        After=network-online.target open-iscsi-target@iqn.2012-10.ca.netremedies:athlon64.netremedies.ca.data.docker.service
        Wants=open-iscsi-target@iqn.2012-10.ca.netremedies:athlon64.netremedies.ca.data.docker.service
        [Mount]
        What=/dev/disk/by-uuid/ae0ea887-9867-449c-87c4-7438e07a3e31
        Where=/var/lib/docker
        Type=btrfs
        Options=rw,noatime,space_cache,subvol=@docker,noauto,x-systemd.automount,x-systemd.device-timeout=30
    - name: docker.service
      command: start
    - name: etcd.service
      command: start
    - name: fleet.service
      command: start
write_files:
[...]

PR #1546, opened 2015-09-16
bump(sys-block/open-iscsi): sync with upstream

mischief:
WIP, please don't merge yet. feedback welcome, let's get this right the first time :-)


darrensessions:
will multipathd be included?

marineam:
Well, for starters please just include a single ebuild.

mischief:
@marineam would you prefer i prune ebuilds/files in the initial commit? i figured i would sync upstream and prune stuff in subsequent commits.

mischief:
i put this pr here so we can get feedback. the other pr #673 seems dead.

marineam:
Pruning now is fine, for history it'll be more interesting to see how we modify the ebuild so unused extras aren't that interesting.

I presume that this doesn't have any systemd units right?

mischief:
@marineam there's no systemd units upstream.

mischief:
@marineam ptal

mischief:
this PR depends on coreos/portage-stable#265

sys-block/open-iscsi/files/iscsidev.sh
```diff
@@ -0,0 +1,78 @@
...
+#ID_TYPE=disk
+#ID_PART_TABLE_TYPE=gpt
+#ID_VENDOR=IET
+#ID_SERIAL_SHORT=IET_00010001
+#SEQNUM=25775
+
+# do the removal
+if [[ $ACTION = 'remove' ]]; then
+  # nohup needed so this isn't constantly run...
+  nohup find -L /dev/disk/by-path/ -type l -lname ${DEVNAME} -exec rm "{}" + 2>/dev/null &
```
marineam: Oh god, I hate this script and everything it stands for. This should really be implemented something that *just* sets gives udev env vars, run as `IMPORT{program}` and then leave managing symlinks to udev.

sys-block/open-iscsi/open-iscsi-2.0.873.ebuild
```diff
@@ -0,0 +1,91 @@
...
+	if kernel_is -lt 2 6 16; then
+		die "Sorry, your kernel must be 2.6.16-rc5 or newer!"
+	fi
+
+	# Needs to be done, as iscsid currently only starts, when having the iSCSI
+	# support loaded as module. Kernel builtion options don't work. See this for
+	# more information:
+	# https://groups.google.com/group/open-iscsi/browse_thread/thread/cc10498655b40507/fd6a4ba0c8e91966
+	# If there's a new release, check whether this is still valid!
+	CONFIG_CHECK_MODULES="SCSI_ISCSI_ATTRS ISCSI_TCP"
```
marineam: Kernel checks should be non-fatal, using ~SCSI_ISCSI_ATTRS and similar. The kernel config isn't always properly detected during `build_image` causing things to fail erroneously.

sys-block/open-iscsi/open-iscsi-2.0.873.ebuild
```diff
@@ -0,0 +1,91 @@
...
+	sys-fs/lsscsi
+	sys-apps/util-linux"
+
+S="${WORKDIR}/${MY_PV}"
+
+pkg_setup() {
+	linux-info_pkg_setup
+
+	if kernel_is -lt 2 6 16; then
+		die "Sorry, your kernel must be 2.6.16-rc5 or newer!"
```
marineam: ewarn instead

sys-block/open-iscsi/open-iscsi-2.0.873.ebuild
marineam: oh, nm. missed the line below, these are already done as warnings..

sys-block/open-iscsi/open-iscsi-2.0.873.ebuild
```diff
@@ -0,0 +1,91 @@
...
+
+	# SSL (--with-security) is broken
+	econf $(use_with slp) \
+		--without-security
+}
+
+src_compile() {
+	use debug && append-flags -DDEBUG_TCP -DDEBUG_SCSI
+
+	KSRC="${KV_DIR}" CFLAGS="" \
```
marineam: What does the build need the kernel source for?

sys-block/open-iscsi/open-iscsi-2.0.873.ebuild
```diff
@@ -0,0 +1,91 @@
...
+
+	# udev pieces
+	insinto /usr/lib/udev/rules.d
+	doins "${FILESDIR}"/99-iscsi.rules
+	insopts -m0755
+	insinto /usr/lib/udev/scripts
+	doins "${FILESDIR}"/iscsidev.sh
+	insopts -m0644
+
+	keepdir /var/db/iscsi
```
marineam: needs a tmpfiles rule to create this on boot if missing.

sys-block/open-iscsi/open-iscsi-2.0.873.ebuild
```diff
@@ -0,0 +1,91 @@
...
+	fperms 700 /var/db/iscsi
+	fperms 600 /etc/iscsi/iscsid.conf
```
marineam: This is mostly an example file so installing to /etc is fine. Unless of course things just explode if it is missing, in which case we need a tmpfiles rule to make sure it exists.

mischief:
@marineam i think all the things you mentioned are fixed. PTAL

upstream, there are systemd units, but they would need to be slightly modified for our use.

mischief:
@marineam ping. should we go to upstream master for the systemd units, e.g. https://github.com/mikechristie/open-iscsi/tree/master/etc/systemd ?

marineam:
@mischief yes, lets use those units as a starting place. May need to check if the socket unit works at all with the current release though.

marineam:
@mischief I remain skeptical of `iscsidev.sh` but we can merge this as-is and improve with follow up PRs.

dlaube:
echoing @darrensessions here as well. Will multipathd be included? 
iSCSI is nice, but HA iSCSI is nicer :)

marineam:
@dlaube we'll see, we have yet to actually try out this stuff. But probably

PR #1549, opened 2015-09-17
RFC: docker: use overlay driver on XFS

marineam:
We missed this back when 4.1 came out, but now here lies a conundrum:
what is docker's behavior if someone has an existing XFS volume? This
isn't a normal setup for us but it is possible someone has done it.


mischief:
lgtm

crawford:
Just bit us in production.

PR #1550, opened 2015-09-17
arm64 dependancies

andrejro:


marineam:
I think the main thing that needs to change here is for cloudinit and docker the version to use should be determined by KEYWORDS in the ebuilds rather than special dependencies in the coreos meta ebuild.

marineam:
otherwise looks good

marineam:
I mucked around with the docker ebuild a bit further, results are here: https://github.com/coreos/coreos-overlay/pull/1554

andrejro:
Fixed the ebuild special dependencies.

marineam:
I've merged this via my own branch which rebased your changes on a different docker ebuild in order to keep the 1.7.1 version master in sync with the one we have been iterating on in the 766 release branch.

Now to give this all a test spin! Thanks :)

PR #1553, opened 2015-09-21
Revert "app-emulation/docker: use journald logger"

marineam:
This reverts commit 3aaf6715af392de36877710c5f0e8d02d60bb1fd.

The `docker log` command does not work with anything other than the
default log driver. Leave this alone for now.

Fixes https://github.com/coreos/bugs/issues/483


crawford:
LGTM

I spent some time investigating this issue and it looks like there is preliminary support upstream, but there are still a few unanswered questions (does the daemon or the client output the logs). We can take another stab at this once upstream figures out what they are doing.

PR #1552, opened 2015-09-21
Go 1.5.1

marineam:
Still not yet marked stable on amd64 but that is coming soon.

Meant to post this a week ago, oops.

PR #1554, opened 2015-09-21
ARM64 ebuild updates/fixes

marineam:
Covers https://github.com/coreos/coreos-overlay/pull/1550 but mucks around with the docker ebuild a bit further, bringing it in sync with the 1.7 release we have in the 766 branch.

Depends on https://github.com/coreos/coreos-overlay/pull/1552 and https://github.com/coreos/coreos-overlay/pull/1553


crawford:
LGTM

PR #1556, opened 2015-09-21
sys-apps/ignition: remove timeouts

crawford:
We can rely on the built-in Ignition timeouts instead. This will allow
individual OEMs to define specific timeouts. The default online-timeout
is one minute, so there is no change in behavior.


vcaputo:
lgtm

PR #1558, opened 2015-09-21
coreos-kernel: sync arm64 kernel args with amd64

marineam:
It is important to mount /sysroot in the initrd as read-write because we
must initialize the filesystem before booting. /sysroot/usr on the other
hand should be read-only by default.


marineam:
Hm, perhaps never mind on this. The option seems to be ignored :(

PR #1560, opened 2015-09-21
coreos-kernel: drop built-in kernel command lines

marineam:
It is a bit misleading to have some of this configured in the kernel
config and the rest in `grub.cfg` so lets simplify and stick to grub.


vcaputo:
sounds good, lgtm

PR #1561, opened 2015-09-23
Add kernel ebuilds for the hikey board.

andrejro:


sys-kernel/coreos-sources/coreos-sources-3.18.0.ebuild
```diff
@@ -0,0 +1,20 @@
...
+
+EAPI=5
+
+ETYPE=sources
+K_SECURITY_UNSUPPORTED=1
+inherit kernel-2
+detect_version
+detect_arch
+
+inherit git-2 versionator
```
marineam: What's versionator for?

Also cloning the whole kernel git tree tends to be kinda slow, it may be better to fetch a snapshot tarball similar to: https://github.com/coreos/coreos-overlay/blob/master/sys-apps/efunctions/efunctions-1.ebuild

coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -153,7 +153,7 @@ RDEPEND="${RDEPEND}
 	sys-fs/e2fsprogs
 	sys-fs/mdadm
 	sys-fs/xfsprogs
-	sys-kernel/coreos-firmware
+	amd64? ( sys-kernel/coreos-firmware )
```
marineam: The 4.x kernel config does happen to have a driver that uses firmware (although admittedly the driver was enabled just to make the ebuild work). Perhaps a `firmware` use flag would be appropriate?

eclass/coreos-kernel.eclass
```diff
@@ -7,6 +7,16 @@
 # Revision of the source ebuild, e.g. -r1. default is ""
 : ${COREOS_SOURCE_REVISION:=}
 
+# @ECLASS-VARIABLE: COREOS_TARGET_BOARD
```
marineam: Since the hikey build is targeting an older kernel anyway we can just leave this out for now, if needed later I'd favor a USE flag or even a different ebuild name than magic environment variables.

eclass/coreos-kernel.eclass
```diff
@@ -7,6 +7,16 @@
...
+# @DESCRIPTION:
+# Target board name, e.g. "hikey". default is ""
+: ${COREOS_TARGET_BOARD:=}
+
+# @ECLASS-VARIABLE: COREOS_TARGET_DTB
```
marineam: Is there a particular need for the DTB to be provided by the kernel ebuild? could it just be a independent package?

crawford:
Closing due to inactivity.

PR #1564, opened 2015-09-28
Enable modularized PPP support in CoreOS kernel.

sllawap:
Enable PPP support per coreos/bugs#313


marineam:
LGTM, could you bump the revision (add -r1 or update to -r2 ec) of the ebuilds?

vcaputo:
lgtm, Hi Phillip!

glevand:
@sllawap  Did you just hand edit the defconfigs, or regenerate them with savedefconfig?  It should be the later so there will be no difference with the one generated by the savedefconfig in coreos-kernel.eclass.

PR #1570, opened 2015-09-30
Add $DOCKER_OPTS to early-docker.service ExecStart

jerrykan:
Provide the ability to easily use extra docker options via a drop-in when early-docker starts up.


marineam:
LGTM, thanks!

PR #1571, opened 2015-09-30
docker: fix 1.7 ebuild, --daemon was in the service, not script

marineam:
Commit 897c35ff attempted to fix compatibility with 1.7 by copying an
older copy of the dockerd script but the script hadn't actually changed.
The incompatible change was in the service files. oops...

Only impacts arm64, amd64 uses 1.8


vcaputo:
lgtm

PR #1576, opened 2015-10-02
dev-db/etcd: notify systemd when etcd2 is ready

mischief:


mischief:
untested.

yichengq:
refer https://github.com/coreos/etcd/issues/3613

marineam:
Please rev bump the ebuild. Then lgtm

PR #1577, opened 2015-10-07
coreos-kernel: restore filesystem cmdline options in amd64 kernel

marineam:
Even though these are now in the bootloader config we cannot drop them
from the kernel because we do not update out bootloader. These should
not have been dropped in commit 24508b203 :(

Fixes regression reported in https://github.com/coreos/bugs/issues/447#issuecomment-145822964 due to the root filesystem being read-only during the initrd phase.


mischief:
lgtm

PR #1578, opened 2015-10-08
profiles/coreos: enable pcre use flag

mischief:
current this turns on pcre for grep, less, swig and wget.

fixes coreos/bugs#906


mischief:
also enables pcre for git.

marineam:
Before we link pcre into more apps could you double check that we are shipping a relatively recent version?

mischief:
dev-libs/libpcre-8.36 is the most recent stable version in upstream gentoo, and that's what we have.

marineam:
LGTM

PR #1579, opened 2015-10-10
Added fleet user and group and polkit rule

kayrus:
/cc @philips @marineam 
relates to https://github.com/coreos/baselayout/pull/38


marineam:
Needs to bump the revision of the fleet ebuild. once the baselayout change is merged also add the bump of the baselayout ebuild to this PR.

kayrus:
@marineam Bumped revision to r2 and added policy rule mentioned in https://github.com/coreos/baselayout/pull/38

kayrus:
Added policy rule in https://github.com/coreos/baselayout/pull/38

sys-apps/systemd/files/98-fleet-org.freedesktop.systemd1.rules
```diff
@@ -0,0 +1,6 @@
+polkit.addRule(function(action, subject) {
```
marineam: I think this makes more sense to install with fleet.

marineam: For clarity, from a distro packager's point of view the ideal package just ships upstream code and doesn't need distro specific hacks/patches. So given that fleet depends on systemd but not the other way around fleet related policy would be weird for systemd itself to ship.

kayrus: Moved into fleet ebuild directory

kayrus:
Added tmpfiles fleet.conf into fleet ebuild directory.

mischief:
@kayrus if fleet does not run as root, how can it write into /run?

kayrus:
@mischief because I've added tmpfiles.d:

```
d    /var/run/fleet 0755 fleet fleet - -
```

jonboulle:
LGTM

app-admin/fleet/files/sysusers.conf
```diff
@@ -0,0 +1,4 @@
+# create fleet group
+g fleet 253 - -
```
vcaputo: @marineam are we intentionally allocating 253 to the fleet group in coreos?

vcaputo:
Just one question about the magic number used for the fleet group, generally looks good to me, not that I'm familiar with polkit rules.

app-admin/fleet/files/sysusers.conf
marineam: 253 is available, although since the ebuild doesn't set the ownership of anything or run tmpfiles in advance I think we can set it to `-` and let systemd allocate the id on first boot. Static ids should only be needed when maintaining compatibility with the old scheme in baselayout or the pre-built filesystems have ownership set for some reason.

app-admin/fleet/files/fleet.socket
```diff
@@ -4,4 +4,6 @@ PartOf=fleet.service
 
 [Socket]
 ListenStream=/var/run/fleet.sock
-
+SocketMode=0660
+SocketUser=fleet
```
marineam: sysusers isn't creating a fleet user, should it?

app-admin/fleet/files/sysusers.conf
marineam: s/^g/u/ as well, you need both a user and a group, not just a group.

app-admin/fleet/files/fleet.socket
marineam: and I see, yes, it should

kayrus:
@marineam fixed

kayrus:
@marineam ping

app-admin/fleet/files/sysusers.conf
```diff
@@ -0,0 +1,4 @@
+# create fleet user and group
```
crawford: Can you move this into `files/sysusers.d/fleet.conf`?

app-admin/fleet/files/fleet.conf
```diff
@@ -0,0 +1 @@
+d    /var/run/fleet 0750 fleet fleet - -
```
crawford: Can you move this into `files/tmpfiles.d/fleet.conf`?

crawford:
Two more suggestions, otherwise LGTM.

marineam:
LGTM

kayrus:
@crawford @marineam done

PR #1582, opened 2015-10-12
Spidermonkey 1.8.5

marineam:
Prep for polkit 0.113 which will only use this version. The polkit .112 ebuild only uses spidermonkey 17.


crawford:
LGTM

PR #1584, opened 2015-10-14
Enable IMA in the kernel

andrewpollock:
Without a policy, this should have no impact at all. It makes it
possible to use IMA if desired.

This is for https://github.com/coreos/bugs/issues/416


crawford:
I could have sworn I closed this already... This was superseded by https://github.com/coreos/coreos-overlay/pull/2134.

PR #1587, opened 2015-10-15
Enabled epoll for libmicrohttpd and systemd-journal-remote

kayrus:
@mischief @marineam @crawford 

resolves https://github.com/coreos/bugs/issues/919

### DIFF
--- a/profiles/coreos/targets/generic/package.use
+++ b/profiles/coreos/targets/generic/package.use
@@ -25,7 +25,8 @@ sys-libs/gdbm		berkdb
 # enable journal gateway and container features, avoid pulling in gnutls
 sys-apps/systemd audit importd http nat -ssl
 
-net-libs/libmicrohttpd -ssl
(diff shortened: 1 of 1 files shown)


kayrus:
updated with ssl useflag. otherwise https doesn't work.

kayrus:
I've noticed that ssl was turned off for systemd too. We have to enable ssl for security. https://github.com/coreos/coreos-overlay/commit/ea0bcc6a80a064e386accbf4ff290737bedac2a4

marineam:
Turning off ssl was a deliberate choice to avoid shipping yet another ssl implementation requiring security updates. (gnutls)

If we start shipping gnutls I would favor migrating away from openssl where possible.

Let's leave ssl out of this PR, I want to review the state of things first.

kayrus:
Left only epoll, as for ssl - I've created new ticket: https://github.com/coreos/bugs/issues/927

marineam:
LGTM

PR #1591, opened 2015-10-20
Should resolve race condition with etcd

kayrus:
https://gist.github.com/normelton/4e64b9e63faf182803d6

### DIFF
--- a/app-admin/flannel/files/flanneld.service
+++ b/app-admin/flannel/files/flanneld.service
@@ -3,6 +3,7 @@ Description=Network fabric for containers
 Documentation=https://github.com/coreos/flannel
 Requires=early-docker.service
 After=etcd.service etcd2.service early-docker.service
+Wants=etcd.service etcd2.service
 Before=early-docker.target
 
 [Service]


eyakubovich:
`etcd.service` and `etcd2.service` have mutual `Conflicts=` so what will happen when there's a `Wants=` for both of them?

kayrus:
etcd will try to start. if it fails - flannel will ignore that. then etcd2 will try to start with the same result.

eyakubovich:
But we need the user to opt into etcd or etcd2 explicitly. @crawford ?

marineam:
We shouldn't trigger etcd/etcd2 to start unconditionally, flannel may not even be configured to talk to localhost etcd for all we know. Also, when attempting to start both whichever wins the conflict resolution will happily start up if it is unconfigured (single node default). If `systemctl enable` is being used to add both to start on boot the After is sufficent to order them. If cloud config is being used to to start both then they need to be listed in the right order.

This *might* be ok if we only had one etcd service but since there are two this isn't going to work and the existing configuration is the best we can do.

marineam:
Closing, we can revisit this if we ever manage to rip etcd 0.4 out of the image. :-/

normelton:
My problem is simply with the suggested drop-in suggested on https://coreos.com/flannel/docs/latest/flannel-config.html. It uses etcdctl to jam the coreos.com/network/config property into etcd at startup. This is presumably not necessary in production, where you're not bootstrapping the etcd cluster every time a host reboots.

I understand the reluctance to depend on etcd2 across the board, but without some sort of dependency, the suggested drop-in does not appear to be reliable.

My suggestion is to add a dependency in the 50-network-config.conf drop-in:

    - name: flanneld.service
      drop-ins:
        - name: 50-network-config.conf
          content: |
            [Unit]
            Requires=etcd2.service
            After=etcd2.service
            
            [Service]
            ExecStartPre=/usr/bin/etcdctl set /coreos.com/network/config '{ "Network": "192.168.0.0/16" }'

marineam:
@normelton ah, yeah. looks like an oversight. Filed a bug to update the doc with more functionally complete examples. https://github.com/coreos/bugs/issues/938

PR #1592, opened 2015-10-20
sys-apps/ignition: remove dependency on OEM part.

crawford:
Not all images have a seperate OEM partition. This was causing boot
failures while the service waited on the device to appear. Instead,
a generator in the initramfs will populate an arg file with the OEM_ID
and the Ignition stages will source that into the environment.


crawford:
Fixes https://github.com/coreos/bugs/issues/933.
Related to https://github.com/coreos/bootengine/pull/74.

vcaputo:
lgtm

crawford:
coreos-metadata depends on oem-release, so we cannot just delete these.

crawford:
Yes we can. coreos-metadata will read the OEM_ID parameter from the kernel cmdline.

PR #1598, opened 2015-10-22
Fixed VMware condition for VMware images which run on Virtualbox

kayrus:
/cc @crawford 

Fixed an issue which occurred with one of our customer.

### DIFF
--- a/coreos-base/oem-vmware/files/cloud-config.yml
+++ b/coreos-base/oem-vmware/files/cloud-config.yml
@@ -21,6 +21,7 @@ coreos:
       content: |
         [Unit]
         Description=Cloudinit from VMware Backdoor
+        ConditionVirtualization=vmware
 
         [Service]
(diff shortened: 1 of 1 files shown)


crawford:
Why is he running a VMware image on VirtualBox?

kayrus:
@crawford because he can ;)

crawford:
I assume coreos-cloudinit fails to run with something like "no providers were available in time". If that is the case, I would argue that it is working correctly.

kayrus:
@crawford this fix doesn't harm at all. you can see that `vmtoolsd.service` has this condition too. Do you have any objective reasons to close this PR?

crawford:
vmtoolsd.service shouldn't have this condition either and it should be removed. The VMware images should only be able to run inside of VMware and this change gives the false sense that we are going to support running these images in non-VMware environments.

PR #1604, opened 2015-10-24
sys-kernel/coreos-kernel: run configs through savedefconfig

mischief:
@marineam @mjg59 @glevand does this look right?


marineam:
LGTM

PR #1613, opened 2015-10-27
coreos-base/coreos: pull sys-block/open-iscsi into coreos

mischief:


vcaputo:
lgtm

marineam:
Why did we merge this? The PR to that landed the ebuild noted there were a number of things we needed to do before it could go into coreos such as:
 - some scheme for testing
 - a scary looking udev script that would be nice to replace
 - systemd units

Also, do we know what it takes to configure iscsi from an empty /etc and so on? There was at least one empty config file that was installed which I presume needs real values for operation, is there anything else? sometimes tools explode if their config file isn't there even if the empty default would have been fine.

marineam:
Also, the moment we added the iscsi ebuild we got asked about multipath. Have we even looked into that?

mischief:
@marineam i figured it would be OK to have the binaries in place.

as for udev, i have no clue about that myself. advice on what to do there is welcome.

i can submit a followup pr to add systemd units after removing goo from upstream's units.

for multipath, it's a similar story. upstream gentoo doesn't ship any systemd units, but the upstream source code has some less-than-ideal units. dunno what to do there, but having the binaries is not a bad start imo.

marineam:
Binaries alone aren't enough, particularly if we don't even know if they work or not. So the most important thing is getting automated testing going. I would recommend start by adding a user space target daemon to the SDK to help with that, will be safer for automated build hosts than kernel provided targets.

The udev thing, as long as the script works, can be left as-is for now. I'm just suspicious of it but presumably it is workable.

Comparatively sorting out units will be a smaller task and will be easy to add while sorting out testing.

PR #1616, opened 2015-10-28
gsutil and gcutil update

marineam:


mischief:
lgtm

marineam:
fixed a dependency in the gcutil ebuild. depends on https://github.com/coreos/portage-stable/pull/358

PR #1617, opened 2015-11-03
Automatically create /etc/sysusers.d directory

kayrus:
this directory is used by [sysusers.d](http://www.freedesktop.org/software/systemd/man/sysusers.d.html)


crawford:
LGTM

PR #1620, opened 2015-11-03
app-admin/locksmith: issue 960

pbx0:
This effectively backports the solution to issue https://github.com/coreos/bugs/issues/944 into the 766 build of coreos-overlay. See https://github.com/coreos/bugs/issues/960


vcaputo:
bump the ebuild

pbx0:
Bumped ebuild to 0.3.0.1 to not confuse with the git tag 0.3.1. PTAL

vcaputo:
the established convention is to just use an -rN suffix, so in this case locksmith-0.3.0-r1.ebuild

pbx0:
@vc 0.3.0-r1 now

vcaputo:
lgtm

marineam:
LGTM

PR #1621, opened 2015-11-04
coreos-kernel: Enabled dm-cache kernel module

kayrus:
@crawford, probably there is a reason why we don't use dm-cache


mischief:
why m for amd64 and y for arm64..?

marineam:
@mischief well, it is following the existing =y for the other dm options so that should be fine for now. At some point we will need to sync up the two because right now they really have very little in common. the original arm64 config wasn't actually based on the existing amd64 config.

marineam:
@kayrus only reason would be we didn't think to enable it. I don't know of any issues/complexity caused by it so should be fine. LGTM

PR #1623, opened 2015-11-05
profiles/coreos: downgrade to go 1.4.2

mischief:
fixes coreos/bugs#956


marineam:
Go is in the overlay, just edit the ebuild directly.

mischief:
@marineam ptal

marineam:
LGTM

PR #1630, opened 2015-11-10
coreos-init: cease enabling socket-activated sshd

vcaputo:
Maybe fixes https://github.com/coreos/bugs/issues/909
Fixes https://github.com/coreos/bugs/issues/966

### DIFF
--- a/coreos-base/coreos-init/coreos-init-9999.ebuild
+++ b/coreos-base/coreos-init/coreos-init-9999.ebuild
@@ -48,7 +48,6 @@ src_install() {
 	emake DESTDIR="${D}" install
 
 	# Enable some sockets that aren't enabled by their own ebuilds.
-	systemd_enable_service sockets.target sshd.socket
 	systemd_enable_service sockets.target docker.socket
 
 	# Enable some services that aren't enabled elsewhere.


marineam:
This is certainly a thing we should do but I'm unsure how to safely migrate. Users who have changed the port or similar by modifying sshd.socket are going to loose that change. If the user has sshd.socket enabled in /etc I'm not sure whether it or sshd.service is more likely to win the conflict resolution. The migration question is the biggest reason we've stuck with this setup so far despite it confusing so many of our users, other distros don't run sshd in this manner so it is rather unexpected.

vcaputo:
Capturing OOB discussion:

Due to the potential for customized sshd_configs and explicit sshd.socket enables / customizations, blindly switching to this simply enabled sshd.service carries some risk for upgrades.

Instead we'll implement a generator which determines if sshd.service can safely be enabled, otherwise enabling sshd.socket.

vcaputo:
@marineam so this is now simply removing the sshd.socket enable, deferring sshd enablement to  https://github.com/coreos/init/pull/188

vcaputo:
I *hope* coreos or whatever it's called now under redhat no longer socket-activates sshd.

bgilbert:
Container Linux still socket-activates sshd, and we have no plans to change that: https://github.com/coreos/bugs/issues/966#issuecomment-423697629.  Fedora CoreOS will (hopefully) avoid doing so.

PR #1631, opened 2015-11-11
coreos-kernel: update fs and security options

marineam:
- Switched overlay from built-in to a module.
 - Squashfs was missing xattr support, required for filesystem
   capabilities to work. ping should now work in PXE and ISO images.
 - We never switched to stackprotector string when we updated to GCC 4.9
 - Enable extra credential and selinux checks (DEBUG_CREDENTIALS)
 - Enable RODATA and syn cookies on arm64.

Fixes https://github.com/coreos/bugs/issues/969


crawford:
LGTM

vcaputo:
lgtm

marineam:
Oops, pushed 0ed98ba by mistake before merging this PR. Meant to put that in a new one. Oh well.

PR #1632, opened 2015-11-11
Enable SELinux for Docker

mjg59:
We ship appropriate SELinux policy for Docker, but the daemon isn't using
it. Make sure Docker is built with SELinux support and turn it on at daemon
startup time.


app-emulation/docker/files/docker.service
```diff
@@ -9,7 +9,7 @@ EnvironmentFile=-/run/flannel_docker_opts.env
 MountFlags=slave
 LimitNOFILE=1048576
 LimitNPROC=1048576
-ExecStart=/usr/lib/coreos/dockerd daemon --host=fd:// $DOCKER_OPTS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ
+ExecStart=/usr/lib/coreos/dockerd daemon --selinux-enabled --host=fd:// $DOCKER_OPTS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ
```
marineam: If the selinux use flag is off (e.g. it will need to be on arm64) will this break startup?

mjg59: No, this is parsed even if there's no SELinux support

marineam:
LGTM

PR #1639, opened 2015-11-17
app-emulation/docker: disable SELinux when using btrfs

vcaputo:
Fixes https://github.com/coreos/bugs/issues/983


crawford:
The logic is slightly odd (vs `maybe_enable_selinux`) but it's fine. LGTM.

PR #1640, opened 2015-11-17
sys-kernel/coreos-sources: fix return value of switchdev_port_fdb_dump()

vcaputo:
Fixes https://github.com/coreos/bugs/issues/980


mjg59:
LGTM

PR #1646, opened 2015-11-18
go: disable arm64-extras flag, currently broken

marineam:
For some reason the arm64 build (intended only for generating a
pre-built std library) is leaking into the amd64 build, leaving the
installed `go` binary with the arm64 defaults compiled in.

Until this is resolved disable building the arm64 std library.

### DIFF
--- a/dev-lang/go/go-9999.ebuild
+++ b/dev-lang/go/go-9999.ebuild
@@ -23,7 +23,7 @@ HOMEPAGE="http://www.golang.org"
 
 LICENSE="BSD"
 SLOT="0/${PV}"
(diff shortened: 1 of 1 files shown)


glevand:
@marineam I wonder if the problem is from the --no-clean flag in the arm64-extras build.  I put that in to speed up the build.

How can I tell if the generated `go` binary is bad?

marineam:
@glevand the `go` command in the SDK will default to the aarch64 compiler, for example:

```
$ go env | grep aarch
CC="aarch64-cros-linux-gnu-gcc"
CXX="aarch64-cros-linux-gnu-g++"
```

PR #1647, opened 2015-11-19
Fix arm64 build

andrejro:
This partially fixes the build for arm64 arch.
etcd and docker still do not build.


glevand:
LGTM. 

@andrejro  I have enough fixups in my repos to build arm64, but I'm still working to get them ready for merge.  Please look there so we aren't both working on the same things.  I'll rebase the profile fixes I have to this.

I have a fix for docker.  Maybe you can work on etcd?  It would be nice if we got 2.2.1 working.  There is an open issue on it: https://github.com/coreos/etcd/issues/3623

marineam:
LGTM

PR #1651, opened 2015-11-19
app-emulation/docker: Set libdm_no_deferred_remove

glevand:
Fix docker 1.7.1 build errors.


glevand:
@marineam Should we just add a fix to the docker 1.7.1 branch?

marineam:
managing build tags in the ebuild is just fine. LGTM I think although I haven't actually looked up what this build tag does yet. :-P

PR #1652, opened 2015-11-19
sys-kernel: move dracut build to coreos-kernel

crawford:
Rather than building the initramfs in the postinst of bootengine, this
moves it to the setup of coreos-kernel. This has the advantage of
failing a build if dracut fails and ensuring that the latest initramfs
makes it into every kernel build.


mischief:
please rebase after #1658

sys-kernel/coreos-kernel/coreos-kernel-4.2.2-r10.ebuild
```diff
@@ -0,0 +1,39 @@
...
+	sys-apps/sed
+	sys-apps/shadow
+	sys-apps/systemd
+	sys-apps/util-linux
+	sys-fs/btrfs-progs
+	sys-fs/mdadm
+	sys-kernel/dracut
+	virtual/udev
+	"
+RDEPEND="${DEPEND}"
```
marineam: This is going to pull ignition into the root filesystem of prod images. Just use DEPEND instead.

sys-kernel/coreos-kernel/coreos-kernel-4.2.2-r10.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+# We are bad, we want to get around the sandbox.  So do the creation of the
+# cpio image in pkg_setup() where we are free to mount filesystems, chroot,
+# and other fun stuff.
+pkg_setup() {
```
marineam: Check for `"${MERGE_TYPE}" == binary` and return. No need to run this when not building the source. Also important to not run this for binary packages when using DEPEND but not RDEPEND as noted above.

marineam:
Why not put this in the eclass?

crawford:
Moved to the eclass and checking against MERGE_TYPE.

sys-kernel/coreos-kernel/coreos-kernel-4.3.0.ebuild
```diff
@@ -7,3 +7,21 @@ inherit coreos-kernel
 
 DESCRIPTION="CoreOS Linux kernel"
 KEYWORDS="amd64 arm64"
+
+DEPEND="
```
marineam: These should also be in the eclass.

marineam: Also, if any of them have sub-slots you can add := to trigger the rebuild, similar to the bootengine dependency in the eclass.

marineam: nm, doesn't look like any do, but you could add that to ignition.

sys-kernel/coreos-kernel/coreos-kernel-4.3.0.ebuild
```diff
@@ -7,3 +7,21 @@ inherit coreos-kernel
...
+	sys-apps/coreutils
+	sys-apps/findutils
+	sys-apps/grep
+	sys-apps/ignition
+	sys-apps/less
+	sys-apps/sed
+	sys-apps/shadow
+	sys-apps/systemd
+	sys-apps/util-linux
+	sys-fs/btrfs-progs
```
mischief: do you need sys-fs/xfsprogs here as well...?

marineam: yeah, that should be there too. e2fs also

sys-kernel/coreos-kernel/coreos-kernel-4.3.0.ebuild
crawford: Added

sys-kernel/coreos-kernel/coreos-kernel-4.3.0.ebuild
crawford: Added

eclass/coreos-kernel.eclass
```diff
@@ -20,8 +20,26 @@ SLOT="0/${PVR}"
...
+	sys-apps/coreutils
+	sys-apps/findutils
+	sys-apps/grep
+	sys-apps/ignition:=
+	sys-apps/less
+	sys-apps/sed
+	sys-apps/shadow
+	sys-apps/systemd
+	sys-apps/util-linux
+	sys-fs/btrfs-progs
```
mischief: i suppose btrfs-progs would pull in sys-fs/e2fsprogs, but it might be better to be explicit about it

marineam: yeah, clearly the deps work as-is but explicit helps. :)

marineam:
LGTM

crawford:
Hmm, I bumped Ignition and ran `./build_packages`. It rebuilt Ignition, but not the kernel.

marineam:
Hm... maybe it thinks it is ok because it isn't a runtime (RDEPEND) then. Bah...

marineam:
but I swear it worked for bootengine which has always been DEPEND not RDEPEND

marineam:
when you did this had both the kernel and ignition been built since you added the subslot and dependency?

crawford:
No, neither had been built. Should the kernel rev be bumped to pick up this change?

marineam:
the installed kernel will need to be built with the := operator in place in order for a change in subslot to trigger it. as far as portage was concerned the deps on the installed version were still valid.

marineam:
so a rev-bump isn't strictly required since in the end on-disk the same thing gets installed but a rebuild is required before := starts working. a rev-bump will force that.

crawford:
That did it. I'll include the rev bump.

marineam:
most of the time the in-ebuild dependencies will amend the recorded dependencies of installed packages, this allows for dependency fixes to be made without requiring users pointlessly compile. apparently the behavior of the := operator doesn't apply though. could be any number of weird reasons why.

marineam:
LGTM

PR #1654, opened 2015-11-20
Bring in TPM components

mjg59:
Land the userspace TPM components we need for TPM-based logging


mischief:
can you strip out the initd/confd files? they won't be used.

app-crypt/go-tspi/go-tspi-9999.ebuild
```diff
@@ -0,0 +1,31 @@
...
+DEPEND="${RDEPEND}"
+
+src_compile() {
+	      go_build "${COREOS_GO_PACKAGE}/tpmd"
+}
+
+src_install() {
+	dobin ${GOBIN}/*
+	systemd_dounit "${FILESDIR}"/tpmd.service
+	systemd_enable_service multi-user.target tpmd.service
```
marineam: What is the behavior of tmpd if there isn't a tpm?

app-crypt/trousers/trousers-0.3.13-r1.ebuild
```diff
@@ -0,0 +1,108 @@
...
+
+src_install() {
+	keepdir /var/lib/tpm
+	default
+	use doc && dodoc doc/*
+	fowners tss:tss /etc/tcsd.conf
+	systemd_dounit "${FILESDIR}"/tcsd.service
+	systemd_enable_service multi-user.target tcsd.service
+	udev_dorules "${FILESDIR}"/61-trousers.rules
+	fowners tss:tss /var/lib/tpm
```
marineam: Is there a tmpfiles rule to do this at runtime? Otherwise this won't have the right ownership on PXE or freshly-formatted ROOT filesystems.

app-crypt/trousers/trousers-0.3.13-r1.ebuild
```diff
@@ -0,0 +1,108 @@
...
+src_configure() {
+	# econf --with-gui=$(usex gtk gtk openssl)
+	econf --with-gui=openssl
+}
+
+src_install() {
+	keepdir /var/lib/tpm
+	default
+	use doc && dodoc doc/*
+	fowners tss:tss /etc/tcsd.conf
```
marineam: Why does the daemon need to own its config?

app-crypt/trousers/files/tmpfiles.d/trousers.conf
```diff
@@ -0,0 +1 @@
+C /var/lib/tpm/system.data 0644 tss tss - /usr/share/trousers/system.data
```
marineam: Does this need to be recopied during updates or should it never be touched after it is first placed there?

app-crypt/go-tspi/go-tspi-9999.ebuild
marineam: I see trousers has a condition, not sure if failing that condition will cause this unit to no op, run, or fail.

mjg59: tpmd won't even attempt to start because its dependency on trousers isn't satisfied

app-crypt/trousers/trousers-0.3.13-r1.ebuild
mjg59: Good point, I'll add that

app-crypt/trousers/trousers-0.3.13-r1.ebuild
mjg59: Great question, and one I don't have an answer to - it exits if it doesn't

app-crypt/trousers/files/tmpfiles.d/trousers.conf
mjg59: It should be left there if it already exists

marineam:
LGTM

app-crypt/go-tspi/go-tspi-9999.ebuild
```diff
@@ -0,0 +1,31 @@
...
+EGIT_REPO_URI="git://github.com/coreos/go-tspi.git"
+
+if [[ "${PV}" == 9999 ]]; then
+        KEYWORDS="~amd64 ~arm64"
+else
+        EGIT_COMMIT="9c5928e0350d9829e4d144b461884a259a176dbc"
+        KEYWORDS="amd64 arm64"
+fi
+
+KEYWORDS="amd64 arm64"
```
marineam: we missed this stray keywords line in the review :(

PR #1659, opened 2015-11-22
app-emulation/docker: do not pass empty argument when selinux is disa…

mischief:
…bled

in https://github.com/coreos/coreos-overlay/pull/1639, the script would
set USE_SELINUX to the blank string, causing an extra blank argument to
be passed to docker daemon, resulting in it not starting when docker
runs on btrfs.


app-emulation/docker/files/dockerd
```diff
@@ -76,7 +77,7 @@ if [[ -z "${ARG_DRIVER}" && -z "${DOCKER_DRIVER}" ]]; then
 fi
 
 # Enable selinux except when known to be unsupported (btrfs).
-USE_SELINUX="--selinux-enabled"
-maybe_disable_selinux
+maybe_enable_selinux
+
+exec docker "$@"
 
-exec docker "$@" "${USE_SELINUX}"
```
crawford: It should be sufficient to just remove the quotes from around `${USE_SELINUX}`. I'm curious if the quotes around `$@` are also going to be potentially problematic.

mischief: seems like it would work. i'll change my patch to address that.

mischief:
@crawford PTAL

mischief:
fixes coreos/bugs#997

marineam:
LGTM

vcaputo:
oops

PR #1660, opened 2015-11-23
app-emulation/docker: do not pass empty argument when selinux is disa…

mischief:
…bled

in https://github.com/coreos/coreos-overlay/pull/1639, the script would
set USE_SELINUX to the blank string, causing an extra blank argument to
be passed to docker daemon, resulting in it not starting when docker
runs on btrfs.


crawford:
LGTM

PR #1663, opened 2015-11-25
coreos-kernel: populate /lib/modules/$(uname -r)/build

vcaputo:
With this changeset one can build a kernel module out-of-tree in a dev
image in the usual fashion:
```
 ~# cat Makefile
 obj-m := noop.o
 ~# make -C /lib/modules/4.3.0-coreos-r1/build M=$PWD
 make: Entering directory '/usr/lib64/modules/4.3.0-coreos-r1/build'
   CC [M]  /tmp/noop/noop.o
   Building modules, stage 2.
   MODPOST 1 modules
   CC      /tmp/noop/noop.mod.o
   LD [M]  /tmp/noop/noop.ko
 make: Leaving directory '/usr/lib64/modules/4.3.0-coreos-r1/build'
 ~#
```

/lib/modules/$(uname -r)/source is now populated with a stripped source
tree (headers, scripts, Makefiles...), prod images may build out-of-tree
modules by simply bind mounting these directories into a toolchain
container without needing kernel source.

Note that this build directory is being populated from the _actual_
$KBUILD_OUTPUT used in producing the kernel in the image, which has simply
been `make clean`ed and stripped down:

```
 /lib/modules/4.3.0-coreos-r1/build# ls -la
 drwxr-xr-x 5 root root   4096 Nov 25 11:56 .
 drwxr-xr-x 4 root root   4096 Nov 25 11:56 ..
 -rw-r--r-- 1 root root 108265 Nov 25 11:52 .config
 -rw-r--r-- 1 root root      2 Nov 25 11:55 .version
 -rw-r--r-- 1 root root    578 Nov 25 11:54 Makefile
 -rw-r--r-- 1 root root 617170 Nov 25 11:55 Module.symvers
 drwxr-xr-x 3 root root   4096 Nov 25 11:23 arch
 -rw-r--r-- 1 root root  23280 Nov 25 11:52 defconfig
 drwxr-xr-x 4 root root   4096 Nov 25 11:23 include
 drwxr-xr-x 6 root root   4096 Nov 25 11:56 scripts
 lrwxrwxrwx 1 root root      9 Nov 25 11:56 source -> ../source
 /lib/modules/4.3.0-coreos-r1/build#
```

### DIFF
--- a/eclass/coreos-kernel.eclass
+++ b/eclass/coreos-kernel.eclass
@@ -114,10 +114,10 @@ kmake() {
 	fi
 	# this can be removed once it is exported globally again
 	export CCACHE_BASEDIR="${S}"
-	emake \
+	emake "--directory=${S}/source" \
 		ARCH="${kernel_arch}" \
 		CROSS_COMPILE="${CHOST}-" \
-		KBUILD_OUTPUT="build" \
+		KBUILD_OUTPUT="../build" \
 		KCFLAGS="${kernel_cflags}" \
 		LDFLAGS="" \
 		"$@"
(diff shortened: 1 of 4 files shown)


vcaputo:
This PR uses https://github.com/coreos/linux/pull/3, so lets land that first.

vcaputo:
Also, the commit message lies, this doesn't work in the dev image (I assumed the kernel source is included there) it only works for the SDK at the moment (working on minimally populating /lib/modules/$(uname -r)/src now so it works for dev and prod all the same)

vcaputo:
@marineam @mjg59 PTAL

crawford:
Addresses https://github.com/coreos/bugs/issues/949.

eclass/coreos-kernel.eclass
```diff
@@ -213,14 +215,38 @@ coreos-kernel_src_install() {
...
-	dosym "../../../src/linux-${version}" "/usr/lib/modules/${version}/source"
 
-	# this is just here for linux-info.eclass, mask from prod images
-	dodir "/usr/lib/modules/${version}/build"
-	dosym "../../../../boot/config-${version}" \
-		"/usr/lib/modules/${version}/build/.config"
-
-	save_config build/defconfig
+	# Install a stripped source for out-of-tree module builds (Debian-derived)
+	( \
```
marineam: This would probably be more readable in a function. I can't think of a need for the subshell otherwise.

eclass/coreos-kernel.eclass
```diff
@@ -213,14 +215,38 @@ coreos-kernel_src_install() {
...
+		local kernel_arch=$(tc-arch-kernel); \
+		# XXX: For some reason tc-arch-kernel is returning x86_64 on > 2.6.24
```
marineam: Hm, that's concerning, the function is used elsewhere too. Will take a look at why.

eclass/coreos-kernel.eclass
```diff
@@ -132,8 +132,10 @@ shred_keys() {
 }
 
 coreos-kernel_src_unpack() {
+	# we more or less reproduce the layout in /lib/modules/$(uname -r)/
 	mkdir -p "${S}/build" || die
-	ln -s "${KERNEL_DIR}"/* "${S}/" || die
+	mkdir -p "${S}/source" || die
```
marineam: Since you are switching from a nested for to side-by-side can source be a symlink instead of a dir of symlinks now?

vcaputo: I kept it this way to ensure any potential ../ traversals would end up in ${S}, if we symlink the source tree a ../ traversal would end up in the parent of the link target.

eclass/coreos-kernel.eclass
vcaputo: Yes, I saw that it was used all over the place and as a result took the conservative approach, since everything else seems to be happy today.

eclass/coreos-kernel.eclass
```diff
@@ -131,9 +131,48 @@ shred_keys() {
...
+	# Install a stripped source for out-of-tree module builds (Debian-derived)
+	echo source/Makefile; \
+	find source/arch/${kernel_arch} -follow -maxdepth 1 -name 'Makefile*' -print; \
+	find source/arch/${kernel_arch} -follow \( -name 'module.lds' -o -name 'Kbuild.platforms' -o -name 'Platform' \) -print; \
+	find $(find source/arch/${kernel_arch} -follow \( -name include -o -name scripts \) -follow -type d -print) -print; \
+	find source/include source/scripts -follow -print | cpio -pd \
+		--preserve-modification-time \
+		--owner=root:root \
+		--dereference \
+		"${D}/usr/lib/modules/${version}" || die
```
vcaputo: @marineam dropped the subshell, this whole thing is in a function now, did you want the chain of finds moved to a function as well?

vcaputo:
@marineam think we can land this as-is or do you see other things you'd like changed?

marineam:
LGTM

PR #1669, opened 2015-11-30
coreos-base/coreos: remove kubelet from image

crawford:
Fixes https://github.com/coreos/bugs/issues/985.


marineam:
Although I keep debating whether this is a good idea or not (it makes it possible to upgrade from an image with a kublet to one without, and thus breaking things if you used it) it seems it'll be for the best.

LGTM

PR #1670, opened 2015-11-30
Fix permissions on tcsd config

mjg59:
Trousers is very paranoid about file permissions, so ensure that we install
them correctly.


crawford:
Can you bump the ebuild version as well? Otherwise, LGTM.

mischief:
lgtm

PR #1677, opened 2015-12-03
Enable arm64 TPM support

glevand:
Change the app-crypt/trousers keyword from ~arm64 to arm64 and enable TPM support in the arm64 kernel.


mischief:
lgtm

PR #1684, opened 2015-12-08
dev-db/etcd: Fix unstable build error

glevand:
Define a value for CROS_WORKON_COMMIT to avoid build errors when building etcd unstable.
Also, add an equals sign to GO_LDFLAGS to fix build warnings like:

  link: warning: option '-X version.GitSHA XXX' may not work in future releases

### DIFF
--- a/dev-db/etcd/etcd-9999.ebuild
+++ b/dev-db/etcd/etcd-9999.ebuild
@@ -12,6 +12,7 @@ COREOS_GO_PACKAGE="github.com/coreos/etcd"
 inherit coreos-doc coreos-go toolchain-funcs cros-workon systemd
 
 if [[ "${PV}" == 9999 ]]; then
+    CROS_WORKON_COMMIT=${CROS_WORKON_COMMIT:="HEAD"}
     KEYWORDS="~amd64 ~arm64"
 else
     CROS_WORKON_COMMIT="b4bddf685b26b4aa70e939445044bdeac822d042" # v2.2.2


mischief:
i don't think you need to set `CROS_WORKON_COMMIT`. cros-workon eclass clones whatever is checked out already if you are using the live ebuild (9999).

glevand:
@mischief The value of CROS_WORKON_COMMIT is used in the ebuild, and if undefined will cause a build error.

crawford:
Let's hold off on the equals sign in the LDFLAGS for now. This change will cause build failures in Go 1.4. Once we completely switch to Go 1.5 we can go through and fix all of these in one go.

glevand:
Pushed out a new version that only fixes the build error.

mischief:
this still breaks the functionality of being able to build any commit you wish when you use cros_workon.

surely there is a different way that we can get the commit hash of the repo..

@marineam any ideas?

glevand:
Pushed out a new version that sets CROS_WORKON_COMMIT to a default if undefined.

@mischief This should allow the user to specify a revision with something like:

  <code># CROS_WORKON_COMMIT=XXX emerge =dev-db/etcd-9999</code>

Is this how you expect to use it?

mischief:
@glevand the idea is like:

```
cros_workon start systemd
cd ../third_party/systemd
git checkout v228
emerge-amd64-usr sys-apps/systemd
```

glevand:
@mischief That usage will work with the current fix, and in fact, this bug will be hit without it, at least with arm64 builds, since CROS_WORKON_COMMIT will be undefined.

mischief:
@marineam any thoughts?

glevand:
Just to mention what the problem is, in the ebuild we have
 
    GO_LDFLAGS="-X ${COREOS_GO_PACKAGE}/version.GitSHA ${CROS_WORKON_COMMIT:0:7}

When CROS_WORKON_COMMIT is undefined, the linker gets called for arm64 builds with ````version.GitSHA -extldflags -Wl,-O1````.  ````-extldflags```` is taken as the value of ````version.GitSHA````, and ````-Wl,-O1```` is taken as an option to go, which it errors on.

crawford:
LGTM

PR #1687, opened 2015-12-09
app-admin/kubelet: force NodeReady condition to be last on existing nodes

aaronlevy:
Upstream ref: https://github.com/kubernetes/kubernetes/pull/18401


mischief:
you need to `git mv kubelet-1.1.2.ebuild kubelet-1.1.2-r1.ebuild` to increase the revision of the ebuild, since it was modified

aaronlevy:
Fixed revision

mischief:
lgtm

PR #1690, opened 2015-12-10
Glibc 2.21 and simplify locale management in the SDK

marineam:
Depends on https://github.com/coreos/scripts/pull/489


vcaputo:
@marineam Are the omitted patches now upstream and included with the new version?

marineam:
@vcaputo the omitted patches are for ebuilds that used to be in Gentoo but have since been removed. I've been syncing the files directory verbatim but pruning the ebuilds in the top level dir to just our own.

marineam:
The one patch that was applicable to us that I've dropped is `locale-default-en_US.patch` because it is no longer applicable, it was patching the `locale.gen` config file which this change now omits entirely.

vcaputo:
Ok, thanks for the clarification. lgtm!

PR #1691, opened 2015-12-14
Use defined ETCD_SSL_DIR path inside the flannel container

kayrus:
Flannel TLS configuration is kinda confusing... When you would like to configure custom keys path you have to set custom `ETCD_SSL_DIR` environment variable, but the actual path to the keys should be `/etc/ssl/etcd`.

This fix tells docker to use same destination path.


crawford:
LGTM

kayrus:
@crawford @marineam merge?

PR #1693, opened 2015-12-15
coreos-devel/sdk-depends: bring in eix

mischief:
depends on https://github.com/coreos/portage-stable/pull/371


marineam:
Note: this also pulls in app-shells/push but I'm assuming that that's pretty small.

Though eix has yet another ebuild cache right? Is that small/quick enough to maintain automatically in our scripts? If not maybe best to leave it out by default and leave it to people to emerge if they want to deal with it.

mischief:
yep, it brings in push.

eix has a db of its own, yes. need a repo sync hook to keep it up to date. not sure what to do yet about sdk vs SYSROOT databases.

marineam:
Well, sdk vs sysroot has the same repos (in fact repos.conf gets coppied verbatim). pkgdb is obviously different but ROOT= is one of the most basic of funky features tools like this should support.

mischief:
i wrote some wrappers by hand for eix, like

eix-sync-amd64-usr:
```
#!/bin/bash

export PORTAGE_CONFIGROOT="/build/amd64-usr"
export SYSROOT="${SYSROOT:-/build/amd64-usr}"
export ROOT="/build/amd64-usr"
exec sudo -E eix-sync  "$@"
```

mischief:
iirc i tried to add these scripts to setup_board to be generated, but ran into trouble. i've forgotten now what the issue was.

marineam:
As far as I can tell you want eix-update, not sync right? Also sysroot appears to be meaningless to eix. so wrapper just would want PORTAGE_CONFIGROOT and ROOT.

PR #1696, opened 2015-12-16
sys-kernel/coreos-kernel: bump to 4.3.3

crawford:


mischief:
you will need to remove 0021-net-switchdev-fix-return-code-of-fdb_dump-stub.patch from coreos-sources as it has been applied as of 4.3.3.

crawford:
Dang, I didn't push the latest commit.

crawford:
Wait......... yes I did. GitHub's caching is always screwed up.

vcaputo:
lgtm, but I'd like it if we kept a branch @ github.com/coreos/linux in sync with the patchset applied in the ebuild, so the set could be reproduced exactly via `git format-patch` of the branch.

crawford:
I can create a v4.3.3-coreos branch, though I would rather not manually keep them in sync.

crawford:
@vcaputo https://github.com/coreos/linux/pull/4

vcaputo:
lgtm

PR #1697, opened 2015-12-17
coreos-init: update to fix resizing on cciss disks

marineam:
https://github.com/coreos/init/pull/190


mischief:
should be r120, not r200..?

marineam:
counting is hard, fixed

mischief:
lgtm

PR #1698, opened 2015-12-17
preserve 'System.map' during kernel build

kdomanski:
allows the proper operation of 'depmod' when building out-of-tree modules

### DIFF
--- a/eclass/coreos-kernel.eclass
+++ b/eclass/coreos-kernel.eclass
@@ -156,6 +156,11 @@ prepare-lib-modules-release-dirs() {
 		"${D}/usr/lib/modules/${version}" || die
 
 	# Clean up the build tree and install for out-of-tree module builds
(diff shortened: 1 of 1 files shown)


eclass/coreos-kernel.eclass
```diff
@@ -156,6 +156,11 @@ prepare-lib-modules-release-dirs() {
 		"${D}/usr/lib/modules/${version}" || die
 
 	# Clean up the build tree and install for out-of-tree module builds
+	find build -follow -maxdepth 1 -name 'System.map' -print \
```
vcaputo: @kdomanski please use "build/" here like the others

vcaputo:
small nit, otherwise reasonable and lgtm, thanks!

kdomanski:
@vcaputo @crawford fixed

PR #1700, opened 2015-12-18
dev-db/etcd: remove service start timeout

mischief:
fixes coreos/bugs#1045


crawford:
Can you just cherry-pick the change instead?

mischief:
no, because the ebuild in the original change is different.

crawford:
LGTM

PR #1701, opened 2015-12-19
prevent overriding DOCKER_OPTS for disabling selinux

ieth0:
exec docker "$@" ${USE_SELINUX} will override DOCKER_OPTS if it contains --selinux-enabled=false, and puts double --selinux-enabled in args which is confusing too.

docker daemon --host=fd:// --selinux-enabled=false --selinux-enabled

### DIFF
--- a/app-emulation/docker/files/dockerd
+++ b/app-emulation/docker/files/dockerd
@@ -24,6 +24,10 @@ parse_docker_args() {
                 ARG_DRIVER="$1"
                 shift
                 ;;
+            --selinux-enabled)
+                ARG_SELINUX="$1"
+                shift
+                ;;
             *)
                 # ignore everything else
                 ;;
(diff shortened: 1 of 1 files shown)


crawford:
LGTM

/cc @vcaputo

app-emulation/docker/files/dockerd
```diff
@@ -64,6 +68,9 @@ maybe_disable_selinux() {
             # Leave enabled for everything else.
             ;;
     esac
+    if [[ -n "${ARG_SELINUX}" ]]; then
```
vcaputo: Let's drop this hunk and instead make USE_SELINUX="--selinux-enabled" and maybe_disable_selinux conditional on -z "${ARG_SELINUX}" similar to how select_docker_driver is happening.

Thanks!

vcaputo:
closing in favor of https://github.com/coreos/coreos-overlay/pull/1721

PR #1703, opened 2015-12-22
Fix up selinux policy for overlays

mjg59:
MCS is restricting us from performing relabelfrom, and docker uses another
label for the underlying files so we need to permit entrypoint from there.


mjg59:
Part of the fix for https://github.com/coreos/bugs/issues/1015

vcaputo:
This lgtm, but I need to spend more time with selinux before that should mean much.

PR #1704, opened 2015-12-27
app-emulation/rkt: bump version 0.13.0 -> 0.14.0

jonboulle:
Also fixes the bad naming in the previous couple of releases (13.0 and
11.0 instead of 0.13.0 and 0.11.0)

/cc @steveej @mischief

Potentially fixes coreos/bugs#1053


app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -19,7 +19,7 @@ if [[ "${PV}" == "9999" ]]; then
 	KEYWORDS="~amd64"
 else
 	KEYWORDS="amd64"
-	CROS_WORKON_COMMIT="830d80ce39f3b68ca39dd5ec88920f7d7431d436" # v0.13.0
+	CROS_WORKON_COMMIT="a38fa2d39388a89f4dc2fb4c6354c07f016555aa" # v0.13.0
```
mischief: update version comment

mischief:
small nit, lgtm after that

app-emulation/rkt/rkt-9999.ebuild
jonboulle: my god

jonboulle:
fixed

crawford:
We might need to mask 13.0 as well.

steveeJ:
I would suggest to enable the *fly* stage1 flavor so people who want to try that on CoreOS are all set.

mischief:
@jonboulle do you intend to enable rkt fly?

@marineam does anything else need to be done here?

PR #1705, opened 2015-12-29
glibc: Do not check /dev/pts options

marineam:
The SDK chroot has no control over /dev/pts since it is bind-mounted
from the host system. As far as I can tell there isn't any significant
issue caused by the `tty` group having a different gid on the host
system from the SDK anyway.


mischief:
lgtm

marineam:
Bumped the ebuild revision to ensure binary packages with this change are preferred over broken ones without.

mischief:
good idea. forgot to check that :)

PR #1707, opened 2015-12-31
glibc: ensure cross-rpcgen is compiled during the compile phase

marineam:
When bootstrapping glibc installs headers but is not built. However a
native tool, cross-rpcgen, must be built in order to generate rpcsvc
header files. By skipping directly to `make headers-install` the tool
winds up getting built during src_install, building as root instead of
the portage user and so on. Although uncommon this may lead to
permission issues with ccache.


mischief:
lgtm

PR #1711, opened 2016-01-07
Minor cleanup/fixes for the kernel eclass

marineam:


crawford:
LGTM

vcaputo:
So did I break the module signing with my kernel changes, and this fixes it?

PR #1714, opened 2016-01-11
app-emulation/rkt: 0.13.0 -> 0.15.0, add fly stage1

steveeJ:
metadata
* Add IUSE and description rkt_stage1_fly

ebuild
* Bump version to 0.15.0
* Add IUSE rkt_stage1_fly and enable by default
* Rewrite rkt ebuild to support multiple stage1 flavors

Deprecates #1704.
Fixes https://github.com/coreos/bugs/issues/1053.


jonboulle:
thanks! lgtm

crawford:
I think you need to add `=app-emulation/rkt-13.0` to `profiles/coreos/base/package.mask` since 13.0 is newer than 0.15.0.

steveeJ:
@crawford does the image builder save state in between builds? I'm assuming it picks the latest available version every time the it creates a new image, so that it just picks up 0.15.0 at the next run. I did run `sudo emerge -av rkt` in the SDK and it downgraded fine too.

Please let me know if I'm missing something and that we really need the mask.

crawford:
LGTM

PR #1717, opened 2016-01-12
sys-apps/shadow: split shadow.conf

crawford:
This will allow bootengine to only apply the shadow config that applies to /etc.


mischief:
lgtm

marineam:
LGTM

PR #1721, opened 2016-01-13
app-emulation/docker: don't clobber --selinux-enabled

vcaputo:
exec docker "$@" ${USE_SELINUX} will override DOCKER_OPTS if it contains
--selinux-enabled=false, and puts double --selinux-enabled in args which
is confusing too.

docker daemon --host=fd:// --selinux-enabled=false --selinux-enabled


crawford:
LGTM

PR #1706, opened 2015-12-30
[WIP - Do not merge] app-admin/kubelet: run the kubelet from a rkt container

aaronlevy:
This is a general proposal for a kubelet.service unit which will run the kubelet via a rkt-fly container & also allow management / updates of the kubelet version.

This would allow two modes of selecting kubelet version:

1. Pinning the version via drop-in:

    ```
    [Service]
    Environment=KUBELET_VERSION=v1.1.2
    ```

1. Using a "VersionGetterThingy" (as referenced in the kubelet.service file)

    If the KUBELET_VERSION env var is not set, we will run a (codename)VersionGetterThingy which uses same flags as kubelet to contact kube-apiserver and determine the version which should be run.

    The architecture of this VersionGetterThingy is still somewhat TBD, but for more info there is a [proposal here](https://docs.google.com/document/d/1ROk3gPMxEFDcyyo6rLySryzq1hVGJeJOy33tFNeiRpY) (referenced as node-agent in the linked doc)

PR #1726, opened 2016-01-15
[WIP- Do not merge] app-admin/kubelet: kubelet-service now runs kubelet via agent

pbx0:
Taking over from: https://github.com/coreos/coreos-overlay/pull/1706

This commit changes `kubelet.service` from running the kubelet to actually running an agent, [`kva`](https://github.com/coreos/coreos-kubernetes/pull/246), that is responsible for launching and updating a kubelet ACI with rkt.

The kubetlet is launched as transient unit named `kubelet-v1.2.3.service`. [The agent](https://github.com/coreos/coreos-kubernetes/pull/246) will be shipped with the OS.

### DIFF
--- a/app-admin/kubelet/files/kubelet.service
+++ b/app-admin/kubelet/files/kubelet.service
@@ -1,19 +1,49 @@
 [Unit]
-Description=Kubernetes Kubelet
-Documentation=https://github.com/GoogleCloudPlatform/kubernetes
+Description=Launches an agent that launches the kubelet and restarts when new versions are detected.
+Documentation=http://kubernetes.io
 
 [Service]
+
+# If left empty the kubelet version will be determined by contacting the kube-apiserver.
+Environment=KUBELET_PINNED_VERSION=
+
+# If left empty the App Container Image name will be:
+# coreos.com/hyperkube
+Environment=KUBELET_ACI=
+
+# CLI flags to pass to the Kubelet binary
 Environment=KUBELET_OPTS=
 
-ExecStart=/usr/bin/kubelet \
- --address=127.0.0.1 \
- --api-servers=http://127.0.0.1:8080 \
- --allow-privileged=false \
- --config=/etc/kubernetes/manifests \
- --v=2 \
- --cadvisor-port=0 \
- $KUBELET_OPTS
+# These are the default mount/volume options necessary for running the Kubelet in a rkt container.
+Environment='KUBELET_RKT_FLAGS=\
+  --volume etc-kubernetes,kind=host,source=/etc/kubernetes \
+  --volume etc-ssl-certs,kind=host,source=/usr/share/ca-certificates \
+  --volume var-lib-docker,kind=host,source=/var/lib/docker \
+  --volume var-lib-kubelet,kind=host,source=/var/lib/kubelet \
+  --volume run,kind=host,source=/run \
+  --mount volume=etc-kubernetes,target=/etc/kubernetes \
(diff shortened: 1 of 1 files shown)


app-admin/kubelet/files/kubelet.service
```diff
@@ -1,19 +1,48 @@
...
-Documentation=https://github.com/GoogleCloudPlatform/kubernetes
+Description=Kubelet Agent. Launches the kubelet and restarts when new versions are detected.
+Documentation=http://kubernetes.io
 
 [Service]
+# Hardcoded version of kubelet agent
+Environment=KUBELET_AGENT_VERSION=0.1.0
+
+# If left empty the kubelet version will be determined by contacting the kube-apiserver.
+Environment=KUBELET_PINNED_VERSION=
```
aaronlevy: Nit: KUBELET_VERSION instead?

app-admin/kubelet/files/kubelet.service
```diff
@@ -1,19 +1,48 @@
...
-ExecStart=/usr/bin/kubelet \
- --address=127.0.0.1 \
- --api-servers=http://127.0.0.1:8080 \
- --allow-privileged=false \
- --config=/etc/kubernetes/manifests \
- --v=2 \
- --cadvisor-port=0 \
- $KUBELET_OPTS
+# These are the default mount/volume options necessary for running the Kubelet in a rkt container.
+Environment='KUBELET_RKT_FLAGS=\
```
aaronlevy: We may need to mount additional locations for rkt to work as a runtime:
/usr/share/rkt
/usr/bin/rkt
/var/lib/rkt/ ?

/cc @yifan-gu

app-admin/kubelet/files/kubelet.service
```diff
@@ -1,19 +1,48 @@
...
+
+ExecStartPre=/usr/bin/mkdir --parents /etc/kubernetes/manifests
+ExecStartPre=/usr/bin/mkdir --parents /var/lib/docker
+ExecStartPre=/usr/bin/mkdir --parents /var/lib/kubelet
+ExecStartPre=/usr/bin/mkdir --parents /run/kubelet
+
+# ENV options can be overridden here
+EnvironmentFile=-/run/kubelet/options.env
+
+ExecStart=/usr/bin/rkt run --stage1-image=/usr/local/rkt/stage1-fly.aci \
```
aaronlevy: Stage1 path on the box I just launched is  /usr/share/rkt/stage1-fly.aci

app-admin/kubelet/files/kubelet.service
pbx0: I don't feel too strongly here, but, I have a slight preference for `KUBELET_PINNED_VERSION` (or anything that makes it clear your pinning) I think its a little more clear that, when set, the user is forgoing any auto-updates that are normally available. 

Normally, when someone sets a version env, they probably just assume they are just swapping an old version for one they need.

app-admin/kubelet/files/kubelet.service
pbx0: Noted, but, I forgot to take this out completely since kva should be on disk now. This path is in kva where rkt fly is actually used.

pbx0:
Closing for https://github.com/coreos/coreos-overlay/pull/1735

app-admin/kubelet/files/kubelet.service
yifan-gu: Yes, it depends on where the rkt bin is on coreos

marineam:
Closing for reals ;-)

PR #1727, opened 2016-01-16
arm64_defconfig: Set CONFIG_MODULE_SIG

glevand:
Needed.  Also fixes kernel build errors.


mischief:
are you editing this by hand or did you use scripts/kernel_menuconfig? aren't there other MODULE_SIG_* parameters needed like in amd64's config?

glevand:
@mischief Yes, generated with savedefconfig.
I pushed out an update that adds CONFIG_MODULE_SIG_SHA256, which is what amd64 has.

mischief:
@glevand bump ebuild revision and LGTM

glevand:
Bumped ebuild to coreos-kernel-4.4.0-r1.

PR #1728, opened 2016-01-16
sys-kernel/coreos-kernel: enable BINFMT_MISC

mischief:
allows one to run qemu for arm64 binaries on a coreos host such as seen in the sdk docs: https://coreos.com/os/docs/latest/sdk-modifying-coreos.html


crawford:
LGTM

PR #1730, opened 2016-01-19
net-misc/openssh: apply CVE-2016-0777 patch to 6.9p1

mischief:


crawford:
Can you break this into two commits: the first being the import, and the second the application of the patch?

mischief:
@crawford PTAL

crawford:
lgtm

PR #1735, opened 2016-01-21
app-admin/kubelet: kubelet-service now runs kubelet via rkt

pbx0:
This PR supersedes https://github.com/coreos/coreos-overlay/pull/1726. This changes the kubelet.service file from using the on-disk disk kubelet to running a kubelet aci specified by the user.

This divorces the kubelet version from the OS but does not provide an update strategy besides specifying which hyperkube aci to run the kubelet from. The on-disk kubelet can eventually be phased out independently.

$KUBELET_VERSION must be explicitly set or the service will fail on start.


marineam:
Can this be closed or are you going to update this based on https://github.com/coreos/coreos-overlay/pull/1750 ?

PR #1737, opened 2016-01-21
sys-apps/shadow: increase minimum password length

crawford:
Several cloud providers have requested that the minimum password length be
increased to eight.


mischief:
lgtm

PR #1738, opened 2016-01-21
app-emulation/rkt: bump version to 0.16.0

mischief:


steveeJ:
* bash completion: I'm not sure if this is interesting on CoreOS, but I'm stating it anway. rkt now supports bash completion, taken from the [release notes](https://github.com/coreos/rkt/releases/tag/v0.16.0): ```Install the new file dist/bash_completion/rkt.bash in /etc/bash_completion.d/.```
I suggest to make this IUSE dependent which should conform to upstream Gentoo policies. At this point I'm not sure if and how fast a rkt ebuild is maintained upstream.

* go version checks: due to known vulnerabilities the rkt build system does version checks and aborts if a known bad version is in use. this is just a heads up as I hope this shouldn't cause trouble on CoreOS
A

mischief:
i don't think bash completion is interesting for coreos. @marineam ?

mischief:
@steveeJ @marineam PTAL. i updated DEPEND so it will explicitly use the go versions which have relevant security fixes.

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -38,7 +38,7 @@ IUSE="doc examples +rkt_stage1_coreos +rkt_stage1_fly rkt_stage1_host rkt_stage1
 REQUIRED_USE="|| ( rkt_stage1_coreos rkt_stage1_fly rkt_stage1_host rkt_stage1_src )"
 
 COMMON_DEPEND="sys-apps/acl"
-DEPEND=">=dev-lang/go-1.4.1
+DEPEND="|| ( ~dev-lang/go-1.4.3 ~dev-lang/go-1.5.3 )
```
marineam: How about `|| ( ~dev-lang/go-1.4.3 >=dev-lang/go-1.5.3 )`

mischief: sounds good.

marineam:
We don't include any bash completion stuff

mischief:
ptal

marineam:
LGTM

PR #1742, opened 2016-01-22
WIP: per-architecture go cross-compilers

marineam:
Positing this as-is since I need to set aside this for now but the basic
idea is to install in a similar style to the rest of the toolchains,
adding x86_64-cros-linux-gnu-go and aarch64-cros-linux-gnu-go while the
plain `go` command is a symlink to the native x86_64-pc-linux-gnu-go.

This will allow us to install different versions of Go on a
per-architecture basis, so amd64 issues do not block arm64 and so on.
The per-architecture packages are named `cross-${CTARGET}/go` and are
configured by crossdev just like the gcc and the rest.

Steps required to finish this change:
- Update crossdev to a version that accepts the `--ex-pkg` argument.
- Add go to `src/scripts/build_library/toolchain_util.sh`.
- Change `coreos-go.eclass` and ebuilds to use `${CHOST}-go`.

The change to toolchain_util.sh may look something like this: (untested)

``` diff
diff --git a/build_library/toolchain_util.sh b/build_library/toolchain_util.sh
index cae6508..c8a5277 100644
--- a/build_library/toolchain_util.sh
+++ b/build_library/toolchain_util.sh
@@ -280,7 +280,7 @@ install_cross_toolchain() {
     local cross_chost="$1"; shift
     local cross_pkgs=( $(get_cross_pkgs $cross_chost) )
     # gdb doesn't respect the --stable flag, set it stable explicitly
-    local cross_flags=( --gdb '[stable]' --ex-gdb --stable --target "${cross_chost}" )
+    local cross_flags=( --gdb '[stable]' --ex-gdb --ex-pkg go --stable --target "${cross_chost}" )
     local cross_cfg="/usr/${cross_chost}/etc/portage/${cross_chost}-crossdev"
     local cross_cfg_data=$(_crossdev_info "${cross_flags[@]}")
     local emerge_flags=( "$@" --binpkg-respect-use=y --update --newuse )
@@ -316,14 +316,16 @@ install_cross_toolchain() {

     # Check if any packages need to be built from source. If so do a full
     # bootstrap via crossdev, otherwise just install the binaries (if enabled).
-    # It is ok to build gdb from source outside of crossdev.
+    # It is ok to build gdb and go from source outside of crossdev.
     if emerge "${emerge_flags[@]}" \
         --pretend "${cross_pkgs[@]}" | grep -q '^\[ebuild'
     then
         $sudo crossdev "${cross_flags[@]}" --stage4
     else
         $sudo emerge "${emerge_flags[@]}" \
-            "cross-${cross_chost}/gdb" "${cross_pkgs[@]}"
+            "cross-${cross_chost}/gdb" \
+            "cross-${cross_chost}/go" \
+            "${cross_pkgs[@]}"
     fi

     # Setup environment and wrappers for our shiny new toolchain
```

### DIFF
--- a/dev-lang/go/go-9999.ebuild
+++ b/dev-lang/go/go-9999.ebuild
@@ -6,6 +6,11 @@ EAPI=5
 
 export CBUILD=${CBUILD:-${CHOST}}
 export CTARGET=${CTARGET:-${CHOST}}
+if [[ ${CTARGET} == ${CHOST} ]] ; then
+	if [[ ${CATEGORY} == cross-* ]] ; then
+		export CTARGET=${CATEGORY#cross-}
+	fi
+fi
 
 inherit eutils toolchain-funcs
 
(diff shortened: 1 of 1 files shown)


bgilbert:
We're dropping arm64-usr from Container Linux and don't plan to add any other architectures.  Closing.

PR #1744, opened 2016-01-25
sys-kernel/coreos-kernel: enable ZRAM

mischief:
cc @glevand


marineam:
LGTM

glevand:
Looks OK, I'll report if I find any trouble.

PR #1745, opened 2016-01-25
portage 2.2.26

mischief:
this brings in portage 2.2.26.

depends on coreos/portage-stable#374.

a newer portage is required for EAPI 6, which recent ebuilds from upstream gentoo are beginning to use.


mischief:
cc @marineam

marineam:
Looks like portage-2.2.8 was the only version marked stable on arm64, so by dropping that building arm64 probably breaks do to the missing dependency. The only place where 2.2.8 would actually get used is in arm64 dev images which probably have other issues too but at the very least the dependency chain needs to work. Either leave the old ebuilds as is for now or mark arm64 stable please.

mischief:
@marineam ptal

marineam:
LGTM

marineam:
I'll review the portage change log, need to double check that there isn't anything that is likely to surprise us in there. But as long as nothing breaks in your tests we are probably ok.

mischief:
ok, i've run checked out my branch and run:

```
sudo emerge -1v sys-apps/portage
./update_chroot
sudo emerge -1v dev-util/catalyst
sudo ./build_toolchains
```

besides this strange message, it all succeeded.

```
 * Messages for package sys-libs/glibc-2.21-r2 merged to /build/amd64-usr/:

 * 
 * Installation of a symlink is blocked by a directory:
 *   '/build/amd64-usr/lib'
 * This symlink will be merged with a different name:
 *   '/build/amd64-usr/lib.backup.0000'
 * 
```

this symlink is inside the resulting tarball:

```
lrwxrwxrwx root/root            0 2016-01-26 03:59 ./build/amd64-usr/lib.backup.0000 -> lib64
drwxr-xr-x root/root            0 2016-01-26 04:04 ./build/amd64-usr/lib/
drwxr-xr-x root/root            0 2016-01-26 03:56 ./build/amd64-usr/lib/gentoo/
-rw-r--r-- root/root         8933 2016-01-26 03:55 ./build/amd64-usr/lib/gentoo/functions.sh
-rwxr-xr-x root/root        14288 2016-01-26 04:04 ./build/amd64-usr/lib/cpp
```

not sure if this is the result of the new portage or normal.

mischief:
this error was present in prior builds, so nothing to do with new portage.

PR #1747, opened 2016-01-27
baselayout: tolerate /proc and friends being mounted read-only

marineam:
This is required for compatibility with cork's experimental chroot implementation which restricts write access to global filesystems to provide some level of protection despite not providing proper container-like isolation.


mischief:
lgtm

PR #1748, opened 2016-01-28
sys-apps/baselayout: include change to disable LLMNR

mischief:


marineam:
Lets call it baselayout-3.0.16 since it is a change in the files rather than an ebuild tweak. but either way LGTM

mischief:
sounds good.

PR #1750, opened 2016-01-29
app-admin/kubelet-wrapper: add script to execute via rkt-fly

aaronlevy:


marineam:
Looks reasonable to me, though to avoid confusion with the existing kublet package that actually provides a kublet perhaps this ebuild should be app-admin/kubelet-rkt or similar?

One other note, if you want to use `--` to split rkt vs kublet args (may avoid quoting issues which sometimes happen when arguments are put into string variables) the following may be of use:

```sh
rktopts=()
kubopts=( "$@" )
for ((i=1; i<=$#; i++)); do
    if [[ "${!i}" == "--" ]]; then
        rktopts=( "${@:1:$((i-1))}" )
        kubopts=( "${@:$((i+1))}" )
        break
    fi
done
```

Might be a more concise way but I can't think of it at the moment.

robszumski:
random question: when you do `systemctl status` on this unit, will it show:

```
systemd
   bash
      rkt
```
or
```
systemd
   rkt
```

joshix:
woot!

@robszumski The former, with fly. It will look just like a normally-forked process in a pstree. Except, you won't see `rkt`, just `kubelet`.

```
PID 1 host systemd
|__ bash
    |__ kubelet
```

aaronlevy:
The bash script is just exec-ing rkt-fly (which execs kubelet), so I would expect to just see:

```
systemd
|__ kubelet
```

crawford:
Why are we doing this? Why not encapsulate all of this in the service unit?

aaronlevy:
One of the discussed reasons has to do with versioning. Because the end user can select the kubelet version, it could get difficult to support the various incantations we need in the service file being shipped.

E.g. We ship a service file that works fine with v1.1.x, but user decides to run v1.2.x and it happens to require new volumes passed to the hyperkube image in rkt fly. So user would have to overwrite service file wholesale. Or, we update the service file to include that new volume, but now that breaks someone on v1.0.x.

The bash wrapper will run into similar issues, but at least we have more flexibility for version-specific logic.

pbx0:
The general consensus seems to be that we move forward with shipping this script as kubelet-rkt. We won't ship the service file since it is more ergonomic to just have one service file rather then use drop-ins to modify it.

The interface to this script we would commit to is:
 - rkt flags are passed via $RKT_OPTS
 - kubelet flags are passed via command line args
 - the image location can be overridden via $KUBELET_ACI
 - aci version tag is set and required via $KUBELET_VERSION

$RKT_OPTS could be folded into the command line as mike suggested, but, since there shouldn't be any issues, it seems ever so slightly more convenient to keep them separate since $RKT_OPTS probably won't be modified in the common case.

The idea is that eventually, $KUBELET_VERSION , will just become the initial kubelet version, and the kubelet will relaunch itself. So, the script interface wouldn't need to change and service files continue working.

Any thoughts/objections on moving forward with this plan? I'll update the PR after some further testing.

pbx0:
@joshix Do you think its worth adding some more doc comments to the top of the file? Should we add a pointer to our own docs or maybe even an example service file?  This way someone looking at it has an idea of how its commonly used.

app-admin/kubelet/files/kubelet
```diff
@@ -0,0 +1,31 @@
...
+
+mkdir --parents /etc/kubernetes
+mkdir --parents /var/lib/docker
+mkdir --parents /var/lib/kubelet
+mkdir --parents /run/kubelet
+
+exec /usr/bin/rkt run \
+  --volume etc-kubernetes,kind=host,source=/etc/kubernetes \
+  --volume etc-ssl-certs,kind=host,source=/usr/share/ca-certificates \
+  --volume var-lib-docker,kind=host,source=/var/lib/docker \
```
philips: cc @yifan-gu the rkt API will keep us from needing to do this for /var/lib/rkt, right?

app-admin/kubelet/kubelet-9999.ebuild
```diff
@@ -0,0 +1,20 @@
...
+EAPI=5
+
+DESCRIPTION="Kubernetes Container Manager"
+HOMEPAGE="http://kubernetes.io/"
+KEYWORDS="amd64"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+IUSE=""
+
```
mischief: should have DEPEND on app-emulation/rkt ?

app-admin/kubelet/files/kubelet
```diff
@@ -0,0 +1,31 @@
...
+# Wrapper for launching kubelet via rkt-fly stage1
+
+set -e
+
+if [ -z "${KUBELET_VERSION}" ]; then
+    echo "ERROR: must set KUBELET_VERSION"
+    exit 1
+fi
+
+KUBELET_ACI="${KUBELET_ACI:-quay.io/coreos/hyperkube}"
```
philips: This is technically HYPERCUBE_ACI, right? Can we use a kubelet ACI instead?

app-admin/kubelet/files/kubelet
```diff
@@ -0,0 +1,31 @@
...
+  --volume var-lib-kubelet,kind=host,source=/var/lib/kubelet \
+  --volume run,kind=host,source=/run \
+  --mount volume=etc-kubernetes,target=/etc/kubernetes \
+  --mount volume=etc-ssl-certs,target=/etc/ssl/certs \
+  --mount volume=var-lib-docker,target=/var/lib/docker \
+  --mount volume=var-lib-kubelet,target=/var/lib/kubelet \
+  --mount volume=run,target=/run \
+  $RKT_OPTS \
+  --stage1-image=/usr/share/rkt/stage1-fly.aci \
+  ${KUBELET_ACI}:${KUBELET_VERSION} --exec=/hyperkube -- kubelet "$@"
```
philips: During discussion we decided to use a kubelet ACI instead. We will need to build this ACI.

app-admin/kubelet/kubelet-9999.ebuild
mischief: RDEPEND? cc @marineam

marineam: Yes, RDEPEND

app-admin/kubelet/files/kubelet
marineam: Just to be clear, there are two reasons we need to use a kublet aci instead:

 - kublet is a smaller compatibility requirement, it is less clear how hyperkube will evolve over time.
 - the calling convention of the aci must be preserved over time, doing --exec=/hyperkube commits us to always doing hyperkube, a kublet-only aci cannot be substituted. And hyperkube doesn't support aliasing via symlinks so it isn't possible to have --exec=/kublet where kublet is either a real kublet or a symlink to hyperkube. If hyperkube eventually gains the ability to respect argv[0] then it is possible for a user to specify a hyperkube aci as the kublet aci if that is useful for caching reasons or whatnot.

joshix:
@pbx0 Yes, I think it's a good idea to document the concept this wrapper script embodies. It should probably be just a URL in the head of the script itself, pointing to a document we will eventually create to cover this clever thing we're doing with rkt.

app-admin/kubelet/files/kubelet
```diff
@@ -0,0 +1,31 @@
...
+mkdir --parents /var/lib/docker
+mkdir --parents /var/lib/kubelet
+mkdir --parents /run/kubelet
+
+exec /usr/bin/rkt run \
+  --volume etc-kubernetes,kind=host,source=/etc/kubernetes \
+  --volume etc-ssl-certs,kind=host,source=/usr/share/ca-certificates \
+  --volume var-lib-docker,kind=host,source=/var/lib/docker \
+  --volume var-lib-kubelet,kind=host,source=/var/lib/kubelet \
+  --volume run,kind=host,source=/run \
```
guilhem: `kubelet` default docker to `/var/run`, if container don't have symlink from `/var/run` -> `/run` (like `docker://gcr.io/google_containers/hyperkube:v1.1.7`) it will fail

app-admin/kubelet/files/kubelet
aaronlevy: We will still need a hyperkube image because it is used for all other k8s components (apiserver, proxy, controller-manager, scheduler). If we go the kubelet-only route we either need to rely on the upstream hyperkube for all other components, or build images individually for each component.

Should we parameterize the exec option to mitigate some of the concerns? e.g:

```
KUBELET_EXEC="${KUBELET_EXEC:-/hyperkube -- kubelet}"
${KUBELET_ACI}:${KUBELET_VERSION} --exec=${KUBELET_EXEC} "$@"
```

app-admin/kubelet/files/kubelet
```diff
@@ -0,0 +1,31 @@
...
+  --mount volume=etc-kubernetes,target=/etc/kubernetes \
+  --mount volume=etc-ssl-certs,target=/etc/ssl/certs \
+  --mount volume=var-lib-docker,target=/var/lib/docker \
+  --mount volume=var-lib-kubelet,target=/var/lib/kubelet \
+  --mount volume=run,target=/run \
+  $RKT_OPTS \
+  --stage1-image=/usr/share/rkt/stage1-fly.aci \
```
steveeJ: this is called `--stage1-path` now

kayrus:
Have got this error message: https://gist.github.com/kayrus/b88275b4a112c4897a78
Using this script: https://gist.github.com/kayrus/d6695734f3df2264a121

Two issues:

* looks like `sudo ./kubelet.sh --docker='unix:///run/bla/bla/should/not/work.sock'` doesn't really pass the option into kubelet
* there are some issue with slices...

app-admin/kubelet/files/kubelet
steveeJ: Eventually, the kubelet will spawn rkt pods, and I think we something for rkt to work properly in rkt fly. Here are some thoughts.
We need `/etc/rkt` for various rkt and CNI configurations. We have possible conflicts with CNI if the kubelet and the host both execute rkt, so we need at least `/var/lib/cni` for the `host-local` allocator. 
Then there might be issues with GCing. If `/var/lib/rkt` will not be mounted the pods started by the kubelet are not going to be GCd by the host's GC timer. 
Note that it won't be mounted recursively and will not have overlayfs mounts etc. from the host's pods but that, but I'm not 100% sure about the consequences this will have if rkt GC would be called by the kubelet and only find the empty pod directories. @jonboulle @vcaputo maybe one of you can foresee this?
Where will the rkt API service be running? This is going to be tricky if the kubelet and the host both run pods and don't share resources examined by the API service. @yifan-gu

steveeJ: This seems a more general issue for _running rkt in rkt_ so I created an issue for further discussion: https://github.com/coreos/rkt/issues/2167

app-admin/kubelet/files/kubelet
marineam: I don't know what conversations happened in meat space but I want the container to just handle this correctly. It is trivial to fix hyperkube to check argv[0] and use symlinks. That allows things to just work without needing another config nob.

aaronlevy: The state that it was left in was to build both a kubelet container and a hyperkube container, and have the bash script reference the kubelet container (while all other master components stay the same an use hyperkube).

To be honest I'm not convinced of this approach, but others felt more strongly about it (more specific footprint is mostly what I gathered). Upstream builds both a kubelet and hyperkube image FWIW.

Also, it does look like hyperkube checks argv[0] (untested):
https://github.com/kubernetes/kubernetes/blob/v1.1.2/cmd/hyperkube/main.go#L40
https://github.com/kubernetes/kubernetes/blob/v1.1.2/cmd/hyperkube/hyperkube.go#L119

If that is the main concern, I would still prefer to just be building a single image...

marineam: Ok, just need to make sure there is a kublet -> hyperkube symlink in the container then.

It is fine if we favor hyperkube for whatever reason, I just don't want what we ship in coreos to be needlessly tied to it when a kublet container is also sufficient. And if the hyperkube container is an exact drop-in replacement for kublet then everything just works no matter what the use selects and no need for extra config goo when simply shipping a symlink means no config goo is required.

pbx0:
`W0209 17:49:32.376516    1488 manager.go:132] Unable to connect to Docker: dial unix /run/bla/bla/should/not/work.sock: no such file or directory` 
@kayrus Isn't this ^ an error from the kubelet meaning the argument was passed correctly?

I don't understand what those slice errors are about...

kayrus:
@pbx0 ah, pardon. didn't notice that. looks like kubelet falls back to the default value.

as for the slice error, me neither. didn't dig into this.

aaronlevy:
There was a mention that the script should not be called "kubelet" is there a preferred name?

app-admin/kubelet/files/kubelet
aaronlevy: @guilhem Hmm. I do see a symlink from /var/run -> /run in hyperkube:v1.1.7

marineam:
@aaronlevy I was more concerned about the package name, `app-admin/kublet-rkt` would avoid confusion with the existing kublet ebuild. Once the ebuild is renamed and given a version other than 9999 I think we are all set :)

crawford:
LGTM

aaronlevy:
Looks like I missed the boat on this one. In the future should I merge my own LGTM'd PRs?

marineam:
@aaronlevy yeah, sorry that wasn't clear. Since here in OS land we have so many repos and sometimes changes depend on something in another repo we always leave it to the author to do the final merge.

PR #1754, opened 2016-02-02
eclass: add get_semver() to cros-workon eclass

mischief:


marineam:
How about just put this in cros-workon and export a read-only variable like CROS_WORKON_VERSION

marineam:
Hm, actually, leave as a function since not all things using crosworkon will have tags. but that seems like the proper place for the function to live.

mischief:
ptal

marineam:
Just need to update the commit description and then LGTM

PR #1755, opened 2016-02-02
coreos-devel/mantle: bump to 0.0.9

mischief:
depends on #1754 but if that doesn't make it in i can copy/paste the semver stuff in here.


marineam:
Should be good, just need to update to match the current semver pr :)

marineam:
LGTM

PR #1765, opened 2016-02-09
Revert "dev-db/etcd: Set CROS_WORKON_COMMIT for arm64"

glevand:
etcd-2.2.5 builds for arm64, so a special conditional for it is no longer needed.
See: https://github.com/coreos/etcd/issues/3623


crawford:
LGTM

PR #1769, opened 2016-02-10
polkit: remove old polkit and spidermonkey

marineam:
Both of these have a number of bugs and were replaced a while ago.
Recently we started re-using them by accident by updating portage but
not pulling in this fix: https://bugs.gentoo.org/show_bug.cgi?id=560484

Remove them to prevent them from showing up again. :)


mischief:
lgtm

PR #1770, opened 2016-02-15
coreos/config: remove obsolete files

marineam:
The version script has been unused since coreos/scripts@2482291e.
The make.conf files have been unused since commit coreos/scripts@99ed29d2.
The others have been obsolete since before coreos even started.


mischief:
lgtm

PR #1771, opened 2016-02-16
coreos-kernel: Add verbose build

glevand:
To aid in debugging kernel builds set the kernel make variable
V=1 to get verbose build output when ECLASS_DEBUG_OUTPUT is
set.


crawford:
Nice. Good idea.

marineam:
How about we just make V=1 default? Disabling pretty build output is pretty common in Gentoo ebuilds in order to ensure build logs actually include relevant information.

marineam:
LGTM

PR #1773, opened 2016-02-17
Kublet wrapper and some cleanup

marineam:


mischief:
i believe arm64 does support TPMs, but that can be handled later.

lgtm

PR #1781, opened 2016-02-18
app-admin/kubelet-wrapper: add comment pointing to available tags

pbx0:


marineam:
Needs an ebuild version bump

pbx0:
@marineam bumped ebuild revision

marineam:
LGTM

PR #1782, opened 2016-02-19
Ignore unknown classes in SELinux base policy

mjg59:
We build a very minimal SELinux policy, so there are cases where code ends up
using classes that don't exist. We attempted to compensate for this case in
3da4994cd332c1f166c9a86ece0c62f9bc02c6f7 but it turns out that this isn't
used for initial policy build. Pass an explicit argument during base policy
build to handle that.

Fixes https://github.com/coreos/bugs/issues/1127


marineam:
LGTM

mischief:
retest this please

mischief:
retest this please

PR #1783, opened 2016-02-23
Cleanup oem related ebuilds and build them automatically

marineam:
Building components during image_to_vm.sh means we never upload binary packages for them.


crawford:
LGTM

PR #1793, opened 2016-02-25
app-admin/mayday: use current master for some new stuff

mischief:


crawford:
Why? We already have cros_workon.

mischief:
to bring in new changes into the image.

crawford:
If you use the live ebuild, it is going to use the code in tree. If you build 0.1.0, it's going to use that tag. I don't see how this will end up ever building 85f8b48da25fd6e3c36a9aa1f7d90c19078777ab.

mischief:
mayday is in manifest - it will build 85f8b48da25fd6e3c36a9aa1f7d90c19078777ab for =app-admin/mayday-0.1.0-r1.

crawford:
Oh, I didn't see that symlink. I see now.

Why are you including code that isn't a part of a release?

mischief:
i'll see about a release then. cc @brianredbeard

PR #1795, opened 2016-02-26
sys-apps/systemd: add --enable-silent-rules

vcaputo:
This flag makes build output less verbose (V=0), appreciated when
hacking on systemd and emerging manually.


marineam:
Hm, although on the other hand it is handy to default to full build logs so when something fails unexpectedly you have the context without having to try to reproduce it which is why the default, inherited from Gentoo, is to always pass --disable-silent-rules to configure scripts. On the plus side emerge happily passes pretty much anything in the environment through to the build so this works:

```
sudo V=0 emerge -v systemd
```

vcaputo:
That works, thanks!

PR #1796, opened 2016-02-26
app-emulation/wa-linux-agent: bump to 2.0.16

crawford:


marineam:
Since I switched from git to tarballs Manifest needs to be updated.

crawford:
Woops. Updated.

marineam:
LGTM

PR #1802, opened 2016-03-02
slotted openssl

mischief:
removes need to pass --enable-sslv2 to openssl configure.

bumps wget since the newer ebuild is the only one that uses a slotted dep.

depends on coreos/portage-stable#383


marineam:
Please split this into two PRs, the wget keyword can go in any time but the openssl change needs to wait.

mischief:
moved wget bump to #1804

marineam:
FYI, looks like upstream is changing course on this.

crawford:
Bump

PR #1805, opened 2016-03-02
profiles/coreos: accept app-misc/jq-1.5-r1

mischief:
depends on coreos/portage-stable#385

fixes coreos/bugs#455


marineam:
LGTM

PR #1809, opened 2016-03-03
coreos-base/coreos-cloudinit: re-add arm64 keyword

crawford:
This was dropped in 0cf04c447ce67f48f8546d745742ca22e5b8e603.


marineam:
LGTM

PR #1816, opened 2016-03-03
use slot operator for go dependencies

mischief:
also pruned old rkt/docker.


app-emulation/docker/docker-9999.ebuild
```diff
@@ -40,6 +40,8 @@ CDEPEND="
 DEPEND="
 	${CDEPEND}
 
+    dev-lang/go:=
```
marineam: Should be a tab

marineam:
Whitespace nit, otherwise LGTM

crawford:
LGTM

PR #1817, opened 2016-03-04
*: bump any ebuild that depends on go so it will be rebuilt

mischief:


marineam:
LGTM

Can you confirm that all of these bumped ebuilds are properly using the slot operator?

mischief:
ill double check..

mischief:
@marineam ptal

marineam:
LGTM

PR #1823, opened 2016-03-08
coreos-overlay: cease ForwardToConsole in journald

vcaputo:
This was introduced to assist users in troubleshooting on cloud
providers without an interactive console but with read-only access to
console output.

However, due to how journald implements the forwarding (synchronously)
it can have a dramatic impact on the overall logging performance.

Fixes https://github.com/coreos/bugs/issues/1162

### DIFF
--- a/coreos-base/oem-azure/files/grub.cfg
+++ b/coreos-base/oem-azure/files/grub.cfg
@@ -3,7 +3,7 @@
 set oem_id="azure"
 
 # Azure only has a serial console.
-set linux_console="console=ttyS0,115200n8 earlyprintk=ttyS0,115200 systemd.journald.forward_to_console=yes"
+set linux_console="console=ttyS0,115200n8 earlyprintk=ttyS0,115200"
 serial com0 --speed=115200 --word=8 --parity=no
 terminal_input serial_com0
 terminal_output serial_com0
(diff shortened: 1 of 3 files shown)


crawford:
Does this need to be applied to any other cloud providers (e.g. Azure)? If we make this change, we'll want to enable this option for the initramfs' instance of journald. That way we can get the logs from early boot, but not bog down the running system.

vcaputo:
oem-azure is in the mix, am I missing something?

vcaputo:
Another option would be to leave this be but add systemd.journald.max_level_console=err or something.

crawford:
Wow, I can't read. Explains a lot...

Ignition logs as low as Debug, so I don't think the `systemd.journald.max_level_console=err` is going to work well. I think this change is probably fine as is. We'll have to follow up with a change to bootengine.

vcaputo:
@crawford am I to interpret that as an lgtm?

crawford:
Yeah, this LGTM. We'll have to remind people that they will need to reprovision their machines if they want this change.

Can you also put together the change for bootengine and test it real quick before merging?

vcaputo:
So for the bootengine approach are you envisioning just tweaking the journald.conf in the initrd to have `ForwardToConsole=yes`?

crawford:
Yeah. I think that should work.

marineam:
Doesn't boot engine already have that tweak? Or at least it used to. Maybe
it got pruned at some point.

crawford:
It must not because we had to explicitly turn this on to get the Ignition logs.

brycechesternewman:
Is there a specific version of CoreOS this change is targeted for?

crawford:
This should be in tomorrow's Alpha.

PR #1824, opened 2016-03-08
app-emulation/docker: Set libdm_no_deferred_remove

glevand:
Apply docker build tag fixup from docker-1.7.1 (Commit
cd1fc3e4b2db741ed165420b1a4fb072b7690a51) and add arm64 keyword.

Fixes docker-1.10.2 builds for arm64.


glevand:
Rebased to latest.

marineam:
I love how Go's simple minded build system inspires people to re-implement autoconf poorly.

Although deferred removal and the dm backend aren't enabled by default I'd bet someone has so pulling it out of amd64 builds can potentially break systems. Could you make this conditional on `use arm64` or probably better yet `has_version '>=sys-fs/lvm2-2.02.110'`?

glevand:
Set it up as less than lvm2-2.02.110 and rebased to the latest.

glevand:
Should be OK now...

marineam:
Perfect, LGTM

PR #1829, opened 2016-03-09
Fix manifests, require strict checking from now on.

marineam:


mischief:
[LGTM](http://lgtm.in/i/K5Pd2aheq)

crawford:
LGTM

PR #1830, opened 2016-03-10
flannel: Support multi-arch

glevand:
Set flannel docker image from ebuild based on build architecture.

I set this up to use ```quay.io/coreos/arm64-flannel``` for the arm64 image.  We could use ```flannel-arm64```, or something else.


glevand:
To build the flannel arm64 docker image I added a flannel package and a 'flannelbox' build config to buildroot.  I think this is a better/easier way to build it than what flannel does now.  https://github.com/glevand/buildroot--buildroot

glevand:
Closing this PR since flannel has been reworked and this change is no longer relevant.

PR #1846, opened 2016-03-24
Update grub

mjg59:


mischief:
lgtm

mischief:
pending successful build that is

PR #1856, opened 2016-03-26
sys-apps/systemd: revert to 225

crawford:


marineam:
util-linux-2.27.1 should be fine, also how about leave the 229 ebuild for easy testing, just change the keywords to ~arch and grab the old ebuild and name it 225

marineam:
Assuming the ebuild is an exact match for what was removed, LGTM :)

crawford:
Double checked.

PR #1859, opened 2016-03-30
bump update_engine to latest

kdomanski:
Brings in a year's worth of fixes.

Adds glog as an RDEPEND of update_engine.
glog is built with `gflags` use flag enabled, so it keeps honoring
the `-logtostderr` command line flag in update_engine.


crawford:
The majority of the changes to update_engine should have been made in a topic branch and then merged. As it is, the master branch has not been tested fully. We need to finish the work on update_engine before we can pull in your other fixes. Sorry.

kdomanski:
@crawford Do you see another scenario where we can include coreos/update_engine@bc11ec7a04b68a7b1288413ab67811b0bdde00d0 into latest CoreOS? It fixes an update-breaking issue for me.

PR #1862, opened 2016-04-04
coreos-devel/mantle: enable Go vendor experiment

crawford:


mischief:
would there be any ill effects if we just did this in coreos-go.eclass?

crawford:
I'm not sure and I didn't care to test the entire distribution.

mischief:
lgtm

PR #1863, opened 2016-04-04
Add tpm policy generation code

mjg59:


app-crypt/tpmpolicy/tpmpolicy-20160404.ebuild
```diff
@@ -0,0 +1,19 @@
...
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="amd64 x86 arm64"
+IUSE=""
+
+S="${WORKDIR}"
+
+src_install() {
+    insinto /usr/sbin
+    doins "${FILESDIR}"/tpm_hostpolicy.sh
```
marineam: There is a `dosbin` shortcut for this which also ensures the right permission bits

app-crypt/tpmpolicy/files/tpm_hostpolicy.sh
```diff
@@ -0,0 +1,24 @@
+o -n '{'
```
marineam: Looks like copy/paste missed a bit here :)

app-crypt/tpmpolicy/tpmpolicy-20160404.ebuild
marineam: Also for a script that lands in the path I'd vote leave off the .sh

mjg59:
Ok, fixed those issues

app-crypt/tpmpolicy/files/tpm_hostpolicy
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
marineam: I'd recommend a `set -e` so it aborts if it can't actually read the tpm and so on.

mjg59: Done, along with an explicit TPM check and error message

marineam:
LGTM

PR #1864, opened 2016-04-05
sys-kernel/bootengine: bump to fix console logging

crawford:
This will allow us to pull in https://github.com/coreos/coreos-overlay/pull/1823.


mischief:
lgtm

PR #1866, opened 2016-04-05
sys-apps/systemd: bump to include recent fixes

crawford:
```
cf0d969 networkd: make sure we remove udev fd from epoll *before* closing it
2d7da74 sd-event: improve debug message when we fail to remove and fd from an epoll
03cf4f5 /dev/console must be labeled with SELinux label
1f1f4f2 containers: systemd exits with non-zero code
ccfefc9 journal: fix already offline check and thread leak
58fee71 needs-update: don't require strictly newer usr
```


crawford:
Fixes https://github.com/coreos/bugs/issues/1186.

mischief:
lgtm

marineam:
LGTM

PR #1867, opened 2016-04-05
coreos-{kernel,sources}: bump to 4.5

crawford:


crawford:
Should resolve https://github.com/coreos/bugs/issues/1156.

mischief:
it looks like a patch file is missing or the ebuild needs to be updated.

crawford:
It looks like the ebuild needs to be named 4.5.0 instead of 4.5.

mischief:
the naming only matters for coreos-sources but not coreos-kernel..? weird.

lgtm

crawford:
I updated the kernel ebuild to be 4.5.0 as well.

PR #1873, opened 2016-04-05
Include PAM support

mjg59:


crawford:
Why is PAM in coreos-overlay and not portage-stable?

mjg59:
Because we need to patch it to have the appropriate behaviour around locked
accounts and to install config in the right place

crawford:
Can you break that into two commits then - one pulling down upstream and the other making changes?

crawford:
(or is that what I am looking at)

mischief:
retest this please

mjg59:
Ok, this one creates /etc/environment in tmpfiles and has the update/modification split up.

marineam:
LGTM

PR #1874, opened 2016-04-06
sys-apps/baselayout: include change to unbreak boot

mischief:
the system was hanging on boot due to systemd removing a job to break a
dep cycle.


marineam:
LGTM

PR #1875, opened 2016-04-06
sys-kernel/bootengine: include latest fixes

crawford:
```
b7403d0 dracut: enable DHCPv6
d1315da dracut: make gpt headers valid before randomizing
```


marineam:
LGTM

PR #1877, opened 2016-04-06
sys-apps/systemd: include latest fixes

crawford:
```
b1ec77c nspawn: don't try to resolve passed binary before entering namespace
```

Fixes https://github.com/coreos/bugs/issues/1196.


marineam:
LGTM

PR #1878, opened 2016-04-07
sys-kernel/coreos-kernel: add seismograph dep

crawford:
Without this, cgpt isn't built in time and can't be installed in the initramfs.


marineam:
LGTM

PR #1883, opened 2016-04-11
profiles: do not wrap pkg-config as a legacy library config script

marineam:
This little bit of glue for handling old-style foo-config scripts was
assuming that anything in /usr/bin with a -config suffix was worth
wrapping. This is certainly not the case if the file in question is a
binary. This is most likely to cause confusion with pkg-config.


crawford:
LGTM

PR #1059, opened 2015-01-20
coreos-base/coreos-cloudinit: bump to v1.2.0

crawford:

PR #1885, opened 2016-04-13
coreos/coreos-overlay: build selinux with coreutils

askb:
This fixes [#1059](https://github.com/coreos/bugs/issues/1059) allows the selinux context information to be displayed.
Now `ls -Z` shows the correct output. ex:

```
core@coreos_developer_qemu-1010-0-0-2 ~ $ id -Z
system_u:system_r:kernel_t:s0
```

### DIFF
--- a/profiles/coreos/amd64/generic/package.use
+++ b/profiles/coreos/amd64/generic/package.use
@@ -4,3 +4,5 @@ sys-apps/dbus               selinux
 sys-apps/systemd            selinux
 sys-kernel/coreos-kernel    selinux
 
+# Enable SELinux for coreutils
+sys-apps/coreutils	    selinux


mischief:
cc @mjg59

mjg59:
As long as the dependency chain here doesn't end up with us pulling in Python

mischief:
test this please

marineam:
I'll double check whether base is the place for this or not, use.mask should ensure this isn't actually enabled in places it shouldn't bit may as well keep it grouped with wherever we toggle this on the systemd package.

profiles/coreos/base/package.use
```diff
@@ -58,3 +58,6 @@ sys-fs/squashfs-tools lzo xz
 
 # make the kernel version discoverable via the traditional gentoo symlink
 sys-kernel/coreos-sources symlink
+
+# enable selinux while building coreutils
+sys-apps/coreutils selinux
```
marineam: Please move this to coreos/amd64/generic/package.use with the other selinux flags.

askb: thanks - I have incorporated the changes.

askb:
@mjg59 - While testing I noticed that python2 is included now, though I am not sure if this is pertaining to the change. If python2 is not essential would you recommend a work around for removing this dependency ?

```
core@coreos_developer_qemu-1010-0-0-2 ~ $ python --version
Python 2.7.9
```
@mischief - the output of `id -Z` is in the PR message. :)

mischief:
@askb the message was for the CI bot to build the PR, sorry for the confusion :)

can you try to build a prod image and see if python is still there?

mischief:
maybe we can force unsetting the python use on libselinux.

marineam:
developer images always have python, is it being pulled into prod? I
haven't looked myself yet.

askb:
Building the prod image, I can confirm `python` does not get pulled into production with this change included.

```
core@coreos_production_qemu-1010-0-0- ~ $ id -Z
system_u:system_r:kernel_t:s0

core@coreos_production_qemu-1010-0-0- ~ $ python -v
-bash: python: command not found          
```

mjg59:
LGTM. Thanks for testing!

PR #1888, opened 2016-04-15
Add iSCSI

mjg59:


sys-block/open-iscsi/files/iscsid.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Open-iSCSI
+Documentation=man:iscsid(8) man:iscsiuio(8) man:iscsiadm(8)
+After=network.target NetworkManager-wait-online.service iscsiuio.service tgtd.service targetcli.service
```
mischief: can we get rid of `NetworkManager-wait-online.service` here? does open-iscsi ship tgtd or targetcli?

sys-fs/multipath-tools/files/multipath.rc
```diff
@@ -0,0 +1,27 @@
+#!/sbin/runscript
```
mischief: drop this file

sys-fs/multipath-tools/files/rc-multipathd
```diff
@@ -0,0 +1,21 @@
+#!/sbin/runscript
```
mischief: drop this file

sys-fs/multipath-tools/multipath-tools-0.5.0-r1.ebuild
```diff
@@ -0,0 +1,67 @@
...
+		libudevdir='${prefix}'/"${udevdir}" \
+		install
+
+	insinto /usr/share/multipath-tools
+	newins "${S}"/multipath.conf.annotated multipath.conf
+	# /etc/udev is reserved for user modified rules!
+	mv "${D}"/etc/udev/rules.d "${D}/${udevdir}"/ || die
+	fperms 644 "${udevdir}"/rules.d/66-kpartx.rules
+	newinitd "${FILESDIR}"/rc-multipathd multipathd
+	newinitd "${FILESDIR}"/multipath.rc multipath
```
mischief: drop newinitd's

sys-block/open-iscsi/files/iscsid.service
mjg59: We can, but we're divergent from upstream in that case - After= is only relevant if those other services exist and are enabled, it'll still start fine without them.

marineam:
LGTM

PR #1889, opened 2016-04-15
app-admin/kubelet-wrapper: mount /etc/os-release

pbx0:
See: https://github.com/coreos/coreos-kubernetes/issues/399

@philips @aaronlevy


crawford:
Does this actually work? `/etc/os-release` is a symlink to `../usr/lib/os-release`.

marineam:
Might work if rkt/nspwan followes symlinks before passing them to mount. But might as well point to the real location under /usr

pbx0:
It does work, but I'll change it to the real location

```
sudo rkt run --interactive --volume os-release,kind=host,source=/etc/os-release  --mount volume=os-
...
root@rkt-0643c3a3-5b2b-4d95-96f8-838e9361208b:/# cat /etc/os-release
NAME=CoreOS
ID=coreos
VERSION=1010.1.0
VERSION_ID=1010.1.0
BUILD_ID=2016-04-07-0355
PRETTY_NAME="CoreOS 1010.1.0 (MoreOS)"
ANSI_COLOR="1;32"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://github.com/coreos/bugs/issues"
```

marineam:
LGTM

PR #1891, opened 2016-04-18
Copy pam.d in tmpfiles

mjg59:
Restore pam.d after /etc is cleaned.


sys-libs/pam/files/tmpfiles.d/pam.conf
```diff
@@ -8,3 +8,4 @@ L /etc/security/limits.conf	-    -	  -	-	../../usr/lib/pam/limits.conf
 L /etc/security/namespace.conf	-    -	  -	-	../../usr/lib/pam/namespace.conf
 L /etc/security/pam_env.conf	-    -	  -	-	../../usr/lib/pam/pam_env.conf
 L /etc/security/time.conf	-    -	  -	-	../../usr/lib/pam/time.conf
+C /etc/pam.d			-    -	  -	-	/usr/lib/pam.d
```
marineam: Why is this a copy? Doesn't pam read lib/pam.d directly now? In theory `d /etc/pam.d` is the only rule needed.

mjg59: Ugh yes how had I forgotten that

mischief:
lgtm

marineam:
LGTM

PR #1893, opened 2016-04-19
oem-exoscale: make DHCP address retrieval more resilient

vincentbernat:
When retrievent the DHCP IP address (to get user data), two steps were
used: the first one waited for one lease file to become present and the
second one extracted and validated the server IP address. If for some
reason the second step failed, an empty IP was provided (and notably,
coreos-setup-environment.service never finishes correctly).

To avoid that, we combine the two steps into a loop: for each available
lease file, extract the IP and validate it. If it succeeds, return
it. If not continue. If nothing was found, restart the same process
after sleeping a bit.


marineam:
LGTM, could you bump the version of the ebuild please?

vincentbernat:
Done.

runseb:
#1901 is the same but for the generic cloudstack build

marineam:
LGTM, thanks!

PR #1895, opened 2016-04-19
sys-kernel/coreos-{sources,kernel}: update to 4.5.1

mischief:
0021-Fix-unallocated-memory-access-in-TPM-eventlog-code.patch is dropped
because it is fixed in upstream commit
186d124f07da193a8f47e491af85cb695d415f2f and backported in the 4.5.1
patch set.

kernel defconfigs have been updated with kernel_menuconfig script.


mjg59:
LGTM

PR #1898, opened 2016-04-19
enable pl2303 and pids/hugetlb controller

mischief:
pl2303 will help me use coreos as a console server, since that's the adapter we have.

pids cgroup controller is used by the recent TasksMax & co feature of systemd.

i figure that we should have hugetlb controller too for completeness.


marineam:
LGTM

PR #1901, opened 2016-04-21
Update cloudstack-dhcp

runseb:
Fixes race condition in finding network interface of virtual router


marineam:
This one also needs an ebuild version bump

runseb:
@marineam ok thanks, done.

marineam:
LGTM, thanks!

PR #1902, opened 2016-04-21
profiles: fix config script wrappers

marineam:
Commit 631890b9 was completely broken and started skipping all *-config
file, not just pkg-config, because I forgot to specify the file path.


crawford:
LGTM

PR #1905, opened 2016-04-22
coreos-devel/mantle: bump to v0.1.0

mischief:


coreos-devel/mantle/mantle-9999.ebuild
```diff
@@ -20,6 +20,8 @@ DESCRIPTION="Mantle: Gluing CoreOS together"
 HOMEPAGE="https://github.com/coreos/mantle"
 LICENSE="Apache-2"
 SLOT="0"
+# objcopy/split have trouble with our cross-compiled kolet
```
marineam: You can define a QA_something var to specifically exclude the arm binary. That way the rest still gets stripped.

marineam: QA_PREBUILT="/usr/lib/kola/*" should do the trick

mischief: oh, i asked some gentoo folks and they said `STRIP_MASK`

marineam:
LGTM

PR #1907, opened 2016-04-26
dev-lang/go{,-bootstrap}: bump bootstrap to 1.5.3

mischief:
go 1.5+ can't be compiled without go, and so does not work on arm64
systems. switch go-bootstrap to use a binary package, and use that to
build a new go.


marineam:
Is 1.5.4 indeed the version contained in go-linux-amd64-bootstrap.tbz and so on?

dev-lang/go-bootstrap/go-bootstrap-1.5.4.ebuild
```diff
@@ -0,0 +1,85 @@
...
+# $Id$
+
+EAPI=6
+
+export CTARGET=${CTARGET:-${CHOST}}
+
+inherit eutils toolchain-funcs
+
+BOOTSTRAP_DIST="https://dev.gentoo.org/~williamh/dist"
+SRC_URI="kernel_linux? (
```
marineam: Since there aren't any alternate kernel keywords no point to having this flag.

dev-lang/go-bootstrap/go-bootstrap-1.5.4.ebuild
```diff
@@ -0,0 +1,85 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=6
+
+export CTARGET=${CTARGET:-${CHOST}}
```
marineam: Not used

dev-lang/go-bootstrap/go-bootstrap-1.5.4.ebuild
```diff
@@ -0,0 +1,85 @@
...
+inherit eutils toolchain-funcs
+
+BOOTSTRAP_DIST="https://dev.gentoo.org/~williamh/dist"
+SRC_URI="kernel_linux? (
+	amd64? ( ${BOOTSTRAP_DIST}/go-linux-amd64-bootstrap.tbz )
+	arm64? ( ${BOOTSTRAP_DIST}/go-linux-arm64-bootstrap.tbz )
+)
+"
+
+KEYWORDS="-* amd64 ~arm arm64 ~x86"
```
marineam: Drop arm and x86 since you don't have a tarball for them.

mischief:
```
$ curl -s http://distfiles.gentoo.org/distfiles/go-linux-amd64-bootstrap.tbz | tar xjO go-linux-amd64-bootstrap/VERSION
go1.5.3
```

oops, thanks for making me double check :)

mischief:
ptal

marineam:
Updated the version everywhere except the commit message

marineam:
Other than that lgtm

PR #1908, opened 2016-04-27
raise low default (512) as of systemd 228+ on kernel 4.3+

bjaglin:
Applying upstream change https://github.com/docker/docker/pull/19391 to prevent `Resource temporarily unavailable` errors when running a large number of containers and/or containers with a large number of processes.

### DIFF
--- a/app-emulation/docker/files/docker.service
+++ b/app-emulation/docker/files/docker.service
@@ -10,6 +10,7 @@ EnvironmentFile=-/run/flannel_docker_opts.env
 MountFlags=slave
 LimitNOFILE=1048576
 LimitNPROC=1048576
+TasksMax=1048576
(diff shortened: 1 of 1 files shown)


mischief:
1048576 seems like quite a large number, what's the issue you're running into?

bjaglin:
I am running a CI workload (around ~100 containers, each with a few forks I assume) and run into this very fast. As far as I understand, this was not bound by default before systemd 228, so the upstream PR aims at bringing this behavior back.

* https://github.com/systemd/systemd/pull/1239
* https://github.com/systemd/systemd/pull/1886

bjaglin:
Actually, the PR above was reverted because of the systemd backward-compatibility issue (<226), but the upstream [recommended value](https://github.com/docker/docker/pull/21491/files) is `infinity` to get the pre-228 behavior.

crawford:
Closing in favor of https://github.com/coreos/coreos-overlay/pull/1912.

PR #1909, opened 2016-04-27
Include upstream patch to whitelist devices with SCSI_IDENT property

mjg59:
Issue [#1236](https://github.com/coreos/bugs/issues/1236).


mischief:
lgtm

PR #1910, opened 2016-04-27
ntp  4.2.8p7

mischief:
fixes the CVEs in https://support.ntp.org/bin/view/Main/SecurityNotice#April_2016_NTP_4_2_8p7_Security


marineam:
LGTM

PR #1919, opened 2016-04-28
google-*: import eutils, fix applying required patches

marineam:
Broken back in 02c004e846, newly installed GCE images haven't worked
right ever since then. :(

### DIFF
--- a/app-emulation/google-compute-daemon/google-compute-daemon-1.1.1-r4.ebuild
+++ b/app-emulation/google-compute-daemon/google-compute-daemon-1.1.1-r4.ebuild
@@ -3,6 +3,8 @@
 
 EAPI=5
 
+inherit eutils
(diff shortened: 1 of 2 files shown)


mischief:
lgtm

stvnwrgs:
@marineam actually, there is only `coreos-alpha-1032-1-0-v20160505` on gce as an image, but i still have `coreos-alpha-1045-0-0-v20160512` in my config.  Got this release revoked? Do you know when we can expect a working version?

mischief:
@stvnwrgs a new alpha should be out thursday. 1045/1047 were affected by https://coreos.com/blog/alpha-security-incident-subset-of-users-affected.html.

marineam:
Beta will be receiving this fix soon too.

PR #1921, opened 2016-05-01
app-emulation/rkt: v1.5.1

mischief:
- bump coreos stage1 flavor source to 1032.0.0 for systemd 229
- fetch image over https


crawford:
Can you bump to v1.5.1? Otherwise, it looks good to me. How does Team Rocket feel?

tmrts:
cc @iaguis @alban @jonboulle

marineam:
LGTM

iaguis:
We didn't update the coreos flavor image to use systemd v229 in v1.5.1 (https://github.com/coreos/rkt/pull/2514). In that PR, we add a missing library (`libpam`) to stage1 so I'm not sure this works.

alban:
I agree with Iago, we should first do a rkt release with https://github.com/coreos/rkt/pull/2514 merged before this change.

This could introduce a regression with libpam because rkt without https://github.com/coreos/rkt/pull/2514 does not take `lib64/libpam.so` from the pxe image.

PR #1928, opened 2016-05-02
sys-apps/baselayout: fix arm64 builds

mjg59:
The symlink for nsswitch can't be resolved during build, so use the target
instead.


marineam:
Doh, LGTM

PR #1929, opened 2016-05-02
backports for 899

mischief:
- jq 1.5-r2
- git 2.7.3-r1

git is moved here because the upstream ebuild uses EAPI=6 and we don't have that in the 899 branch.


mischief:
depends on coreos/portage-stable#416

crawford:
LGTM

PR #1930, opened 2016-05-03
sys-apps/baselayout: Fix argument order to sed

mjg59:
I've been doing this for 20 years I'm a trained professional


mischief:
lgtm

PR #1937, opened 2016-05-03
backports for 1010

mischief:
depends on coreos/portage-stable#417


mischief:
depends on coreos/portage-stable#418 actually

crawford:
LGTM

PR #1939, opened 2016-05-04
enable go compiler on ppc64 and x86 arch

houstar:
1. go-1.6.2 dropped go-boostrap package. Then we can build go without gccgo and go-bootstrap.
2. enable ppc64 and x86 arch detection in eclass/coreos-go.

### DIFF
--- a/dev-lang/go/go-1.6.2.ebuild
+++ b/dev-lang/go/go-1.6.2.ebuild
@@ -1,0 +1,256 @@
+# Copyright 1999-2016 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=6
+
+export CBUILD=${CBUILD:-${CHOST}}
+export CTARGET=${CTARGET:-${CHOST}}
+
+MY_PV=${PV/_/}
+
+inherit toolchain-funcs
+
+BOOTSTRAP_DIST="https://dev.gentoo.org/~williamh/dist"
+SRC_URI="!gccgo? (
+kernel_Darwin? (
+	x64-macos? ( ${BOOTSTRAP_DIST}/go-darwin-amd64-bootstrap.tbz )
+)
+kernel_FreeBSD? (
+amd64-fbsd? ( ${BOOTSTRAP_DIST}/go-freebsd-amd64-bootstrap.tbz )
+x86-fbsd? ( ${BOOTSTRAP_DIST}/go-freebsd-386-bootstrap-1.tbz )
+)
+kernel_linux? (
+	amd64? ( ${BOOTSTRAP_DIST}/go-linux-amd64-bootstrap.tbz )
+	arm? ( ${BOOTSTRAP_DIST}/go-linux-arm-bootstrap.tbz )
+	arm64? ( ${BOOTSTRAP_DIST}/go-linux-arm64-bootstrap.tbz )
+	ppc64? (
+		${BOOTSTRAP_DIST}/go-linux-ppc64-bootstrap.tbz
+		${BOOTSTRAP_DIST}/go-linux-ppc64le-bootstrap.tbz
+	)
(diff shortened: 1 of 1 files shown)


marineam:
@houstar in order to support cross-compiling Go itself we do still use the go-bootstrap package, but updated to use the bootstrap binary builds instead of building 1.4 from source as the old go-bootstrap package did. So the 1.6 ebuild will need to be derived from our current 1.5 one, not directly from upstream.

marineam:
Created https://github.com/coreos/coreos-overlay/pull/1951 without the 1.6 ebuild

houstar:
@marineam go-1.6.2 and go-9999 ebuild from upstream has been updated to use the bootstrap binary builds instead of building 1.4 from source as the old go-bootstrap package did.  I would like to derived from upstream because it's already supported cross-compiling . See upstream below. 

BOOTSTRAP_DIST="https://dev.gentoo.org/~williamh/dist"
SRC_URI="!gccgo? (
kernel_Darwin? (
	x64-macos? ( ${BOOTSTRAP_DIST}/go-darwin-amd64-bootstrap.tbz )
)
kernel_FreeBSD? (
amd64-fbsd? ( ${BOOTSTRAP_DIST}/go-freebsd-amd64-bootstrap.tbz )
x86-fbsd? ( ${BOOTSTRAP_DIST}/go-freebsd-386-bootstrap-1.tbz )
)
kernel_linux? (
	amd64? ( ${BOOTSTRAP_DIST}/go-linux-amd64-bootstrap.tbz )
	arm? ( ${BOOTSTRAP_DIST}/go-linux-arm-bootstrap.tbz )
	arm64? ( ${BOOTSTRAP_DIST}/go-linux-arm64-bootstrap.tbz )
	ppc64? (
		${BOOTSTRAP_DIST}/go-linux-ppc64-bootstrap.tbz
		${BOOTSTRAP_DIST}/go-linux-ppc64le-bootstrap.tbz
	)
	x86? ( ${BOOTSTRAP_DIST}/go-linux-386-bootstrap-1.tbz )
)
kernel_SunOS? (
	x64-solaris? ( ${BOOTSTRAP_DIST}/go-solaris-amd64-bootstrap.tbz )
)
)
"

marineam:
@houstar right, and our go-bootstrap ebuild uses those pre-built tarballs now: https://github.com/coreos/coreos-overlay/blob/master/dev-lang/go-bootstrap/go-bootstrap-1.5.3.ebuild

houstar:
@marineam upstream has already dropped go-bootstrap packages and from go-1.6 version its integrated with go-bootstrap which is supported ppc64 & ppc64le, etc. Why we reinvent the wheel ? And coreos go-bootstrap ebuild just supported amd64 and arm64.  I would like to drop go-bootstrap packages as upstream does.  so this will be convenient to enable CoreOS ppc64le. Thanks

https://github.com/gentoo/gentoo/blob/master/dev-lang/go/go-1.6.2.ebuild

marineam:
@houstar the bundled bootstrap scheme upstream has fetches a bootstrap tarball for the same architecture as the ebuild's target, making cross-compiling impossible. Our scheme depends on go-bootstrap being installed in the SDK and the go ebuild always uses it, even when it is installing to a different $ROOT with a different arch.

(Due to the there being no distinction between CBUILD and CHOST dependencies in DEPEND, the CHOST's go-bootstrap will get installed to $ROOT but not actually used.)

houstar:
@marineam  According above, so how about enable go-bootstrap on ppc64le ?

bgilbert:
We're now working on [Fedora CoreOS](https://groups.google.com/d/topic/coreos-user/zgqkG88DS3U/discussion), the successor to Container Linux, and we expect most major development to occur there instead.  Meanwhile, Container Linux will be fully maintained into 2020 but won't see many new features.  Thank you for submitting this PR and we're sorry that we won't be able to merge it.

PR #1947, opened 2016-05-05
sec-policy: Permit execmem in selinux policy

mjg59:
polkit is failing when selinux is enforcing as it is attempting to mmap
pages as both writable and executable and selinux is forbidding this.
Since we want selinux for container isolation rather than general system
confinement, the easiest fix for now is to just add the selinux boolean
to permit execmem.

The selinux eclass is modified to hardcode the gentoo patchset that we're
basing our policy on - otherwise bumping the revision for our local
builds tries to pull down versions that don't exist.


mischief:
lgtm

PR #1950, opened 2016-05-09
grub: enable building as a cross-compiled package for arm64

marineam:
This will drop the need to reconfigure/rebuild GRUB with the special
`arm64` target in `setup_board`. The amd64 targets could receive the
same treatment eventually but the xen dependency needs fixing first.
After that the SDK's grub can be built with the special `none` target so
it only builds user-space tools.


mischief:
lgtm

profiles/coreos/targets/sdk/make.defaults
```diff
@@ -3,3 +3,12 @@ USE="man -pam"
 # Used by some old goo in profiles/coreos/base/profile.bashrc
 # TODO: clean up that old goo
 CROS_SDK_HOST="cros-sdk-host"
+
+# Enable our assorted GRUB targets
+GRUB_PLATFORMS="efi-64 pc xen"
```
crawford: Why is this included here?

PR #1951, opened 2016-05-10
eclass/coreos-go: add ppc64 and x86 arch to go_get_arch

marineam:
Just the eclass update from https://github.com/coreos/coreos-overlay/pull/1939


crawford:
LGTM

PR #1953, opened 2016-05-10
app-emulation/google-compute-daemon: bump to 1.3.2

crawford:


app-emulation/google-compute-daemon/files/0003-Fix-path-to-ip.patch
```diff
@@ -0,0 +1,35 @@
...
+--- a/google-daemon/usr/share/google/google_daemon/address_manager.py
++++ b/google-daemon/usr/share/google/google_daemon/address_manager.py
+@@ -49,7 +49,7 @@ class AddressManager(object):
+     self.system = system_module
+     self.urllib2 = urllib2_module
+     self.time = time_module
+-    self.ip_path = '/sbin/ip'
++    self.ip_path = '/usr/bin/ip'
+     if not os.path.exists(self.ip_path):
+         self.ip_path = '/bin/ip'
```
marineam: The /bin/ip path will work so we can drop this patch.

marineam:
Are you planning on updating the startup scripts ebuild to? I'm guessing the two shouldn't be out of sync but startup-scripts is a weirder set of goo and not all of it may be applicable.

crawford:
Probably should update that other ebuild.

crawford:
Pruned that patch and updated the scripts.

crawford:
I also sent one of the patches [upstream](https://github.com/GoogleCloudPlatform/compute-image-packages/pull/235).

app-emulation/google-startup-scripts/google-startup-scripts-1.3.2.ebuild
```diff
@@ -19,12 +19,10 @@ RDEPEND="dev-lang/python-oem"
...
+	epatch "${FILESDIR}/0001-Allow-location-of-startup-scripts-to-be-overridden.patch"
 }
 
 src_install() {
-	mkdir -p ${D}/usr/share/oem/google-startup-scripts/usr/share/google/
-	cp -Ra ${WORKDIR}/usr/share/google/. ${D}/usr/share/oem/google-startup-scripts/ || die
-	# We don't install python or gsutil so skip this
-	rm -R ${D}/usr/share/oem/google-startup-scripts/boto || die
+	into "/usr/share/oem/google-startup-scripts/"
+	find "${S}/usr/share/google" -type f | xargs dobin || die
```
crawford: Is this too ugly?

crawford:
```
ERROR: Publisher com.cloudbees.jenkins.GitHubCommitNotifier aborted due to exception
java.io.IOException: Cannot determine sha1 of the commit. The status cannot be reported
```

Build passes.

marineam:
LGTM, does this update depend on the other GCE changes?

crawford:
No, this one can stand on its own.

PR #1955, opened 2016-05-11
coreos-base/oem-gce: remove legacy scripts

crawford:
Remove the cloud-config (will be taken over by Ignition), the /etc/hosts
script (will be taken over by Ignition), and the SSH key script (will be
taken over by coreos-metadata).


crawford:
This one needs to be merged with a new Ignition.

marineam:
LGTM

PR #1958, opened 2016-05-13
app-emulation/rkt: use CoreOS alpha 1010.1.0 without PAM

mischief:
CoreOS 1029.0.0 introduced PAM which was linked into systemd. rkt copies
files and libraries out of CoreOS images with a static manifest to
construct the CoreOS stage1. since this manifest lacks PAM libraries in
rkt 1.5.1, we can't use CoreOS 1032.0.0, so fall back to a CoreOS
version without PAM linked into systemd.


mischief:
fixes coreos/bugs#1283

crawford:
Have we done any testing with 225? Why not just go back to 222?

mischief:
@crawford upstream rkt 1.5.1 is using 225 so i think it's best to stick with what they used.

crawford:
Ok. LGTM

PR #1959, opened 2016-05-13
app-admin/kubelet-wrapper: fix stage1 reference

mischief:
since rkt 1.3.0, rkt has began checking stage1 hashes, and will do so if
the stage1 is not in the default stage1 directory. for compatibility,
df131e85d5240a75fb9cd4e1fb85e86e6f5290c1 symlinked the old stage1
directory to the new one, however, if a stage1 is referred to by the old
stage1 path under rkt >1.3.0, image verification will fail because no
signature exists.

instead of using a path, use stage1-fly.aci by name and let rkt handle
finding it in the default directory.


mischief:
fixes coreos/bugs#1282

dghubble:
LGTM

crawford:
LGTM

PR #1963, opened 2016-05-14
google-*: import eutils, fix applying required patches

marineam:
Broken back in 02c004e846, newly installed GCE images haven't worked
right ever since then. :(

Fixes https://github.com/coreos/bugs/issues/1285


mischief:
lgtm

PR #1965, opened 2016-05-16
sys-auth/pambase: Fall back to pam_deny in auth

mjg59:
Setting pam_unix and pam_sss to sufficient means that if both fail,
control will be passed to the following pam module. If this is
pam_permit then permission will be granted even if the previous modules
failed. Switch to pam_deny and require it rather than permitting it to
be optional - if sss or unix succeed, we'll jump out before we get to
this point.


marineam:
What is the difference between this scheme of using `sufficient` vs the way pambase optionally includes krb5 support?

From `basic-conf`:
```
#define KRB5_PARAMS DEBUG ignore_root try_first_pass

/* By using the extended Linux-PAM syntax for this, it is possible to
   fine-tune the Kerberos handling so that it works out of hte box on
   most desktop systems.

   What this control operation does is ignore failures and errors from
   Kerberos (falling back on local pam_unix auth), but if it's good,
   it'll skip over the following module (pam_unix) with an accepted
   status.

   IMPORTANT! Make sure that the only thing that comes right after
   pam_krb5 with KRB5_CONTROL is pam_unix!
 */
#define KRB5_CONTROL [success=1 default=ignore]
```

Which expands into this for `system-auth`:
```
uth            required        pam_env.so
auth            [success=1 default=ignore]      pam_krb5.so  ignore_root try_first_pass
auth            required        pam_unix.so try_first_pass likeauth nullok
auth            optional        pam_permit.so

account         [success=1 default=ignore]      pam_krb5.so  ignore_root try_first_pass
account         required        pam_unix.so
account         optional        pam_permit.so

password        [success=1 default=ignore]      pam_krb5.so  ignore_root try_first_pass
password        required        pam_unix.so try_first_pass use_authtok nullok md5 shadow
password        optional        pam_permit.so

session         required        pam_limits.so
session         required        pam_env.so
session         [success=1 default=ignore]      pam_krb5.so  ignore_root try_first_pass
session         required        pam_unix.so
session         optional        pam_permit.so
-session        optional        pam_systemd.so
```

marineam:
That asside, given the complicated ifdef magic I'm thinking we should just stick the plain, not-so-magic files in baselayout. pambase could then go in package.provided for compatibility with Gentoo ebuilds.

marineam:
Basically worried that toggling a use flag some day will change behavior dramatically in some non-obvious way.

mjg59:
The scheme I've added is consistent with the upstream documentation, and given our experiences diverging from that in the past I'm kind of keen on sticking as close to that as possible

marineam:
Agreed, the `success=1` rule jumping kinda scares me, particularly since originally I thought `1` was a boolean, not a jump.

marineam:
Alright, LGTM, this should be good for this week's release.

I'm going to follow up afterwards with moving this to baselayout for the sake of easier maintenance.

PR #1967, opened 2016-05-16
Linux 4.6

mischief:
should fix coreos/bugs#1092, maybe some others.


mischief:
crap, need to fix patch paths.

mischief:
fixed. ptal.

marineam:
LGTM

PR #1969, opened 2016-05-17
Create overlay.xml

necrose99:
layman -f -a spike -o https://raw.github.com/coreos/coreos-overlay.git/master/overlay.xml 
Gentoo or Sabayon etc users can use away... and any specials from Coreos via overlay...


marineam:
Ideally our portage-stable repo would be tracking gentoo fairly closely so this swapping it for gentoo outright would work but that is not the case right now and I don't want to give the impression that mixing coreos-overlay with anything other than portage-stable is expected to work.

If you really do want to track coreos-overlay on your system you can add it to repos.conf and emerge --sync will just work. No need for layman in the first place!

PR #1970, opened 2016-05-17
coreos-base/coreos-dev: remove dhcp

mjg59:
New versions of dhcp are difficult to cross build and we don't need it
anyway


marineam:
LGTM

PR #1975, opened 2016-05-18
profiles: stop disabling SHA512 password hashes in PAM

marineam:
Likely inherited from ChromeOS but even for them it is a completely
ridiculous flag to disable. We had SHA512 enabled pre-PAM since shadow
does not have this use flag so this restores previous behavior.


mischief:
lgtm

PR #1978, opened 2016-05-19
net-misc/google-cloud-sdk: add gcloud and gsutil

crawford:


marineam:
Is this for use in the SDK? How does packaging this mesh with the fact that gcloud is it itself a package manager?

net-misc/google-cloud-sdk/google-cloud-sdk-108.0.0.ebuild
```diff
@@ -0,0 +1,36 @@
...
+KEYWORDS="amd64"
+IUSE=""
+
+S="${WORKDIR}/${PN}"
+
+DEPEND="${PYTHON_DEPS}"
+RDEPEND="${DEPEND}"
+
+src_prepare() {
+	epatch "${FILESDIR}/0001-Hardcode-the-root-directory.patch"
```
marineam: git add

crawford:
@marineam yeah, for use in the SDK. The package management functionality just errors out.

```
$ gcloud components list
ERROR: (gcloud.components.list) The components management action could not be performed because the installation root of the Cloud SDK could not be located. If you previously used the Cloud SDK installer, you could re-install the the SDK and retry again.
```

marineam:
Ok, as long as the magic doesn't do anything ;-)

LGTM

Though this isn't providing gsutil in the path but that can be a follow up, replacing the old gsutil ebuild.

PR #1979, opened 2016-05-19
profiles: update protobuf from 1.5 to 1.6

marineam:
The newer version is needed for arm64, and need the same version across
all architectures because the build host and target's version must match.


mischief:
lgtm

PR #1986, opened 2016-05-23
sec-policy/selinux-virt: Allow setattr on devpts ptys and grant pipef…

mjg59:
…s access

Policy was blocking the modification of attributes on devpts ptys, making it
impossible to enter a rkt container interactively. Fix that. In addition,
pipefs access is being blocked which makes Docker unhappy. Fix that too.


marineam:
Looks plausibly correct!

PR #1991, opened 2016-05-26
app-emulation/rkt: backport /etc/rkt permission patch

mischief:
patch should be removed when rkt is bumped to the next release.


mischief:
superseded by #1994

PR #1992, opened 2016-05-26
sys-fs/e2fsprogs: Moving to overlay with patch

ajeddeloh:
moving e2fsprogs to overlay from portage_stable. Adds -p flag to protect
exisiting partitions if mke2fs is run non-interactively, instead of just
overwriting them.


mischief:
can you sync sys-fs/e2fsprogs seperately from modifying the ebuild and applying the patch?

is this patch upstream or going to be?

ajeddeloh:
I will upstream it. It needs a minor tweak (the patch for 1.42.13 wont work for their master branch, but it's a trivial change).

sys-fs/e2fsprogs/files/e2fsprogs-1.42.13-protect-existing-fs.patch
```diff
@@ -0,0 +1,90 @@
...
+ ]
+ [
+@@ -608,6 +611,14 @@ For more information about the features which can be set, please see
+ the manual page
+ .BR ext4 (5).
+ .TP
++.B \-p
++Protect existing partitions, even if run in a script. Prevents
++.B mke2fs
++from overwriting existing partitions, even if running in a script. If a
```
marineam: I don't think "even if run in a script" needs to be repeated twice. Also the deciding factor is if there is a TTY, not a script right?

sys-fs/e2fsprogs/files/e2fsprogs-1.42.13-protect-existing-fs.patch
```diff
@@ -0,0 +1,90 @@
...
+-	if (isatty(0) && isatty(1))
++	if ((isatty(0) && isatty(1)) || protect)
+ 		flags |= CHECK_FS_EXIST;
+ 	if (!quiet)
+ 		flags |= VERBOSE_CREATE;
+ 	if (fs_blocks_count == 0)
+ 		flags |= NO_SIZE;
+ 	if (!check_plausibility(device_name, flags, &is_device) && !force)
+-		proceed_question(proceed_delay);
++		if (protect)
```
marineam: does check_plausibility report an error message?

ajeddeloh: Yes, it will print the error message that you'd get if you were running it form a tty on stdout before exiting.

sys-fs/e2fsprogs/files/e2fsprogs-1.42.13-protect-existing-fs.patch
ajeddeloh: correct, I've fixed it now

crawford:
LGTM

marineam:
LGTM

PR #1993, opened 2016-05-27
arm64 thunderx kernel updates

mischief:
cc @glevand @marineam


glevand:
LGTM

marineam:
Should we enable acpi?

mischief:
@marineam: the default of the ACPI symbol in kernel Kconfig is actually 'y', but it has `depends on IA64 || X86 || (ARM64 && EXPERT)`. so, enabling `EXPERT` here has the effect of enabling ACPI, among other things that have boolean conditions on `EXPERT`. :shipit:

PR #1995, opened 2016-06-01
sec-policy/selinux-virt: Grant more permissions on chr_files

mjg59:
apt seems to use character device nodes for a couple of things, so give the full set
of permissions for them when operating in the container's own context.


marineam:
LGTM

PR #1997, opened 2016-06-03
WIP: add etcd wrapper

crawford:


app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,55 @@
...
+
+source /usr/share/coreos/update.conf
+source /etc/coreos/update.conf
+
+ETCD_VERSION="${ETCD_VERSION:-${GROUP:-stable}}"
+ETCD_ACI="${ETCD_ACI:-quay.io/coreos/etcd}"
+
+case ${ETCD_VERSION} in
+	stable)
+		ETCD_VERSION=v2.3.5
```
crawford: This isn't v2.3.6 because I get the following error:

```
run: requested value for label "version": "v2.3.6" differs from fetched aci label value: "latest"
```

It looks like we end up downloading "latest" instead of "v2.3.6" even though they are the same thing.

crawford:
This doesn't address the `localhost` issue. Ideally, we would be able to append `127.0.0.1       localhost` to the host's host file, but then that would require us making a copy of the file instead of actually mounting it through.

crawford:
I also noticed that all stdout/stderr are logged under a different machine-id. Ideally, this would pipe straight through so users can run this interactively.

/cc @jonboulle

app-admin/etcd-wrapper/etcd-wrapper-0.0.1.ebuild
```diff
@@ -0,0 +1,24 @@
+#
+# Copyright (c) 2015 CoreOS, Inc.. All rights reserved.
```
xiang90: 2016 CoreOS

app-admin/etcd-wrapper/etcd-wrapper-0.0.1.ebuild
```diff
@@ -0,0 +1,24 @@
...
+# Distributed under the terms of the GNU General Public License v2
+# $Header:$
+#
+
+EAPI=6
+
+DESCRIPTION="etcd Wrapper"
+HOMEPAGE="https://github.com/coreos/etcd"
+KEYWORDS="amd64"
```
glevand: I'd like to see arm64 in the keywords here.

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,55 @@
...
+		echo "${USAGE}"
+		exit 0
+		;;
+esac
+
+source /usr/share/coreos/update.conf
+source /etc/coreos/update.conf
+
+ETCD_VERSION="${ETCD_VERSION:-${GROUP:-stable}}"
+ETCD_ACI="${ETCD_ACI:-quay.io/coreos/etcd}"
```
glevand: How are you going to publish arm64 images? ```quay.io/coreos/etcd_arm64```?  I think we need to set ETCD_ACI via a conditional on the arch.

crawford: Woops, this should actually be `coreos.com/etcd`. rkt will fetch the correct architecture given the target. I'll update this.

app-admin/etcd-wrapper/etcd-wrapper-0.0.1.ebuild
crawford: I only left it out since it depends on rkt. I guess it doesn't hurt to include it - it just won't run properly yet.

glevand: Hopefully all will work OK soon https://github.com/glevand/coreos--rkt/commits/master...

app-admin/etcd-wrapper/files/etcd-wrapper
glevand: OK, I'll look into the details of that.

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
marineam: hm? we kinda know where bash is

marineam:
See https://github.com/coreos/baselayout/pull/50 for your localhost concerns.

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,55 @@
...
+		;;
+esac
+
+source /usr/share/coreos/update.conf
+source /etc/coreos/update.conf
+
+ETCD_VERSION="${ETCD_VERSION:-${GROUP:-stable}}"
+ETCD_ACI="${ETCD_ACI:-coreos.com/etcd}"
+
+case ${ETCD_VERSION} in
```
steveeJ: This should go directly into one of the sources files

marineam: The sourced files are just providing the default etcd "channel" based on the one use for coreos itself, they aren't shell scripts nor should they be have etcd info, they are configuration for update_engine.

Also, there isn't a place where we can set etcd version that will change based on the coreos channel, if it lived in config in /etc it wouldn't update during upgrades. If it lived in /usr it would have to be the same for all channels since /usr does not change between channels. So presuming this is behavior we want it has to be logic in code.

BUT... that is assuming this is something we want in the first place. Now that I think about it I'd argue that this is something we really do NOT want as it is contrary to the existing pattern of versions being promoted verbatim between alpha/beta/stable. Things get surprising when versions change like this, dropping the kublet between beta/stable has been bad enough.

So if we specify a version it should be a single version, but if it is a single version then it doesn't really need to be a container in the first place. Instead can we go with specifying no version like we did for kublet-wrapper? If we want to do alpha/beta/stable channels for etcd they should be version aliases/tags in quay.io or whatever. Doing that also means that etcd updates would not be bound to coreos updates which is probably reasonable if you want to promote etcd channels differently than coreos channels as the following code tries to do.

marineam: If etcd updates are being tied to coreos updates kinda/sorta as are done here to avoid updating etcd by surprise or mixed cluster versions, piggy backing on locksmith/update_engine for coordination, then instead we need to resurrect the rkt container updater project and implement per-container coordination. Either components are in containers and maintained independently of the OS, or the component is simply included in the OS image. kinda sorta doing both means we get to combine the problems of both into one unfortunate pile of sadness.

steveeJ: The updates to etcd, or any other containerized system service, doesn't have to depend on the CoreOS updates. I imagine there to be a default version per CoreOS version, but the user, or the container updater service, should be able to update it independently. In an OOB discussion with @crawford about this, the implementation we had in mind consists of targets for component groups (e.g. kubelet + docker) which uses service templates to adjust the component version. Details will go in a design doc as this is not specific to etcd.

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,29 @@
...
+	--volume etc-ssl-certs,kind=host,source=/etc/ssl/certs \
+	--volume notify,kind=host,source=/run/systemd/notify \
+	--mount volume=etc-ssl-certs,target=/etc/ssl/certs \
+	--mount volume=notify,target=/run/systemd/notify \
+	--trust-keys-from-https \
+	--net=host \
+	--interactive \
+	$RKT_OPTS \
+	--stage1-from-dir=stage1-fly.aci \
+	${ETCD_ACI}:${ETCD_VERSION} --exec=/etcd -- "$@"
```
steveeJ: I would not enforce _:_ to be part of the name as it prevents people from using local files (with sane names).

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,29 @@
...
+# must be set to one of the tags published here:
+#
+#   https://github.com/coreos/etcd/releases
+#
+# If no version is provided, a default version of etcd is used.
+
+set -e
+
+ETCD_VERSION="${ETCD_VERSION:-v2.3.6}"
+ETCD_ACI="${ETCD_ACI:-coreos.com/etcd}"
```
steveeJ: `ETCD_ACI="${ETCD_ACI:-coreos.com/etcd:$ETCD_VERSION}"`, then later on only use `ETCD_ACI`. This allows to supply a local file which could also be prefetched and/or baked into the CoreOS image.

app-admin/etcd-wrapper/files/etcd-wrapper
steveeJ: I'm sure they said that about python at one point ;-)

steveeJ:
Now we could fetch the ACI at build time and store it on disk. this way we don't really on the network and availability of the specific ACI online. is there an eclass for fetching and verifying signed files?

marineam:
@steveeJ that's built into portage, add the URL to SRC_URI

app-admin/etcd-wrapper/files/etcd-wrapper
marineam: we can say that about python too (when applicable) as we are the ones installing things! There is no reason to use env in this context.

philips:
A few questions:

1) Is there a reason why the wrapper has the versions instead of an external environment file?
2) Can we please confirm that the process is running as a non-root user and that /var/lib/etcd or wherever matches that UID?

philips:
Oh, and is there a reason to not run by hash?

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,29 @@
...
+#
+# If no version is provided, a default version of etcd is used.
+
+set -e
+
+ETCD_VERSION="${ETCD_VERSION:-v2.3.6}"
+ETCD_ACI="${ETCD_ACI:-coreos.com/etcd:${ETCD_VERSION}}"
+ETCD_DATA_DIR="${ETCD_DATA_DIR:-/var/lib/etcd}"
+
+exec /usr/bin/rkt run \
```
philips: Need to separate out fetch from run

crawford: Why? Not a rkt power-user yet :(

crawford:
> 1) Is there a reason why the wrapper has the versions instead of an external environment file?

Just hacking this together for now. I was going to pull that version into a separate file once we've determined that this approach will roughly work.

app-admin/etcd-wrapper/files/etcd-wrapper
steveeJ: I'm working on it.

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,29 @@
...
+	--volume data-dir,kind=host,source=${ETCD_DATA_DIR} \
+	--volume etc-ssl-certs,kind=host,source=/etc/ssl/certs \
+	--volume notify,kind=host,source=/run/systemd/notify \
+	--mount volume=etc-ssl-certs,target=/etc/ssl/certs \
+	--mount volume=notify,target=/run/systemd/notify \
+	--trust-keys-from-https \
+	--net=host \
```
steveeJ: this will be ignored by stage1-fly

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,29 @@
...
+	--interactive \
```
steveeJ: this will be ignored by stage1-fly. out of curiosity, why would we want this at all?

crawford: left over from the default stage-1

app-admin/etcd-wrapper/files/etcd-wrapper
crawford: left over from the default stage-1

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -0,0 +1,29 @@
...
+ETCD_ACI="${ETCD_ACI:-coreos.com/etcd:${ETCD_VERSION}}"
+ETCD_DATA_DIR="${ETCD_DATA_DIR:-/var/lib/etcd}"
+
+exec /usr/bin/rkt run \
+	--volume data-dir,kind=host,source=${ETCD_DATA_DIR} \
+	--volume etc-ssl-certs,kind=host,source=/etc/ssl/certs \
+	--volume notify,kind=host,source=/run/systemd/notify \
+	--mount volume=etc-ssl-certs,target=/etc/ssl/certs \
+	--mount volume=notify,target=/run/systemd/notify \
+	--trust-keys-from-https \
```
steveeJ: I don't think this is actually a good idea. If we know all the container image sources that are used within the core system, we should establish their trust in the rkt ebuild's postinstall.

steveeJ:
@philips 

> Can we please confirm that the process is running as a non-root user and that /var/lib/etcd or wherever matches that UID?

Which process exactly do you mean? Do we not trust rkt to do the UID change right?

glevand:
We'll need coreos/rkt#2758 for this to work for arm64.

PR #1999, opened 2016-06-08
profile: begin explicitly masking packages for prod images

marineam:
This serves as a sanity check to avoid mistakes like pulling in python.


crawford:
LGTM

PR #2000, opened 2016-06-10
profiles/sdk: Build unstable qemu

glevand:
qemu-2.6.0 includes a bug fix for arm64.  Without this the 4.6 kernel
will hang early in its boot with qemu-system-aarch64.

Depends on https://github.com/coreos/portage-stable/pull/437

@marineam Is this how we specify an unstable version (~amd64) for the SDK?


marineam:
yup, LGTM

PR #2002, opened 2016-06-13
profiles: accept newer mtools

mischief:


profiles/coreos/base/package.accept_keywords
```diff
@@ -103,3 +103,8 @@ dev-util/checkbashisms
 
 # Must use the same version across all architectures
 =dev-libs/protobuf-2.6.1-r3
+
+# mtools 4.0.18 fixed a bug that prevented 'mformat -i efi.img -v EFI -f 2880'
+# from working for unknown reasons...
```
marineam: I think the reason simply is that the -i flag didn't exist ;-)

marineam:
LGTM

PR #2004, opened 2016-06-13
app-emulation/rkt: v1.8.0

pbx0:


mischief:
any testing done, like a rkt run etcd, coreos+rkt+k8s?

pbx0:
Whats the bar on testing for this? I did not build it and use it within the sdk. It has been used and tested for rktnetes. I've used it a bit on CoreOS but by downloading the github release.

mischief:
ideally rkt built with this ebuild is tested, since that's what we're actually going to ship.

marineam:
Well, of kola gets a test for rktnetes then jenkins will take care of the rest ;-)

marineam:
LGTM

pbx0:
A kola test for rktnetes is certainly on my plate after we are all done with those related releases. I emerged rkt in the SDK and do some basic usage of it. Merging

PR #2009, opened 2016-06-15
flannel: use rkt to launch flannel on amd64

ajeddeloh:
Launch flannel with rkt instead of an early instance of docker and remove early-docker

### DIFF
--- a/app-admin/flannel/flannel-9999.ebuild
+++ b/app-admin/flannel/flannel-9999.ebuild
@@ -4,9 +4,9 @@
 EAPI=4
 
 if [[ "${PV}" == 9999 ]]; then
-	KEYWORDS="~amd64 ~arm64"
+	KEYWORDS="~amd64"
 else
-	KEYWORDS="amd64 arm64"
+	KEYWORDS="amd64"
 fi
 
 inherit systemd
@@ -19,11 +19,11 @@ LICENSE="Apache-2.0"
 SLOT="0"
 IUSE=""
 
-RDEPEND="app-admin/sdnotify-proxy"
+RDEPEND="app-emulation/rkt"
 S="$WORKDIR"
 
 src_install() {
-	sed "s/{{flannel_ver}}/${PV}/" "${FILESDIR}"/flanneld.service >"${T}"/flanneld.service
+	sed "s/{{flannel_ver}}/${PV}/" "${FILESDIR}"/flanneld-rkt.service >"${T}"/flanneld.service
 	systemd_dounit "${T}"/flanneld.service
 
 	insinto /usr/lib/systemd/network
--- a/app-admin/flannel/files/flanneld-rkt.service
+++ b/app-admin/flannel/files/flanneld-rkt.service
@@ -1,0 +1,38 @@
+[Unit]
+Description=Network fabric for containers
+Documentation=https://github.com/coreos/flannel
+After=etcd.service etcd2.service
+
+[Service]
+Type=notify
+Restart=always
+RestartSec=5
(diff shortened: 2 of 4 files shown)


marineam:
Both ebuild revisions need bumping

app-admin/flannel/files/flanneld.service
```diff
@@ -15,28 +13,29 @@ Environment="FLANNEL_VER={{flannel_ver}}"
 Environment="FLANNEL_IMG=quay.io/coreos/flannel"
 Environment="ETCD_SSL_DIR=/etc/ssl/etcd"
 Environment="FLANNEL_ENV_FILE=/run/flannel/options.env"
```
marineam: Is FLANNEL_ENV_FILE still needed?

app-admin/flannel/files/flanneld.service
```diff
@@ -15,28 +13,29 @@ Environment="FLANNEL_VER={{flannel_ver}}"
...
-  --env-file=${FLANNEL_ENV_FILE} \
-  --volume=/usr/share/ca-certificates:/etc/ssl/certs:ro \
-  --volume=${ETCD_SSL_DIR}:${ETCD_SSL_DIR}:ro \
-  ${FLANNEL_IMG}:${FLANNEL_VER} /opt/bin/flanneld --ip-masq=true
+ExecStart=/usr/bin/rkt run --net=host \
+   --stage1-path=/usr/lib/rkt/stage1-images/stage1-fly.aci \
+   --insecure-options=image \
+   --set-env=NOTIFY_SOCKET=/run/systemd/notify \
+   --set-env=AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID} \
+   --set-env=AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY} \
```
marineam: If inherit-env is true why do these vars need to be duplicated?

ajeddeloh: Good point, fixed.

app-admin/flannel/files/flanneld.service
ajeddeloh: Nope, and neither is DOCKER_HOST. Fixed.

glevand:
We'll need https://github.com/coreos/rkt/pull/2758 for this to work for arm64.

marineam:
Looks ok, though a thought so this can go forward without impacting arm while we wait on rkt there: leave early docker in place and change the flannel commit to create a new ebuild reversion and service file, keeping the old available. The new ebuild should only be keyworded on amd64 and use the new rkt service.

ajeddeloh:
Bump, added new ebuild for amd64 only, left early-docker alone

app-admin/flannel/flannel-9999.ebuild
```diff
@@ -23,7 +23,7 @@ RDEPEND="app-admin/sdnotify-proxy"
 S="$WORKDIR"
```
marineam: Remove `RDEPEND="app-admin/sdnotify-proxy"` since it is no longer used. Set to rkt instead.

ajeddeloh: Fixed.

marineam: I don't see it

ajeddeloh: woops, my bad. there now

marineam:
LGTM

glevand:
Seems like that will work.

PR #2011, opened 2016-06-15
sys-auth/polkit: remove gobject-introspection from RDEPEND

marineam:
polkit's gobject library doesn't actually link against the library
provided by gobject-introspection so this appears to be a build-time
only dependency. Fixing this prunes python and lots of other packages:

```diff
--- developer-1068.0.0+2016-06-15-1524-a1/coreos_production_image_packages.txt	2016-06-15 15:27:59.353348324 -0700
+++ developer-1068.0.0+2016-06-15-1609-a1/coreos_production_image_packages.txt	2016-06-15 16:12:36.694523784 -0700
@@ -1,4 +1,3 @@
-app-admin/eselect-1.4.5::portage-stable
 app-admin/flannel-0.5.5-r2::coreos
 app-admin/fleet-0.11.7::coreos
 app-admin/kubelet-1.1.2_p0-r3::coreos
@@ -6,7 +5,6 @@
 app-admin/locksmith-0.3.4-r1::coreos
 app-admin/logrotate-3.9.1-r1::portage-stable
 app-admin/mayday-0.1.0-r1::coreos
-app-admin/python-updater-0.11::portage-stable
 app-admin/sdnotify-proxy-0.1.0-r2::coreos
 app-admin/setools-3.3.8-r7::coreos
 app-admin/sudo-1.8.15-r1::portage-stable
@@ -28,11 +26,9 @@
 app-emulation/rkt-1.8.0::coreos
 app-emulation/xenserver-pv-version-6.2.0::coreos
 app-emulation/xenstore-4.4.0-r1::coreos
-app-eselect/eselect-python-20111108::portage-stable
 app-misc/c_rehash-1.7-r1::portage-stable
 app-misc/ca-certificates-3.19.1::coreos
 app-misc/jq-1.5-r2::portage-stable
-app-misc/mime-types-9::portage-stable
 app-misc/pax-utils-1.0.3::portage-stable
 app-shells/bash-4.2_p53-r1::coreos
 coreos-base/coreos-0.0.1-r268::coreos
@@ -47,8 +43,6 @@
 dev-db/etcd-0.4.9-r2::coreos
 dev-db/etcd-2.3.2::coreos
 dev-db/sqlite-3.8.10.2::portage-stable
-dev-lang/python-2.7.9-r1::portage-stable
-dev-lang/python-exec-2.0.1-r1::portage-stable
 dev-lang/spidermonkey-1.8.5-r4::coreos
 dev-libs/cyrus-sasl-2.1.26-r9::coreos
[...]

PR #2012, opened 2016-06-15
sys-auth/polkit: remove gobject-introspection from RDEPEND for 1068

marineam:
Backport https://github.com/coreos/coreos-overlay/pull/2011 and https://github.com/coreos/coreos-overlay/pull/1999


crawford:
LGTM

PR #2015, opened 2016-06-16
amd64: etcd-wrapper ebuild + script

steveeJ:
Continuing #1997.

- [x] etcdctl wrapper
- [x] systemd service
- [x] use etcd user for stage2 (https://github.com/coreos/rkt/issues/2831)
- [x] test
    - [x] etcd2
    - [x] etcd3
- [x] bump ebuild revisions for coreos and hard-host-depends


coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -69,11 +69,9 @@ RDEPEND=">=sys-apps/baselayout-3.0.0"
 # If protocol 1 is installed it must be configured to provide the default
 # implementation based on whether protocol 2 is enabled or not.
 RDEPEND="${RDEPEND}
-	etcd_protocols_1? (
-		dev-db/etcd:1
-		!etcd_protocols_2? ( dev-db/etcdctl )
-	)
-	etcd_protocols_2? ( dev-db/etcd:2 )
+	dev-db/etcdctl
```
steveeJ: this is needed because etcd-wrapper doesn't provide etcdctl

marineam: No, that ebuild is the etcd0 version of etcdctl, you do not want it. we should remove it at this point.

marineam:
Need to bump ebuild revisions for coreos and hard-host-depends

coreos-base/coreos/coreos-0.0.1.ebuild
steveeJ: I see! We need to define an approach about how we're going to deal with CLI tools that live inside container images. I'll start a discussion in a different place.

dev-db/etcd-wrapped/files/etcd.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=etcd%i-wrapped
+
+[Service]
+Type=notify
+Environment=ETCD_NAME=%m
+Environment=ETCD_IMG_USER=root
```
steveeJ: This can be changed to etcd when https://github.com/coreos/rkt/issues/2831 is dealt with.

dev-db/etcd-wrapped/files/etcd-wrapped.ebuild
```diff
@@ -0,0 +1,97 @@
...
+	SLOT="3"
+	;;
+*)
+	die "Unknown etcd version ${VERSION}"
+	;;
+esac
+
+
+SLOT_MANGLED=${SLOT/0/}
+IMG_FILENAME="etcd-v${VERSION}-linux-${ARCH}.aci"
```
marineam: Since IMG_FILENAME helps populate SRC_URI don't encode arch using the $ARCH variable as that value is set at runtime. SRC_URI will need to specify all architectures and just guard the correct ones with "amd64? ( ... ) arm64? (...)" syntax

dev-db/etcd-wrapped/etcd-wrapped-0.4.9.ebuild
```diff
@@ -0,0 +1 @@
+files/etcd-wrapped.ebuild
```
marineam: Lets not bother 0.4.9. that version is only still around because we don't have an established pattern for deprecating and removing things safely from the image yet. It isn't worth touching with any new development like this.

dev-db/etcd-wrapped/files/etcd-wrapped.ebuild
```diff
@@ -0,0 +1,97 @@
...
+#		--skip-fingerprint-review \
+#		${IMG_PUBKEY} \
+#		|| die "rkt trust failed"
+#
+#	${RKT} fetch \
+#		"${DISTDIR}/${IMG_FILENAME}" \
+#		> fetch.log || die "rkt fetch failed"
+#
+#	echo ${ENV_PREFIX}_IMG_FACTORY_HASH=$(tail -n1 fetch.log) >> "${D}/${IMG_VERSION_FILE}"
+	echo ${ENV_PREFIX}_IMG="/usr/lib/coreos/versions/etcd${SLOT_MANGLED}.d/${IMG_FILENAME}" >> "${D}/${IMG_VERSION_FILE}"
```
marineam: etcd${SLOT}.d would only ever contain a single aci at a time. Why bother with the extra indirection when it could just be etcd${SLOT}.aci in the first place?

dev-db/etcd-wrapped/etcd-wrapped-0.4.9.ebuild
steveeJ: So can I remove all workarounds/efforts for etcd0 from the PR? This also means no more `etcd.service` on CoreOS.

dev-db/etcd-wrapped/files/etcd-wrapped.ebuild
steveeJ: There's the signature and pubkey file next to it too. It doesn't have to be in a directory but it's cleaner.

dev-db/etcd-wrapped/files/etcd-wrapped.ebuild
steveeJ: What exactly do you mean by guard? We need to specify the ARCH in the SRC_URI because we need the appropriate container image.

dev-db/etcd-wrapped/files/etcd-wrapper
```diff
@@ -0,0 +1,60 @@
...
+REQUIRED_ENV_VARS="ETCD_IMG ETCD_IMG_USER ETCD_DATA_DIR"
+for rev in $REQUIRED_ENV_VARS ; do
+    if [[ -z ${!rev} ]]; then
+        echo ${rev} is not set
+        exit 1
+    fi
+done
+
+RKT=${RKT:-/usr/bin/rkt}
+NOTIFY_SOCKET=/run/systemd/notify
```
marineam: Isn't this supposed to be set by systemd? Its pretense in the env is the signal the running process has whether it needs to notify or not.

dev-db/etcd-wrapped/files/etcd-wrapper
```diff
@@ -0,0 +1,60 @@
...
+SSL_CERTS=/etc/ssl/certs
+ETC_HOSTS=/etc/hosts
```
marineam: Why bother with vars?

dev-db/etcd-wrapped/files/etcd-wrapper
```diff
@@ -0,0 +1,60 @@
...
+ETCD_IMG_EXEC=${ETCD_IMG_EXEC:-/etcd}
+RKT_STAGE1_ARG=${RKT_STAGE1_ARG:-"--stage1-from-dir=stage1-fly.aci"}
+SYSTEMD_DIR=/run/systemd/system
```
marineam: ditto, it isn't configurable so the var just obfuscates.

dev-db/etcd-wrapped/files/etcd-wrapped.ebuild
```diff
@@ -0,0 +1,97 @@
...
+inherit systemd
+
+DESCRIPTION="etcd Wrapper"
+HOMEPAGE="https://github.com/coreos/etcd"
+KEYWORDS="amd64 arm64"
+
+LICENSE="Apache-2.0"
+IUSE="etcdctl"
+
+VERSION=${PV}
```
marineam: It is typical ebuild style to use the name "MY_PV" for this

dev-db/etcd-wrapped/files/etcd.service.template
```diff
@@ -0,0 +1,18 @@
...
+EnvironmentFile=/usr/lib/coreos/versions/etcd{{SLOT_MANGLED}}
+EnvironmentFile=-/var/lib/coreos/versions/etcd{{SLOT_MANGLED}}
+ExecStart=/usr/lib/coreos/etcd{{SLOT_MANGLED}}-wrapper
+Restart=always
+RestartSec=10s
+LimitNOFILE=40000
+TimeoutStartSec=0
+
+[Install]
+Alias=etcd{{SLOT_MANGLED}}
```
marineam: The service name already is `etcd${SLOT_MANGLED}.service`

dev-db/etcd-wrapped/files/etcd-wrapper
steveeJ: What's a viable alternative?

dev-db/etcd-wrapped/files/etcd.service.template
steveeJ: Thanks, leftover from an experiment.

dev-db/etcd-wrapped/files/etcd-wrapped.ebuild
marineam: I mean SRC_URI must always have all architectures:

```
SRC_URI="amd64? (
        https://github.com/coreos/etcd/releases/download/v${VERSION}/etcd-v${VERSION}-linux-amd64.aci
        https://github.com/coreos/etcd/releases/download/v${VERSION}/etcd-v${VERSION}-linux-amd64.aci.asc
    )
    arm64? (
        https://github.com/coreos/etcd/releases/download/v${VERSION}/etcd-v${VERSION}-linux-arm64.aci
        https://github.com/coreos/etcd/releases/download/v${VERSION}/etcd-v${VERSION}-linux-arm64.aci.asc
    )
    https://coreos.com/dist/pubkeys/app-signing-pubkey.gpg -> etcd-v${VERSION}_pubkey.gpg"
```

dev-db/etcd-wrapped/etcd-wrapped-0.4.9.ebuild
marineam: We can continue to build/install dev-db/etcd:0 until we finally have the guts to delete it.

dev-db/etcd-wrapped/files/etcd-wrapped.ebuild
```diff
@@ -0,0 +1,97 @@
...
+SRC_URI="\
+	https://github.com/coreos/etcd/releases/download/v${VERSION}/${IMG_FILENAME} \
+	https://github.com/coreos/etcd/releases/download/v${VERSION}/${IMG_FILENAME}.asc \
+	https://coreos.com/dist/pubkeys/app-signing-pubkey.gpg -> ${PUBKEY_FILENAME} \
+"
+
+DEPEND=""
+RDEPEND="\
+	app-emulation/rkt[rkt_stage1_fly]
+	!dev-db/etcd
```
marineam: Be more specific and use `!dev-db/etcd[$SLOT]`

dev-db/etcd-wrapped/files/etcd-wrapper
marineam: They never change so just spell em out.

dev-db/etcd-wrapped/files/etcd-wrapped.ebuild
marineam: k

dev-db/etcd-wrapped/etcd-wrapped-0.4.9.ebuild
steveeJ: For the sake of not having to maintain an old ebuild I would support etcd:0 with this one.

steveeJ:
@marineam 

> Need to bump ebuild revisions for coreos and hard-host-depends

Done. PTAL

coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
```diff
@@ -163,6 +163,11 @@ RDEPEND="${RDEPEND}
 	virtual/cdrtools
 	"
 
+# Host dependencies for pre-populating rkt container images
+RDEPEND="${RDEPEND}
+	app-emulation/rkt
```
steveeJ: Shall we disable all stage1s for the host dependency explicitly? This would shrink the build environment, if that is something we care about.

app-admin/etcd-wrapper/files/etcd-wrapper.ebuild
```diff
@@ -0,0 +1,85 @@
...
+	echo ${ENV_PREFIX}_DATA_DIR=/var/lib/etcd${SLOT} >> "${D}/${IMG_VERSION_FILE}"
+	echo ${ENV_PREFIX}_IMG_PREFIX=${IMG_PREFIX} >> "${D}/${IMG_VERSION_FILE}"
+	echo ${ENV_PREFIX}_IMG_PUBKEY=${IMG_VERSION_DIR}/${PUBKEY_FILENAME} >> "${D}/${IMG_VERSION_FILE}"
+}
+
+src_install() {
+	insinto "${IMG_VERSION_DIR}"
+	doins "${DISTDIR}/${PUBKEY_FILENAME}"
+
+    local wrapper_file="${TMP}/etcd${SLOT}-wrapper"
```
marineam: You have a mix of spaces/tabs. Always use tabs in ebuilds.

steveeJ: I noticed too and fixed it a couple minutes ago. hit refresh please :)

coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -69,11 +69,12 @@ RDEPEND=">=sys-apps/baselayout-3.0.0"
...
 RDEPEND="${RDEPEND}
-	etcd_protocols_1? (
-		dev-db/etcd:1
-		!etcd_protocols_2? ( dev-db/etcdctl )
-	)
-	etcd_protocols_2? ( dev-db/etcd:2 )
+	etcd_protocols_1? ( dev-db/etcd:1 )
+	etcd_protocols_2? (
+        dev-db/etcd:2
+        app-admin/etcd-wrapper:2
```
marineam: etcd-wrapper should be inside a `amd64? (  )` block.

steveeJ: Shouldn't we use profiles for configuring architecture specific packages?

coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
marineam: rkt isn't actually used in the SDK now is it?

app-admin/etcd-wrapper/files/etcd-wrapper.service.template
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=etcd{{SLOT}} (System Application Container)
+Conflicts=etcd{{SLOT}}.service
+
+[Service]
+Type=notify
+Environment=ETCD_NAME=%m
+Environment=ETCD_IMG_USER=etcd
+Environment="SOURCE_FILES={{SOURCE_FILES}}"
```
marineam: How come this is indirect and sourced in the wrapper instead of just using two env files in the service?

coreos-base/coreos/coreos-0.0.1.ebuild
marineam: I don't follow. This ebuild controls what lands in the image and we cannot pull in rkt on other architectures.

steveeJ: Typically [portage profiles](https://wiki.gentoo.org/wiki/Profile_(Portage)) are used for arch specific packages. It looks like the SDK makes no use of profiles at all, so my comment is not really applicable.

app-admin/etcd-wrapper/files/etcd-wrapper.service.template
steveeJ: The wrapper has default values to ensure standalone functionality.

coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
steveeJ: Right, not as it is now. It was still convenient to test the wrappers within the SDK, however that doesn't justify the hard host dependency. Should I remove it?

coreos-base/coreos/coreos-0.0.1.ebuild
marineam: I promise we use extensive use of profiles. But it is up to ebuilds to have correct dependencies for any given arch. You can use a hacky workaround by adding the etcd wrapper to "package.provided" in the arm profile which makes it basically go away but that's not what package.provided is really for and should be used sparingly.

coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
marineam: Lets just remove it for now.

coreos-base/coreos/coreos-0.0.1.ebuild
steveeJ: I would like to pursue the idea to use the @profile set for arch specific packages as an (IMHO more idiomatic and cleaner) alternative to the current approach. As it seems too hasty for this PR I'll add the change as per your suggestion.

marineam:
LGTM

krnowak:
Support for arm64 in rkt just got merged.

steveeJ:
I've to etcd-3.0.0. Are we shipping this?

crawford:
Can you squash a21aa4a and b795e20, and a8ccff2 and a8f1c0f?

steveeJ:
> Can you squash a21aa4a and b795e20, and a8ccff2 and a8f1c0f?

Happened.

crawford:
Well, that's one way to do it. I actually wanted two separate commits (which is why there was a comma), but it's fine.

steveeJ:
> Well, that's one way to do it. I actually wanted two separate commits (which is why there was a comma), but it's fine.

Sorry, failed to parse your syntax there.

crawford:
Fixes https://github.com/coreos/bugs/issues/1303.

crawford:
LGTM

PR #2016, opened 2016-06-17
bootengine: add network kernel modules

marineam:
Update to include https://github.com/coreos/bootengine/pull/90


mischief:
lgtm

PR #2020, opened 2016-06-21
include sys-fs/quota in the image

mischief:
fixes coreos/bugs#1417


marineam:
Is the rpc use flag already set?

mischief:
@marineam it wasn't, fixed it.

marineam:
LGTM

PR #2025, opened 2016-06-23
app-emulation/rkt: check that ebuild and rkt CoreOS PXE versions match

mischief:


steveeJ:
Add a link to https://github.com/coreos/rkt/blob/${PV}/stage1/usr_from_coreos/coreos-common.mk#L14 in the comment for convenience?

marineam:
Maybe even add a test in src_prepare to assert that the version is correct.

mischief:
@marineam ptal

marineam:
LGTM

PR #2031, opened 2016-06-24
coreos-base/coreos: fix revision

marineam:
Commit aeec3f0 accidentally decremented the revision number instead of
incrementing it.


mischief:
lgtm

PR #2033, opened 2016-06-27
Fix segfaults when reading long lines from /etc/gshadow

dm0-:
See coreos/bugs#1394.


crawford:
So this won't allow us to read long lines, we'll just avoid the segfault, right?

dm0-:
No, this patch passes up ERANGE which causes the buffer to be reallocated larger, then it retries parsing the line.  So, it should work as expected.

philips:
@dm0- do you have a link to the upstream submission, just for tracking?

dm0-:
I have sent the patch to the [libc-alpha list](https://sourceware.org/ml/libc-alpha/2016-06/msg01015.html).

philips:
cc @mjg59 as our resident GNU expert in case @dm0- needs help getting upstream

crawford:
LGTM

PR #2035, opened 2016-06-27
coreos-{kernel,sources}: bump to 4.5.7

mischief:
'Fix unallocated memory access in TPM eventlog code' patch is dropped
because it is now included in upstream patches present in 4.5.7.


marineam:
LGTM

PR #2037, opened 2016-06-28
coreos-base/oem-azure: disable buffering on std*

crawford:
From the python manpage:

```
  - u    Force  stdin,  stdout  and  stderr  to  be  totally unbuffered.
         On systems where it matters, also put stdin, stdout and stderr
         in binarymode.
```


mischief:
lgtm

PR #2041, opened 2016-07-05
sys-kernel/coreos-modules: TCM/FS config

mischief:
- enable TCM for torus
- sync some fs options for arm64 and enable scsi cd driver


barakmich:
LGTM; happy to test it if there's a test image being built soon

glevand:
LGTM

PR #2045, opened 2016-07-07
profiles: prune scripts which lack interpreters

mischief:


crawford:
Is `profiles/coreos/targets/generic/prod/make.defaults` a better place for this?

mischief:
you're probably right, in theory these things work fine in the SDK.

mischief:
PTAL

crawford:
LGTM

PR #2047, opened 2016-07-07
arm64 rkt

glevand:
Bumps rkt from 1.9.1 to 1.10.0 and adds arm64 rkt builds.


app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -77,7 +78,7 @@ src_prepare() {
 
 	rkt_coreos_version=$(awk '/^CCN_IMG_RELEASE/ { print $3 }' stage1/usr_from_coreos/coreos-common.mk)
 	if [ "${rkt_coreos_version}" != "${PXE_VERSION}" ]; then
-		die "CoreOS versions in ebuild and rkt build scripts are mismatched expecting ${rkt_stage1_coreos}!"
```
mischief: oops. my fault :p

app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -15,16 +15,18 @@ CROS_WORKON_LOCALNAME="rkt"
...
+COREOS_BOARD="${ARCH}-usr"
+
 PXE_VERSION="1068.0.0"
 PXE_SYSTEMD_VERSION="v229"
-PXE_URI="https://alpha.release.core-os.net/amd64-usr/${PXE_VERSION}/coreos_production_pxe_image.cpio.gz"
-PXE_FILE="${PN}-pxe-${PXE_VERSION}.img"
+PXE_URI="https://alpha.release.core-os.net/${COREOS_BOARD}/${PXE_VERSION}/coreos_production_pxe_image.cpio.gz"
+PXE_FILE="${PN}-pxe-${COREOS_BOARD}-${PXE_VERSION}.img"
 
 SRC_URI="rkt_stage1_coreos? ( $PXE_URI -> $PXE_FILE )"
```
mischief: i think that for SRC_URI, images from both architectures will need to be included and gated on arch.

see https://github.com/coreos/coreos-overlay/blob/master/dev-lang/go-bootstrap/go-bootstrap-1.5.3.ebuild#L10 for an example of this.

glevand: Looking at https://devmanual.gentoo.org/ebuild-writing/variables/index.html#src_uri, it seems that such usage is allowed, but not required.

In testing every thing seemed to work OK, in that the arm64 version was pulled to the rkt-pxe-arm64-usr-1068.0.0.img dist file, and amd64 to rkt-pxe-amd64-usr-1068.0.0.img.

mischief: i believe the reasoning here is that listing both allows the distfiles mirror thing to download all of them. without that, only one of the two images would be in the distfiles mirror.

glevand: We could do something like this, but it is just guessing at how the mirror selects files:

```
PXE_URI_AMD64="https://alpha.release.core-os.net/amd64-usr/${PXE_VERSION}/coreos_production_pxe_image.cpio.gz"
PXE_URI_ARM64="https://alpha.release.core-os.net/arm64-usr/${PXE_VERSION}/coreos_production_pxe_image.cpio.gz"

PXE_FILE_AMD64="${PN}-pxe-amd64-usr-${PXE_VERSION}.img"
PXE_FILE_ARM64="${PN}-pxe-arm64-usr-${PXE_VERSION}.img"

SRC_URI="rkt_stage1_coreos? (
	amd64? ( ${PXE_URI_AMD64} -> ${PXE_FILE_AMD64} )
	arm64? ( ${PXE_URI_ARM64} -> ${PXE_FILE_ARM64} )
)"
```
@marineam What do you think?

mischief: this looks fine to me

marineam: Yes, the `amd64? ( ...) arm64? ( ... )` is the correct way. Runtime vars like $ARCH should never influence SRC_URI in order for the Manifests file and mirroring to work correctly.

glevand:
Rebased to latest, fixed ebuild SRC_URI.

mischief:
lgtm

PR #2048, opened 2016-07-07
sys-auth/sssd: Enable Kerberos support

mjg59:
Add the appropriate flags for building with Kerberos support and bump the version

### DIFF
--- a/sys-auth/sssd/sssd-1.13.1-r1.ebuild
+++ b/sys-auth/sssd/sssd-1.13.1-r1.ebuild
@@ -14,7 +14,7 @@ SRC_URI="http://fedorahosted.org/released/${PN}/${P}.tar.gz"
 
 LICENSE="GPL-3"
 SLOT="0"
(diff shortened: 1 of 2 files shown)


mjg59:
Should fix 1443 and partially fix 968

sys-auth/sssd/sssd-1.13.1-r1.ebuild
```diff
@@ -14,7 +14,7 @@ SRC_URI="http://fedorahosted.org/released/${PN}/${P}.tar.gz"
 
 LICENSE="GPL-3"
 SLOT="0"
-KEYWORDS="~amd64 ~arm ~hppa ~ppc ~ppc64 ~x86"
```
mischief: how did this work before if it was ~amd64?

also, this should be 'arm64', arm is 32 bit.

mjg59: Great question, and the answer is I have no idea.

mischief:
lgtm

PR #2050, opened 2016-07-07
bump packages for arm64

glevand:
In an effort to clean up the arm64 package.accept_keywords, bump a number of packages that now have updated arm64 support.

See also: https://github.com/coreos/portage-stable/pull/458


mischief:
lgtm

PR #2051, opened 2016-07-08
profiles/coreos: Turn on Kerberos in ssh

mjg59:
Enable Kerberos support in openssh, since we're already shipping the libraries


mischief:
Lgtm

PR #2052, opened 2016-07-08
Fix equery for e2fsprogs

glevand:
Revert the removal of metadata.xml from commit
cc7f579db2bb8e12011b884623689922bf245998.

Fixes equery errors like these:

  !!! Package sys-fs/e2fsprogs-1.42.13-r1 is missing metadata.xml

cc: @ajeddeloh


crawford:
LGTM

PR #2056, opened 2016-07-08
sys-kernel/coreos-modules: include PATA_JMICRON

crawford:
Fixes https://github.com/coreos/bugs/issues/1449.


mischief:
bump coreos-kernel to match coreos-modules (uses ${PVR})

mischief:
lgtm

PR #2058, opened 2016-07-10
one weird trick to reduce image size

mischief:
before:

```
$ equery-amd64-usr s coreos-kernel
 * sys-kernel/coreos-kernel-4.6.3-r2
         Total files : 9
         Total size  : 34.84 MiB
$ equery-amd64-usr s coreos-modules
 * sys-kernel/coreos-modules-4.6.3-r2
         Total files : 9504
         Total size  : 72.38 MiB

-rw-r--r-- 1 mischief mischief 894M Jul  8 16:49 coreos_production_qemu_image.img
-rw-r--r-- 1 mischief mischief 226M Jul  8 13:47 coreos_production_pxe_image.cpio.gz
-rwxr-xr-x 1 mischief mischief  33M Jul  8 13:47 coreos_production_pxe.vmlinuz
```

after:

```
$ equery-amd64-usr s coreos-kernel   
 * sys-kernel/coreos-kernel-4.6.3-r2
         Total files : 9
         Total size  : 21.16 MiB
$ equery-amd64-usr s coreos-modules
 * sys-kernel/coreos-modules-4.6.3-r2
         Total files : 9492
         Total size  : 39.70 MiB

-rw-r--r-- 1 mischief mischief 673M Jul  9 19:48 coreos_production_qemu_image.img
-rw-r--r-- 1 mischief mischief 211M Jul  9 19:47 coreos_production_pxe_image.cpio.gz
-rwxr-xr-x 1 mischief mischief  19M Jul  9 19:47 coreos_production_pxe.vmlinuz
```


mischief:
this results in an unbootable system right now, as `/usr/lib/modules` appears to be empty in `bootengine.cpio`.

marineam:
I'm guessing dracut needs an update but that'll be a bit of work. Do the modules now have a .ko.xz extension instead of just .ko? Also is the kmod package up to date?

Also, how much of your space savings is actually due to the fact that you've removed the modules? ;-)

mischief:
our kmod is 21, upstream is stable at 22.

the extensions are .ko.xz now.

i'll have to check the sizes out after fixing the cpio :(

euank:
It might be worth checking the change in boot-time in cpu-limited environments. For example, on AWS the t2.* instances will not gain much from the change in size because the AMI is already quite near, but CPU is at a premium.

PR #2067, opened 2016-07-12
profile/arm64: libpcre GLSA fix

glevand:


mischief:
you can use 8.38-r1 to match amd64.

glevand:
Changed from libpcre-8.39 to libpcre-8.38-r1.

mischief:
lgtm

PR #2068, opened 2016-07-12
app-admin/etcd-wrapper: remove slotting and complexity

steveeJ:
Passing arguments via this environment variable is more convenient than e.g.
overriding the ExecStart line in a drop-in.


crawford:
I think we use _OPTS elsewhere. We should probably standardize on one name.

steveeJ:
@crawford +1 on standardizing, you know more about the projects than I do so just tell me what it's going to be and I'll adapt the PR.

crawford:
Let's go with ETCD_OPTS

crawford:
LGTM

steveeJ:
@crawford I reworked this one and checked it on a new build. After logging in per SSH, one needs to do the following to be able to use it:

1. write `ETCD_TAG=v2.3.7` to /var/lib/coreos/versions/etcd
2. `sudo systemctl start etcd-wrapper`

`ETCD_TAG` could also be provided as a dropin `Environment=` for the service file. I think this is a good balance of convenience and responsibility for the user.

/cc @philips

crawford:
> 1. write ETCD_TAG=v2.3.7 to /var/lib/coreos/versions/etcd

I don't like this. Why even provide the option? I think the `Environment=` is the correct mechanism.

app-admin/etcd-wrapper/etcd-wrapper-0.0.1-r0.ebuild
```diff
@@ -0,0 +1,27 @@
...
+SLOT=0
+
+DEPEND=""
+RDEPEND=">=app-emulation/rkt-1.9.1[rkt_stage1_fly]"
+
+S=${WORKDIR}
+
+src_install() {
+    dobin "${FILESDIR}"/etcd-wrapper
+  	systemd_dounit "${FILESDIR}"/etcd-wrapper.service
```
mischief: this indentation seems strange, at least in github ui.

steveeJ:
@crawford 

> I don't like this. Why even provide the option? I think the Environment= is the correct mechanism.

You can do either, or both even if you want. The `EnvironmentFile` path starts with a `-` declaring it non-critical. You can easily just use a drop-in that containts `Environment=ETCD_TAG=v2.3.7`.

app-admin/etcd-wrapper/etcd-wrapper-0.0.1-r0.ebuild
steveeJ: Fixed, thanks

crawford:
> The EnvironmentFile path...

But why is this here at all? I think we should remove this since we are no longer providing a default.

app-admin/etcd-wrapper/files/etcd-wrapper.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=etcd (System Application Container)
+Conflicts=etcd.service
+Conflicts=etcd2.service
+
+[Service]
+Type=notify
```
crawford: Let's add an empty `Environment=ETCD_TAG=` so it's obvious that it needs to be overridden.

steveeJ:
> But why is this here at all? I think we should remove this since we are no longer providing a default.

It's so people don't have to necessarily create a drop-in. Placing an environment file is easier as it doesn't require a `daemon-reload`.

app-admin/etcd-wrapper/files/etcd-wrapper.service
steveeJ: ACK

crawford:
> Placing an environment file is easier as it doesn't require a daemon-reload.

Using EnvironmentFile is frowned upon in the systemd community. Having multiple ways to configure this is just going to confuse people. Please remove.

steveeJ:
> Using EnvironmentFile is frowned upon in the systemd community. Having multiple ways to configure this is just going to confuse people. Please remove.

Frankly, I don't think that the mere fact that something is frowned upon is a valid argument nor is it an explanation at all. I don't see why it would be bad to use that option, but I'll remove the line for the sake of moving forward.
Sorry folks, I've tried to save your systemd from an unnecessary reload!

crawford:
LGTM

PR #2074, opened 2016-07-13
libgpg-error error fix

glevand:
Import latest from portage stable, add cross compile fix.


mischief:
can you prune unused ebuilds from 951d805?

glevand:
Removed unused ebuilds.

mischief:
@glevand yeah, fixup 9f61e4a into 2613c6e, so you're just importing the ebuild we use. other than that, lgtm! thanks!

PR #2083, opened 2016-07-18
app-admin/etcd-wrapper: Added missing newline escape

clbh:
This was preventing arguments supplied to etcd-wrapper from being
passed on to etcd, as well as preventing specifying a custom exec
command for the ACI


vcaputo:
oof, lgtm

PR #2084, opened 2016-07-19
app-shells/bash: sync bashrc with upstream

dm0-:
This fixes coreos/bugs#1464.  The upstream Gentoo `bashrc` just lets `dircolors` handle parsing the database instead of trying to do it itself.

This also syncs `~/.bash_profile` with Gentoo to fix a minor cosmetic issue.  With the previous conditional command list, it was possible to always start your session an error exit code.  An `if` block is always successful, which is nicer to people with `$?` in their `PS1`.


mischief:
lgtm

PR #2086, opened 2016-07-20
coreos-base/coreos: bump revision

marineam:
Commit bcfd9db37 failed to update the revision when modifying the
ebuild, breaking build_image with emerge: there are no binary packages
to satisfy "app-admin/etcd-wrapper:3" errors.


mischief:
Lgtm

steveeJ:
Ah my bad, apologies!

PR #2087, opened 2016-07-21
app-emulation/wa-linux-agent: enable base64 decoding of userdata

mischief:
fixes coreos/bugs#1463


vcaputo:
lgtm

crawford:
Verified the fix. LGTM.

@szarkos refresh my memory, has the wire endpoint always returned CustomData encoded as Base64?

szarkos:
Yes, it needs to be base64-encoded when sent to our APIs, and it's delivered to the VM as-is with the other configuration data in the ovf-env.xml file.

crawford:
@szarkos gotcha. So the xplat-cli is doing the encoding for us.

PR #2088, opened 2016-07-21
Extend selinux pipefs permissions

mjg59:
We need ioctl() and getattr() on these


vcaputo:
lgtm, too bad about the kitchen sink syscall being needed.

PR #2094, opened 2016-07-24
app-admin/*-wrapper: introduce consistency

steveeJ:
- [x] decide on etcd default version @philips 
- [ ] decide whether to include _kubelet.service_ @aaronlevy


crawford:
Can you double check to make sure that the names between kubelet and etcd wrappers are the same? (e.g. ETCD_TAG vs KUBELET_VERSION) We want to make sure that those interfaces are consistent.

steveeJ:
@crawford working on it locally, also looking at the flannel wrapper.

app-admin/etcd-wrapper/files/etcd-member.service
```diff
@@ -8,7 +8,7 @@ Type=notify
 Environment=ETCD_NAME=%m
 Environment=ETCD_IMG_USER=etcd
 Environment=ETCD_DATA_DIR=/var/lib/etcd
-Environment=ETCD_TAG=
+Environment=ETCD_TAG=v2.3.8
```
steveeJ: @philips do you suggest to use etcd v3 here?

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -7,12 +7,36 @@
...
+    echo One of $@ must be set
     exit 1
-fi
+}
+
+KUBELET_VERSION="${KUBELET_VERSION:-${KUBELET_TAG}}"
+KUBELET_IMG="${KUBELET_ACI}"
 
-KUBELET_ACI="${KUBELET_ACI:-quay.io/coreos/hyperkube}"
+require_ev_one KUBELET_IMG KUBELET_TAG
```
aaronlevy: What about the current case that an existing user has only specified `KUBELET_VERSION`? Requiring `KUBELET_TAG` would seem backward incompatible.

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -7,12 +7,36 @@
...
+
+if [[ ! -z ${KUBELET_TAG} ]]; then
+    KUBELET_IMG="quay.io/coreos/hypberkube:${KUBELET_TAG}"
```
aaronlevy: We can't assume people always want to use our quay images.

app-admin/kubelet-wrapper/files/kubelet-wrapper
marineam: Yes, please don't change the variable names. For consistency sake etcd should be updated to match kublet since kublet was first, not the other way around. @philips requested that change at a recent os meeting.

steveeJ: @marineam: @aaron, @philips and I went through this discussion too and agreed on _\_TAG_ because it's a commonly understood term for image discovery and a better match than version. The requirement was to make it compatible with _\_VERSION which is fulfilled. I had a bug in the previous version of the PR but it's now updated.

app-admin/kubelet-wrapper/files/kubelet-wrapper
steveeJ: Good point, I already made the URL adjustable for flannel and etcd, now did the same for kubelet.

app-admin/kubelet-wrapper/files/kubelet-wrapper
marineam: Ah, ok. wfm

app-admin/etcd-wrapper/files/etcd-member.service
```diff
@@ -0,0 +1,23 @@
...
+[Service]
+Type=notify
+Restart=always
+RestartSec=10s
+LimitNOFILE=40000
+TimeoutStartSec=60
+Environment=ETCD_NAME=%m
+Environment=ETCD_IMG_USER=etcd
+Environment=ETCD_DATA_DIR=/var/lib/etcd
+Environment=ETCD_TAG=v2.3.7
```
marineam: If we are going to go with a default version the ebuild version should be updated to match it.

app-admin/flannel-wrapper/files/flannel.service
```diff
@@ -0,0 +1,24 @@
...
+Restart=always
+RestartSec=10
+LimitNOFILE=40000
+LimitNPROC=1048576
+TimeoutStartSec=60
+ExecStartPre=/sbin/modprobe ip_tables
+ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/lib/coreos/flannel-wrapper1.uuid
+ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/lib/coreos/flannel-wrapper2.uuid
+ExecStartPre=/usr/bin/mkdir -p /var/lib/coreos
+Environment=FLANNEL_TAG=0.5.5
```
marineam: ditto, ebuild should be 0.5.5

mischief: in that case can we just use ${PV}?

marineam: via sed but yeah, that'd be reasonable.

app-admin/etcd-wrapper/files/etcd-member.service
steveeJ: It's a tag rather than a version. I _can_ include a version but it does not have to follow the Gentoo versioning scheme. I suggest to keep it simple and bump the revision when the TAG is changed.

app-admin/flannel-wrapper/files/flannel.service
steveeJ: It's a tag rather than a version. I can include a version but it does not have to follow the Gentoo versioning scheme. I suggest to keep it simple and bump the revision when the TAG is changed.

marineam: Except they are versions that are compatible with gentoo because we try to stick to semver as much as possible for our projects. But doesn't really matter so it can stay as is.

steveeJ:
I've been contemplating if we want to add a _kubelet.service_, but a discussion with @aaronlevy reduced my motivation.

I think this is good to go as is.

app-admin/etcd-wrapper/files/etcd-member.service
steveeJ: @philips would you prefer to have etcd v3 as default version here to spread that faster?

steveeJ: @philips ping

philips:
Whatever is the latest release for etcd should be default and we should bump the ebuild to make this so. I have not heard a convincing argument for why we wouldn't provide a default version that doesn't simply shift an initially arbitrary decision from us to the user.

We should document that you SHOULD set a version if you want maximum stability in a production environment. But, for the default it should be whatever is here: https://github.com/coreos/etcd/releases

app-admin/etcd-wrapper/files/etcd-member.service
philips: v3 as the default, start with v3.0.7

steveeJ:
> Can you double check to make sure that the names between kubelet and etcd wrappers are the same? (e.g. ETCD_TAG vs KUBELET_VERSION) We want to make sure that those interfaces are consistent.

For consistency this is now called $PRODUCT_IMG_TAG. for a lack of documentation so far it's safe to rename at this point. we should start documenting the approach after this PR has been merged.

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -20,11 +20,12 @@ function require_ev_one() {
     exit 1
 }
 
-require_ev_one ETCD_IMG ETCD_TAG
+require_ev_one ETCD_IMG ETCD_IMG_TAG
```
crawford: This change is going to break people's clusters when they update. Is that okay? (no idea if anyone is actually using this)

app-admin/etcd-wrapper/etcd-wrapper-0.0.3.ebuild
```diff
@@ -22,6 +22,8 @@ RDEPEND=">=app-emulation/rkt-1.9.1[rkt_stage1_fly]"
 S=${WORKDIR}
 
 src_install() {
-	dobin "${FILESDIR}"/etcd-wrapper
-  	systemd_dounit "${FILESDIR}"/etcd-wrapper.service
+	exeinto /usr/lib/coreos
+	doexe "${FILESDIR}"/etcd-wrapper
+
+  	systemd_dounit "${FILESDIR}"/etcd-member.service
```
crawford: Is "member" the correct term? /cc @xiang90

crawford:
What's going on with the flannel ebuilds?

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,22 +4,56 @@
...
     exit 1
+}
+
+if [[ ! -z ${KUBELET_VERSION} ]]; then
+    echo KUBELET_VERSION environment variale is deprecated, please use KUBELET_IMG_TAG instead
 fi
 
-KUBELET_ACI="${KUBELET_ACI:-quay.io/coreos/hyperkube}"
+KUBELET_IMG_TAG="${KUBELET_IMG_TAG:-${KUBELET_VERSION}}"
+KUBELET_IMG="${KUBELET_ACI}"
```
crawford: Isn't `KUBELET_IMG` supposed to be exposed to the user?

app-admin/etcd-wrapper/etcd-wrapper-0.0.3.ebuild
steveeJ: /cc @philips

app-admin/etcd-wrapper/files/etcd-wrapper
steveeJ: We haven't written any documentation or public announcements of etcd-wrapper. I dare to say it's OK to rename it at this point.

app-admin/kubelet-wrapper/files/kubelet-wrapper
steveeJ: That's right, it deprecates `KUBELET_ACI` which is kept only for backwards compatibility. I'll add a deprecation warning here too.

app-emulation/docker/docker-1.11.2-r1.ebuild
```diff
@@ -0,0 +1 @@
+docker-9999.ebuild
```
steveeJ: @coreos/team-os I don't know why I'm adding this file, it must've slipped in when rebasing/merging. Should I remove this file?

steveeJ:
> What's going on with the flannel ebuilds?

Can you elaborate?

app-admin/etcd-wrapper/etcd-wrapper-0.0.3.ebuild
xiang90: i am not sure why we do add the member suffix. in etcd we call each logic node at etcd application level member. we call raft logical node peer.

crawford:
@steveeJ just curious why we are creating a second flannel ebuild. Why not just remove the first altogether or rewrite the existing ebuild?

app-emulation/docker/docker-1.11.2-r1.ebuild
crawford: Yes

app-admin/etcd-wrapper/etcd-wrapper-0.0.3.ebuild
crawford: We can't call it `etcd.service` because that already exists (as much as I would have loved to replace it). What would you prefer we call it?

app-admin/etcd-wrapper/files/etcd-wrapper
crawford: I kept this from rolling to Stable so that we could break it. I'm fine with the change.

app-emulation/docker/docker-1.11.2-r1.ebuild
crawford: Oh, I see. It was added because you are changing `docker.service`. You will need to bump the docker version (I think it's 1.12.1 at this point) to pull in the changes.

app-admin/kubelet-wrapper/files/kubelet-wrapper
aaronlevy: But now KUBELET_IMG is always overridden? Should we instead have:

```
KUBELET_IMG="${KUBELET_IMG:-${KUBELET_ACI}}"
KUBELET_IMG="${KUBELET_IMG:-quay.io/coreos/hyperkube}"
```

aaronlevy: nevermind, I see that this is handled below.

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,22 +4,60 @@
...
+    echo KUBELET_ACI environment variable is deprecated, please use the KUBELET_IMG instead
 fi
 
-KUBELET_ACI="${KUBELET_ACI:-quay.io/coreos/hyperkube}"
+KUBELET_IMG_TAG="${KUBELET_IMG_TAG:-${KUBELET_VERSION}}"
+KUBELET_IMG="${KUBELET_ACI}"
+
+require_ev_one KUBELET_IMG KUBELET_IMG_TAG
+
+if [[ ! -z ${KUBELET_IMG_TAG} ]]; then
```
aaronlevy: We will never have a case where KUBELET_IMG_TAG (version) is empty

aaronlevy: Hmm. I'm misinterpreting a lot here. So KUBELET_IMG is actually "repo:tag".

Why are we conflating so many things (img, img_url, tag). Is there a reason we can't just go with something like:

```
KUBELET_REPO
KUBELET_TAG
```

ignore me if this has already been discussed.

aaronlevy:
As a general "fwiw" I'd prefer not to ship a kubelet.service right now.

Not shipping one hasn't hurt us at all, and leaves us a lot of flexibility without having to worry about the underlying OS version of the service file. And if it ever does prove to be an issue, we will just have to override it wholesale anyway -- which negates any benefit. But I'm open to other opinions.

steveeJ:
@crawford 

> just curious why we are creating a second flannel ebuild. Why not just remove the first altogether or rewrite the existing ebuild?

My intention is to keep a distinction between ebuilds that are able to build the software and the ones that rely on a the artifact/image already being available somewhere. There is a chance that we use ebuilds to build the container images, for which we might use the build-ebuilds at one point. For correctness we could introduce virtual packages and then be able to satisfy them with either the building ebuilds or the wrapper ebuilds.

@aaronlevy 
> As a general "fwiw" I'd prefer not to ship a kubelet.service right now.

I have no objection to that.

crawford:
Superseded by https://github.com/coreos/coreos-overlay/pull/2205.

PR #2106, opened 2016-07-29
Upgrade to systemd v231

dm0-:
This fixes coreos/bugs#1480.


dm0-:
When coreos/bootengine#94 looks okay, I will amend this pull request to update the bootengine ebuild with the corresponding merge commit.

iaguis:
Maybe we can backport a (future) fix for https://github.com/systemd/systemd/issues/3851. See https://github.com/coreos/rkt/issues/3013 for context.

sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -24,11 +24,11 @@ inherit autotools bash-completion-r1 linux-info multilib \
 	multilib-minimal pam systemd toolchain-funcs udev user
 
 DESCRIPTION="System and service manager for Linux"
-HOMEPAGE="http://www.freedesktop.org/wiki/Software/systemd"
+HOMEPAGE="https://www.freedesktop.org/wiki/Software/systemd"
 
 LICENSE="GPL-2 LGPL-2.1 MIT public-domain"
 SLOT="0/2"
-IUSE="acl apparmor audit cryptsetup curl elfutils gcrypt gnuefi http
+IUSE="acl apparmor audit cryptsetup curl elfutils +gcrypt gnuefi http
```
mischief: is there any reason to turn on gcrypt here?

dm0-: This was just to bring us closer to upstream's build, not for any specific issues.  We're installing libgcrypt, and enabling it in systemd just seems to provide better security for the  journal and dnssec.  It can be dropped if we don't want it.

vcaputo:
lgtm

sys-apps/systemd/systemd-9999.ebuild
mischief: seems OK to keep it, but make sure it is called out in the release notes.

vcaputo:
lgtm

sys-apps/systemd/systemd-9999.ebuild
vcaputo: hmmm, would be more prudent to keep gcrypt off.

vcaputo:
capturing OOB discussion: keep gcrypt off, we can always turn it on if users request it in the future.

vcaputo:
lgtm lgtm lgtm

marineam:
LGTM

dongsupark:
FWIW while playing with CoreOS 1122 + systemd v231, I met an issue with systemd-resolved that particular hostnames like `storage.googleapis.com` were not resolved. That looks like an already known bug. See https://github.com/systemd/systemd/issues/3826

PR #2109, opened 2016-08-04
profiles/coreos: accept net-misc/curl-7.50.1

mischief:
depends on https://github.com/coreos/portage-stable/pull/466


marineam:
LGTM, and fyi, we don't need to always have ~amd64 ~arm64 on these things, ~arch is always implied if none is specified.

PR #2112, opened 2016-08-05
arm64 kconfig sync

mischief:
@marineam what were the other modules groups you mentioned should be sync'd?


marineam:
Filesystems, dm/md, crypto, and so on. Just diff the two configs and skim for options that aren't arch specific. Even USB devices could get coppied since those aren't tied to the specific platform.

marineam:
In the end the two configs should be pretty similar

damoon:
maybe https://github.com/coreos/bugs/issues/1500 (enable CONFIG_DNS_RESOLVER=y) could be synced too?

glevand:
What you have so far looks fine.  My original file was based on the upstream arm64 defconfig.

marineam:
LGTM, we can do more in another PR

PR #2113, opened 2016-08-05
app-emulation/docker: use default cgroupdriver

crawford:
Fixes https://github.com/coreos/bugs/issues/1435.

/cc @aaronlevy @pbx0


pbx0:
Are you leaving `$DOCKER_CGROUPS ` in the `ExecStart` line for compatibility reasons? I imagine most people override that line rather then the `Environment` if they were customizing this setting.  

Also, there are only two settings for this option anyway. So everyone is either moving from the systemd driver to cgroupfs or already on cgroupfs. Unless there exist configs that pin their daemon to the  systemd driver by overriding the environment.

crawford:
Good point. @marineam do you think we should pull that option entirely? Anybody who overrode it to cgroupfs isn't going to be affected and I doubt anyone overrode the flag with the existing value (these cases would break).

mischief:
:ship: :it: ?

crawford:
@vcaputo and I talked it over and think it's good as is. We can deprecate the `DOCKER_CGROUPS` variable, but should probably leave it for now.

vcaputo:
lgtm

PR #2115, opened 2016-08-08
eclass: omit dwarf symbol table from go binaries

crawford:
This saves a couple kilobytes from each binary. Not much, but we'll take
it.

### DIFF
--- a/eclass/coreos-go.eclass
+++ b/eclass/coreos-go.eclass
@@ -43,7 +43,7 @@ go_build() {
 
 	go build -v \
 		-p "$(makeopts_jobs)" \
-		-ldflags "${GO_LDFLAGS} -extldflags '${LDFLAGS}'" \
+		-ldflags "${GO_LDFLAGS} -w -extldflags '${LDFLAGS}'" \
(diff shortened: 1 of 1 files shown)


vcaputo:
Do we even have enough go binaries for this to really matter?  Any impact on native golang debugging?

crawford:
@vcaputo I did some quick investigation this morning and found that `strip` is doing a pretty good job (enough so that `-s` to the linker doesn't do anything for most of our binaries). This new flag did remove a little bit though (I guess the splitdebug feature misses some of this).

This doesn't effect the stack traces that Go outputs and using real debugging tools with Go is a joke. I don't think we lose anything by tossing DWARF.

marineam:
It is useful to be able to throw gdb at go binaries from time to time when something is segfaulting. So we do still want the split debug info to be intact. I have no idea if this option impacts that or not. I vote not bothering, this really isn't a meaningful contribution to image size.

marineam:
There are meaningful cleanups we can do in the image, lets spend our time elsewhere.

PR #2118, opened 2016-08-09
update_engine: fix skipping unknown update procedures

marineam:
https://github.com/coreos/update_engine/pull/114


crawford:
LGTM

PR #2119, opened 2016-08-09
Add TPM policy installer

mjg59:
Add a small tool for injecting TPM policy into Kubernetes

### DIFF
--- a/app-crypt/tpmpolicy/files/policy_install
+++ b/app-crypt/tpmpolicy/files/policy_install
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+POLICYNAME=$1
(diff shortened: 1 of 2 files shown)


crawford:
This seems awfully specific to Tectonic. I have no idea what program is fielding the HTTP requests, nor do I like the use of 8080. Is this needed within the OS, or can this be installed as part of provisioning (e.g. via Ignition or a daemon-set)?

mjg59:
It's Kubernetes specific rather than Tectonic specific (8080 is the local listener), but yes. Providing this via some other mechanism seems reasonable.

crawford:
Okay, I'm going to close this one out.

PR #2120, opened 2016-08-09
sys-auth/sssd: Enable SSH helpers

mjg59:
Build the SSH-related helpers for sssd so authorized_keys can be stored
in LDAP.


marineam:
Why change the ebuild instead of toggeling the use flag?

mjg59:
Good question. Let's do it this way.

marineam:
LGTM

PR #2123, opened 2016-08-11
dev-lang/go: bump to 1.6.3

mischief:
apparently required by docker 1.12.


crawford:
Let's do it. Get some testing in.

PR #2127, opened 2016-08-11
coreos-devel/sdk-depends: pull app-emulation/actool into sdk

vcaputo:
actool is simpler than acbuild for scripted creation of ACIs from
a rootfs and manifest.


crawford:
LGTM

PR #2129, opened 2016-08-12
sys-apps/systemd: revert to the previous /etc/resolv.conf

dm0-:
This is a workaround for systemd/systemd#3826 until it has a proper fix.


marineam:
LGTM

PR #2131, opened 2016-08-15
eclass/coreos-kernel.eclass: Add support for shared kernel configuration

mjg59:
Add support for sharing configuration between architectures, and break the
common configuration out from our existing architectures.


mischief:
this will beautifully break https://github.com/coreos/scripts/blob/master/kernel_menuconfig

mischief:
needs more cleanup but we can do it later.

lgtm

glevand:
Seems like a lot of the difference between the arch's is ```=m``` vs. ```=y```.

PR #2133, opened 2016-08-15
coreos-base/oem-qemu: add qemu oem definition

crawford:
This is needed so that CoreOS identifies that it is running on QEMU.

### DIFF
--- a/coreos-base/oem-qemu/files/grub.cfg
+++ b/coreos-base/oem-qemu/files/grub.cfg
@@ -1,0 +1,3 @@
+# CoreOS GRUB settings
+
+set oem_id="qemu"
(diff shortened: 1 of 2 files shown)


crawford:
Another option would be to use `systemd-detect-virt` in a generator. Not sure if that makes much sense though...

marineam:
Is this for feeding fw_cfg into ignition? The logic could actually be in grub, and if it there is an ignition config in fw_cfg it adds the oem line to kernel args. I want grub to be able to pull arbitrary kernel args from fw_cfg anyway.

crawford:
Yes, this is for Ignition.

marineam:
also, considering kola qemu uses the base system image instead of the qcow one from image_to_vm it'd be nice to have fw_cfg work without a special oem package.

crawford:
How should we proceed on this one? Use `systemd-detect-virt`?

crawford:
Bump

marineam:
Yes, please use systemd-detect-virt or similar early detection so the base image works, after all that's what kola+qemu uses.

PR #2139, opened 2016-08-18
Revert "app-emulation/rkt: Remove go environment settings"

mjg59:
After this we're no longer passing -nopie to the linker, rkt is
ending up with a TEXTREL section and SELinux is blocking execution.
Revert until it's fixed.

### DIFF
--- a/app-emulation/rkt/rkt-9999.ebuild
+++ b/app-emulation/rkt/rkt-9999.ebuild
@@ -119,6 +119,12 @@ src_configure() {
 		append-ldflags -nopie
 	fi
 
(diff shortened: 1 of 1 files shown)


marineam:
LGTM

marineam:
Oh! missed this earlier, but please bump the revision of the rkt ebuild to ensure people recompile it.

mischief:
@marineam will that be necessary with https://github.com/coreos/coreos-overlay/pull/2140 ?

mjg59:
Should be fine with #2140, but we'll have to bump it if we backport this

PR #2141, opened 2016-08-19
sys-apps/ignition: GO_LDFLAGS: use = in -X assignment

vcaputo:
Silences complaint from go build:
link: warning: option -X github.com/coreos/ignition/internal/version.Raw v0.9.2-4-gdfe9e8b-dirty may not work in future releases;
use -X github.com/coreos/ignition/internal/version.Raw=v0.9.2-4-gdfe9e8b-dirty


dm0-:
LGTM

PR #2143, opened 2016-08-22
coreos-base: add dep on app-admin/sdnotify-proxy

vcaputo:
Flannel lost its dependency on sdnotify-proxy in moving to rkt, but
external users may still depend on sdnotify-proxy being in the image, so
we need to keep it around.

Fixes https://github.com/coreos/bugs/issues/1528


crawford:
LGTM

PR #2145, opened 2016-08-23
grub: improve validation of GPT and prevent disk corruption

marineam:
Systems that previously suffered from disk corruption during upgrades
may still be unable to boot properly but at least the chance of causing
corruption is reduced. Workarounds for these systems are still in
progress.


dm0-:
LGTM

PR #2149, opened 2016-08-25
app-emulation/runc: patch out installing deps in makefile

mischief:
runc invokes go build -i, which installs dependencies to GOROOT. the
build is run without privilege to write there, so it fails.


crawford:
Why on Earth would the default make target do an installation?!? LGTM

PR #2153, opened 2016-08-29
app-crypt/gnupg: move from portage-stable and fix arm64 cross build.

garygrebus:
Fix gnupg-pth.m4 to look in the right place for pth-config on cross builds,
which then requires the ebuild to run autoreconf.

Signed-off-by: gary.grebus <gary.grebus@verizon.com>


mischief:
could you please split the commits? first, the vanilla import of gnupg, and then the addition of the patch.

glevand:
Also, please add the error message to the commit log.  See: https://github.com/coreos/coreos-overlay/commit/19fd7920be6ba55b998cdab3547fc3166039f131

garygrebus:
Updated the branch to make the change in 2 commits.

glevand:
LGTM

mischief:
@garygrebus last nit - can you regenerate Manifest? `ebuild-amd64-usr ./gnupg-2.0.26-r4.ebuild manifest`. after that lgtm.

garygrebus:
I ran the command, but the resulting Manifest only contains the hashes for one file:

DIST gnupg-2.0.26.tar.bz2 4303384 SHA256 7758e30dc382ae7a7167ed41b7f936aa50af5ea
2d6fccdef663b5b750b65b8e0 SHA512 5dd23baaac764fd48abd23 etc.

No hashes for the .ebuild file or the patches under files/.
That didn't seem right to me.  Is that the expected behavior?

mischief:
@garygrebus yes, that's correct, because now the hashes of the other files are effectively inside git - but the source tarball is not, so the hash is needed to verify it when it is downloaded.

garygrebus:
Makes sense.  The Gentoo stuff is all a bit foreign to me.  Updated the last commit to add the Manifest change.

marineam:
This broke the arm build because `dev-libs/pth-2.0.7-r3` refused to build.

glevand:
Here's the pth build error:

```
./libtool --mode=compile --quiet aarch64-cros-linux-gnu-gcc -c -I. -O2 -pipe -mtune=generic pth_mctx.c

pth_mctx.c: In function '__pth_mctx_set':
pth_mctx.c:480:2: error: #error "Unsupported Linux (g)libc version and/or platform"
 #error "Unsupported Linux (g)libc version and/or platform"
  ^
 * ERROR: dev-libs/pth-2.0.7-r3::portage-stable failed (compile phase):
```

dm0-:
It looks like it's just a hard-coded test that disables `makecontext`/`swapcontext` when cross-compiling.  Adding `ac_cv_check_mcsc=yes` to the configure line appears to work.

garygrebus:
I see this change got reverted.   (Why it caused pth build errors was puzzling, but it looks like maybe gnupg is the only package with a dependency on pth?).

I've been using this patch for a while and haven't seen the failure building for arm64, either running scripts/build_packages, or explicitly doing emerge_arm64_usr on dev_libs/pth.

Is there something I can do to help resolve the problem?   Or, I noticed the libgcrypt package was added, so perhaps you're going to address the need for crypto in coreos_install in some other way?

dm0-:
I've merged a few updates that move to GnuPG 2.1 and build it on arm64 in #2229, which avoids the pth cross-compilation problems.

garygrebus:
Cool!   Thanks you.

PR #2159, opened 2016-09-01
Slotted Go: The Happening

marineam:
In short, if ebuilds need to specify a particular version of Go they can set `COREOS_GO_VERSION=go1.6` or similar. Otherwise the version is controlled by the default value in `coreos-go-depend.eclass`

For command line usage outside of ebuilds the `go` and `gofmt` commands by default use to the highest version installed but that can be changed with `eselect go`. Regardless of what the default version is you always use versioned commands like `go1.6` and `gofmt1.6` directly.

Depends on https://github.com/coreos/eselect-go/pull/1


mischief:
lgtm

PR #2160, opened 2016-09-01
rkt: 1.13.0 -> 1.14.0

steveeJ:


crawford:
LGTM

dm0-:
The systemd version in the overlay says v229, but that CoreOS version is v231.

steveeJ:
You're probably right that it needs changing but I don't know on top of my head what effect the variable has. one of the @coreos/rkt-maintainers might know.

alban:
Yes, it needs to match.

PXE_SYSTEMD_VERSION is passed to `./configure --coreos-local-pxe-image-systemd-version`, then the build system uses that to write the `/systemd-version` file in the stage1 ACI with the correct version. Then, stage1's init uses it to know the version via [GetFlavor](https://github.com/coreos/rkt/blob/v1.14.0/stage1/init/common/pod.go#L1002). It is used to know if systemd supports some feature, to know if we can use them when writing the unit files.

I don't think stage1's code behaves differently between v229 and v231 at the moment. But it might happen later with https://github.com/coreos/rkt/pull/2826, depending how it is implemented.

steveeJ:
@alban 
It would be more elegant to introspect the version from the image and be able to drop this variable from the ebuild.

PR #2163, opened 2016-09-06
Revert "app-emulation/rkt: Remove go environment settings"

crawford:
After this we're no longer passing -nopie to the linker, rkt is
ending up with a TEXTREL section and SELinux is blocking execution.
Revert until it's fixed.


crawford:
Fixes https://github.com/coreos/bugs/issues/1525.

dm0-:
LGTM

PR #2167, opened 2016-09-06
app-emulation/docker: revert and backport stats api fix

crawford:
Revert back to 1.11.2 and include a fix for the stats API.


crawford:
Fixes https://github.com/coreos/bugs/issues/1526.

mischief:
@crawford it appears your revert will goto 1.11.2, not 1.10.3. is that intended?

crawford:
@mischief yes. 1.10.3 shipped in the Stable. We're going to stick with 1.11.2 for Beta.

mischief:
ok, your original message said 1.10.3 so i wanted to check. lgtm

crawford:
woops, good catch.

PR #2172, opened 2016-09-06
app-emulation/docker: use correct git revision

crawford:
Broken in b9a6d189. Not sure where that revision came from.


mischief:
eff. lgtm

PR #2174, opened 2016-09-07
add support for ppc64le platform

houstar:


profiles/coreos/ppc64le/README
```diff
@@ -0,0 +1,4 @@
+This profile (and its sub-profiles) should *only* contain amd64
```
marineam: This isn't amd64 :)

profiles/coreos/ppc64le/generic/package.use
```diff
@@ -0,0 +1,8 @@
+# Enable SELinux for amd64 targets
+coreos-base/coreos	    selinux
```
marineam: Does selinux work on ppc64le? Or even if it does perhaps save this for a later PR, cross-compiling selinux is problematic iirc.

bgilbert:
We're now working on [Fedora CoreOS](https://groups.google.com/d/topic/coreos-user/zgqkG88DS3U/discussion), the successor to Container Linux, and we expect most major development to occur there instead.  Meanwhile, Container Linux will be fully maintained into 2020 but won't see many new features.  Thank you for submitting this PR and we're sorry that we won't be able to merge it.

PR #2175, opened 2016-09-07
app-emulation: revert docker components

crawford:
Forgot that Docker was broken into many different, coupled components. Need to roll these back as well to be compatible with 1.11.2.


mischief:
*cross fingers* lgtm

PR #2179, opened 2016-09-07
sys-kernel/bootengine: more provider support

crawford:
Includes cdrom_id rules and network support for DigitalOcean (via
coreos-metadata).


dm0-:
LGTM

PR #2187, opened 2016-09-09
multilib-strict and stricter fixes

marineam:
I'm tinkering w/ performing builds that make a number of checks portage performs fatal. This would avoid things like missing TEXTREL issues. Before that can happen here are a few minor things to fix up.


mischief:
lgtm

PR #2188, opened 2016-09-10
sys-libs/ncurses: fix some compiler warnings

marineam:
From http://ncurses.scripts.mit.edu/?p=ncurses.git;a=commit;h=73ab536b636227eed291dad213ca88c93d422fb8

The important part is the tiny change to use the attr_get function
instead of the macro fixing:

    test/ncurses.c:1681:29: warning: the comparison will always evaluate
    as ‘true’ for the address of ‘old_attr’ will never be NULL [-Waddress]

which causes portage to abort with FEATURES=stricter


vcaputo:
lgtm

PR #2189, opened 2016-09-14
coreos-base/oem-ec2-compat: use ttyS0 for console

crawford:
On HVM instances, tty0 is actually a graphical interface. Since the
interface exposed is non-interactive, it's not possible to scroll
through the logs, which results in lots of missing information. Setting
the console to ttyS0, on the other hand, sends the output to the system
log.

PV instances don't have this issue since they are xen-based and use
hvc0.


coreos-base/oem-ec2-compat/files/grub-ec2.cfg
```diff
@@ -5,3 +5,7 @@ set oem_id="ec2"
 # Blacklist the Xen framebuffer module so it doesn't get loaded at boot
 # Disable `ens3` style names, so eth0 is used for both ixgbevf or xen.
 set linux_append="modprobe.blacklist=xen_fbfront net.ifnames=0"
+
+if [ "$grub_platform" = pc ]; then
+	set linux_console="console=ttyS0,115200n8"
```
marineam: Add grub serial so you see it's output too:

          serial com0 --speed=115200 --word=8 --parity=no
￼          terminal_input console serial_com0
￼          terminal_output console serial_com0

crawford: That's an unfortunate side-effect. Added.

coreos-base/oem-ec2-compat/files/grub-ec2.cfg
```diff
@@ -5,3 +5,10 @@ set oem_id="ec2"
...
```
marineam: Very intentional, if the config has customized the linux console it is responsible for configuring grub too.

Oh, but you can actually drop console, so grub also doesn't do the dual output, only writing to serial_com0

marineam: So forcing you to configure grub actually makes grub behave correctly too! :-P

PR #2194, opened 2016-09-16
sys-libs/glibc: import the C.UTF-8 locale from Fedora

marineam:
The 2.23 version of the patch is used by Fedora for glibc versions
2.21 through 2.23. Although currently unused by the ebuild the 2.24
version of the patch is included for future reference.

The latest revision can be found here:
http://pkgs.fedoraproject.org/cgit/rpms/glibc.git/tree/glibc-c-utf8-locale.patch

Note: this does not enable C.UTF-8 in images yet, INSTALL_MASK must be adjusted and calling localedef must be added to the image building process. This does enable C.UTF-8 in the SDK though.


marineam:
Bug: https://github.com/coreos/bugs/issues/112

PR #2205, opened 2016-09-28
app-admin/*-wrapper: clean up wrappers

crawford:
TODO:

  - [x] Double check service timeouts
  - [x] Figure out etcd certs
  - [x] Merge flannel.service and flanneld.service


crawford:
Supersedes https://github.com/coreos/coreos-overlay/pull/2094.

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,36 +4,72 @@
...
 fi
 
-KUBELET_ACI="${KUBELET_ACI:-quay.io/coreos/hyperkube}"
+KUBELET_IMG_TAG="${KUBELET_IMG_TAG:-${KUBELET_VERSION}}"
+KUBELET_IMG="${KUBELET_ACI}"
+
+require_ev_one KUBELET_IMG KUBELET_IMG_TAG
+
+if [[ ! -z ${KUBELET_IMG_TAG} ]]; then
+	KUBELET_IMG_URL=${KUBELET_IMG_URL:-"quay.io/coreos/hypberkube"}
```
aaronlevy: s/hypberkube/hyperkube

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,36 +4,72 @@
...
+	KUBELET_IMG="${KUBELET_IMG_URL}:${KUBELET_IMG_TAG}"
+	RKT_RUN_ARGS="${RKT_RUN_ARGS} --trust-keys-from-https"
```
aaronlevy: Why do we only add this if user provides a KUBELET_IMG_TAG?

crawford: It's assumed that when they use the tag, they use our default image URL which needs that argument. If the user provides KUBELET_IMG_URL directly, this flag isn't necessarily needed.

What this should probably check instead is whether or not KUBELET_IMG_URL is set.

crawford: I just updated the logic for the kubelet. Let me know if that makes more sense. I'll update the others if so.

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,36 +4,74 @@
...
+	for rev in $@ ; do
+		if [[ ! -z ${!rev} ]]; then
+			return
+		fi
+	done
+	echo One of $@ must be set
+	exit 1
+}
+
+if [[ ! -z ${KUBELET_VERSION} ]]; then
```
aaronlevy: nit: use -n and quote the variable: `[[ -n "${FOO}" ]]`

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,36 +4,74 @@
...
+	echo KUBELET_VERSION environment variable is deprecated, please use KUBELET_IMG_TAG instead
+fi
+
+if [[ ! -z ${KUBELET_ACI} ]]; then
```
aaronlevy: nit: use -n and quote the variable: `[[ -n "${FOO}" ]]`

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,36 +4,74 @@
...
+	echo KUBELET_ACI environment variable is deprecated, please use the KUBELET_IMG instead
```
aaronlevy: I think it should be "please use the KUBELET_IMG_URL instead" -- that var is closer to KUBELET_ACI use.

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,36 +4,74 @@
...
 fi
 
-KUBELET_ACI="${KUBELET_ACI:-quay.io/coreos/hyperkube}"
+KUBELET_IMG_TAG="${KUBELET_IMG_TAG:-${KUBELET_VERSION}}"
+KUBELET_IMG="${KUBELET_ACI}"
```
aaronlevy: I think this is incorrect. KUBELET_ACI is equivalent to KUBELET_IMG_URL. See usage: 
https://github.com/coreos/coreos-kubernetes/blob/master/single-node/user-data#L11
https://github.com/coreos/coreos-kubernetes/blob/master/single-node/user-data#L88

aaronlevy: Based on this, the logic below likely needs to change too.

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -4,36 +4,74 @@
...
-  --trust-keys-from-https \
-  $RKT_OPTS \
-  --stage1-from-dir=stage1-fly.aci \
-  ${KUBELET_ACI}:${KUBELET_VERSION} --exec=/kubelet -- "$@"
-
+RKT=${RKT:-"/usr/bin/rkt"}
+RKT_STAGE1_ARG=${RKT_STAGE1_ARG:-"--stage1-from-dir=stage1-fly.aci"}
+set -x
+exec ${RKT} ${RKT_GLOBAL_ARGS} \
+	run ${RKT_RUN_ARGS} \
```
chancez: Needs `/var/log` mounted as RW.

app-admin/kubelet-wrapper/files/kubelet-wrapper
crawford: Why would the logic need to change?

app-admin/kubelet-wrapper/files/kubelet-wrapper
crawford: It needs write access? What is it writing and why isn't it using the journal?

app-admin/kubelet-wrapper/files/kubelet-wrapper
aaronlevy: Sorry for the vague comment, so actually seems pretty close.

Common case (our deployment scripts) will be:

```
KUBELET_VERSION=v1.1.1
KUBELET_ACI=quay.io/coreos/hyperkube
```

So if we change to:

```
KUBELET_IMG_URL=${KUBELET_IMG_URL:-${KUBELET_ACI}}
```

Then maybe the only logic change we need is to check `if KUBELET_IMG_URL == quay.io/coreos/hyperkube; then trust keys`

app-admin/kubelet-wrapper/files/kubelet-wrapper
aaronlevy: I'm also on the fence if we really need to hard-code this or we should leave it up to operator/deployment scripts to set this via RKT_OPTS

chancez: @crawford It's creating symlinks from `/var/lib/docker/containers/$containerID/$containerID-json.log` to `/var/log/containers for the logging addon. The symlinks contain extra k8s specific metadata within the filename so the logging addon can correlate docker container logs to the actual k8s pod they correspond to.

We already set this in [coreos-kubernetes](https://github.com/coreos/coreos-kubernetes/blob/master/multi-node/generic/controller-install.sh#L103) now by default so we don't neecessarily need to do it here I guess.

app-admin/kubelet-wrapper/files/kubelet-wrapper
crawford: I updated it so that if the image comes from quay, it will pull in the keys.

app-admin/etcd-wrapper/files/etcd-member.service
```diff
@@ -0,0 +1,26 @@
+[Unit]
+Description=etcd (System Application Container)
+Documentation=https://github.com/coreos/etcd
+Conflicts=etcd.service
+Conflicts=etcd2.service
+
+[Service]
+Type=notify
+Restart=always
```
xiang90: not sure if this is a good choice. if user removed the local member, etcd will exit. but systemd will restart it. this wont cause etcd problem, but etcd will fall into restart loop.

app-admin/etcd-wrapper/files/etcd-member.service
```diff
@@ -0,0 +1,26 @@
...
+RestartSec=10s
+LimitNOFILE=40000
+TimeoutStartSec=60
+
+Environment="ETCD_IMG_TAG=v3.0.7"
```
xiang90: why 3.0.7? do we have a policy to set the default version?

app-admin/etcd-wrapper/files/etcd-member.service
```diff
@@ -0,0 +1,26 @@
...
+Environment="ETCD_NAME=%m"
+Environment="ETCD_IMG_USER=etcd"
+Environment="ETCD_DATA_DIR=/var/lib/etcd"
+Environment="RKT_RUN_ARGS=--uuid-file-save=/var/lib/coreos/etcd-member-wrapper.uuid"
+
+ExecStartPre=/usr/bin/mkdir --parents /var/lib/coreos
+ExecStartPre=-/usr/bin/rkt rm --uuid-file=/var/lib/coreos/etcd-member-wrapper.uuid
+ExecStart=/usr/lib/coreos/etcd-wrapper $ETCD_OPTS
+ExecStop=-/usr/bin/rkt stop --uuid-file=/var/lib/coreos/etcd-member-wrapper.uuid
+
```
xiang90: do we need Wants=network.target?

app-admin/etcd-wrapper/files/etcd-member.service
crawford: Ah, let's change it to restart on failure.

app-admin/etcd-wrapper/files/etcd-member.service
crawford: We are going to use whichever version the etcd team suggests. Since there is a default (which will change as the OS updates), etcd will always need to be able to migrate forward.

app-admin/etcd-wrapper/files/etcd-member.service
crawford: Probably. I'll toss that in.

crawford:
Fixes https://github.com/coreos/bugs/issues/984.

crawford:
@aaronlevy @xiang90 can you take another look? I've made the requested changes.

app-admin/etcd-wrapper/files/etcd-member.service
xiang90: let's use 3.0.10. we can almost always try latest patch version of etcd. It is usually the most stable one. We probably want to hold off for upgrading to the next minor version for 1 or 2 patch release. The first release of a minor version bump is usually not stable.

xiang90:
looks good on etcd part with my limited systemd knowledge.

crawford:
etcd still needs access to user-provided certs

crawford:
Ready for a final review!

@aaronlevy can you take one more peak? @pbx0 is currently testing an image.

@heyitsanthony can you guys take one more look at the script? I want to be sure that discovery and TLS are going to work properly.

gyuho:
For TLS, etcd needs the following flags

```
ETCD_CERT_FILE
ETCD_KEY_FILE
ETCD_CLIENT_CERT_AUTH
ETCD_TRUSTED_CA_FILE
ETCD_PEER_CERT_FILE
ETCD_PEER_KEY_FILE
ETCD_PEER_CLIENT_CERT_AUTH
ETCD_PEER_TRUSTED_CA_FILE
```

e.g.

```
etcd --name test-1 --data-dir /tmp/test-1.data \
    --listen-client-urls https://localhost:2379 --advertise-client-urls https://localhost:2379 \
    --listen-peer-urls https://localhost:2380 --initial-advertise-peer-urls https://localhost:2380 \
    --initial-cluster test-1=https://localhost:2380,test-2=https://localhost:22380,test-3=https://localhost:32380 \
    --initial-cluster-token test-token --initial-cluster-state new \
    --client-cert-auth \
    --cert-file /tmp/test-certs/test-1.pem \
    --key-file /tmp/test-certs/test-1-key.pem \
    --trusted-ca-file /tmp/test-certs/trusted-ca.pem \
    --peer-client-cert-auth \
    --peer-cert-file /tmp/test-certs/test-1.pem \
    --peer-key-file /tmp/test-certs/test-1-key.pem \
    --peer-trusted-ca-file /tmp/test-certs/trusted-ca.pem
```

app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -1,39 +1,75 @@
...
+	--mount volume=etc-kubernetes,target=/etc/kubernetes \
+	--mount volume=etc-ssl-certs,target=/etc/ssl/certs \
+	--mount volume=usr-share-certs,target=/usr/share/ca-certificates \
+	--mount volume=var-lib-docker,target=/var/lib/docker \
+	--mount volume=var-lib-kubelet,target=/var/lib/kubelet \
+	--mount volume=os-release,target=/etc/os-release \
+	--mount volume=run,target=/run \
+	${RKT_STAGE1_ARG} \
+	${KUBELET_IMAGE} \
+		${KUBELET_IMAGE_ARGS} \
```
aaronlevy: This doesn't seem to be set anywhere. We can probably default to `--exec/kubelet`

aaronlevy:
LGTM on kubelet pending functional review by @pbx0

crawford:
Fixes https://github.com/coreos/bugs/issues/1566.

crawford:
@aaronlevy @pbx0 tests pass now.

pbx0:
LGTM

gyuho:
LGTM on etcd-wrapper. 

Confirmed that etcd-wrapper works well (tested 3-node cluster with client  +  peer TLS with the image from Crawford), tested some rebooting and client requests, tested discovery. looks good and all work as expected on my side.

app-admin/flannel-wrapper/files/flannel-docker-opts.service
```diff
@@ -0,0 +1,23 @@
+[Unit]
+Description=flannel docker export service - Network fabric for containers (System Application Container)
+Documentation=https://github.com/coreos/flannel
+After=flannel.service
+PartOf=flannel.service
```
monder: Shouldn't it be flannel**d**.service?

crawford: Yes it should. Good catch.

app-admin/etcd-wrapper/files/etcd-wrapper
```diff
@@ -1,76 +1,80 @@
...
-        --user=$(id -u ${ETCD_IMG_USER}) \
-        --exec=${ETCD_IMG_EXEC} -- \
-        "$@"
+	run ${RKT_RUN_ARGS} \
+	--volume data-dir,kind=host,source="${ETCD_DATA_DIR}",readOnly=false \
+	--volume etc-ssl-certs,kind=host,source="${ETCD_SSL_DIR}",readOnly=true \
+	--volume usr-share-certs,kind=host,source=/usr/share/ca-certificates,readOnly=true \
+	--volume etc-hosts,kind=host,source=/etc/hosts,readOnly=true \
+	--volume etc-resolv,kind=host,source=/etc/resolv.conf,readOnly=true \
+	--mount volume=data-dir,target=/var/lib/etcd \
```
monder: That's not right as well.
Imagine this:
- We overwrite ETCD_DATA_DIR from the calling unit to /mnt/etcd-node1
- rkt mounts /mnt/etcd-node1 into container as /var/lib/etcd
- etcd trues to load data in /mnt/etcd-node1 within the container.
Obviously it doesn't work because of --inherit-env

We either should mount it as ${ETCD_DATA_DIR} inside or (imho better) don't propagate this env variable to the container to avoid any permission errors and collisions.

PR #2211, opened 2016-09-30
ON HOLD: app-emulation/{containerd,docker}: separately managed containerd unit

joeatwork:
This changeset proposes a separately managed, socket activated containerd service.


app-emulation/docker/files/docker.service
```diff
@@ -7,11 +7,14 @@ Requires=docker.socket early-docker.target
 [Service]
 Type=notify
 EnvironmentFile=-/run/flannel_docker_opts.env
-MountFlags=slave
+# We need shared mounts on the file system so that containerd can
```
marineam: Perhaps JoinsNamespaceOf=containerd.service?

marineam:
- I'd say do mount-flags=slave in containerd and then have docker join the containerd namespace.
- No particular reason to enable units in coreos-overlay in coreos-init instead of their respective ebuilds.
- early-docker is for containers that themselves provide resources needed by docker. Previously this was flannel which has moved to rkt but our users have come to depend on early-docker. iirc mischief or philips said there was a 1-to-1 requirement for docker and containerd so you will need a special early-containerd.service as well.
- I'm not sure there is a particular need for containerd to be socket activated. docker.service could merely have Requires=containerd.service After=containerd.service and both will start when docker gets activated.

joeatwork:
Discussed IRL - will wait on JoinNamespaceOf changes in favor of the other revisions

joeatwork:
Another change needed - this shows up in the journal
```
level=warning msg="containerd: low RLIMIT_NOFILE changing to max" current=1024 max=4096
```

Use the same resource un-constraints from docker.socket in containerd.socket

app-emulation/docker/files/docker.service
joeatwork: Discussed IRL - deferring JoinsNamespaceOf for now

joeatwork:
Will need to be rebased/tested after #2215 is merged.

joeatwork:
This change is on hold pending a port of https://github.com/coreos/coreos-overlay/pull/2215 into master

joeatwork:
Closing this in favor of #2220

PR #2212, opened 2016-09-30
app-emulation/qemu-binfmt: Add new ebuild

glevand:
When building CoreOS on a CoreOS installation we need to have qemu binfmt support setup. This adds arm64 support to amd64 hosts.


glevand:
Rebased to latest.
/cc @marineam

glevand:
Moved from board-packages to coreos.

crawford:
Is this intended for dev images? This won't work in the production images since QEMU isn't present.

marineam:
@crawford we already chatted, the idea was to have it in prod images so the SDK works w/o extra configuration on the host. I suggested a `cork setup` command instead so we don't ship a config that isn't valid/applicable on most systems. Closing pending something more SDK-specific.

PR #2215, opened 2016-10-04
app-emulation/docker: revert to 1.11.2

crawford:


marineam:
No, don't use revert here where there are changes unrelated to the version like the go migration. You will be building withe wrong go. I don't know what else is significant there.

marineam:
Dropping a number of things from the service file that should be there too.

crawford:
> Dropping a number of things from the service file that should be there too.

Why would we make any modifications to what we shipped in 1153.4.0?

marineam:
Because we fixed a number of things since then

crawford:
Can you give me specific changes since I can't recall what we fixed?

marineam:
I'd say leave the following alone:

```
-# Having non-zero Limit*s causes performance problems due to accounting overhead
￼ -# in the kernel. We recommend using cgroups to do container-local accounting.
￼ -LimitNPROC=infinity
￼ -LimitCORE=infinity
￼ -# Uncomment TasksMax if your systemd version supports it.
￼ -# Only systemd 226 and above support this version.
￼ -TasksMax=infinity
￼ -TimeoutStartSec=0
￼ -# set delegate yes so that systemd does not reset the cgroups of docker containers
￼ -Delegate=yes
```

I don't know if Type=notify is valid for the older version so you can stick with removing that one.

app-emulation/docker/docker-9999.ebuild
```diff
@@ -263,12 +260,7 @@ src_compile() {
 
 src_install() {
 	VERSION="$(cat VERSION)"
-	newbin "bundles/$VERSION/dynbinary-client/docker-$VERSION" docker
-	newbin "bundles/$VERSION/dynbinary-daemon/dockerd-$VERSION" dockerd
-	newbin "bundles/$VERSION/dynbinary-daemon/docker-proxy-$VERSION" docker-proxy
-	dosym containerd /usr/bin/docker-containerd
```
marineam: Is it correct to not install these symlinks?

PR #2218, opened 2016-10-05
Merge pull request #2215 from crawford/docker

joeatwork:
app-emulation/docker: revert to 1.11.2

These changes are unchanged from #2215

PR #2220, opened 2016-10-06
Manage containerd as a systemd unit

joeatwork:
Changes from #2211, rebased on docker 1.11.2

~~Hold on review until #2218 is merged~~


app-emulation/docker/files/early-docker.service
```diff
@@ -1,13 +1,17 @@
...
+Requires=containerd.service early-docker.socket
 
 [Service]
 Environment=TMPDIR=/var/tmp
-MountFlags=slave
-ExecStart=/usr/lib/coreos/dockerd daemon --host=fd:// --bridge=none --iptables=false --ip-masq=false --exec-root=/var/run/early-docker --graph=/var/lib/early-docker --pidfile=/var/run/early-docker.pid $DOCKER_OPTS $DOCKER_CGROUPS
+
+# We need shared mounts on the file system so that containerd can
+# see our bundles.
+# MountFlags=slave
```
marineam: Don't leave old config/code commented out, just delete.

app-emulation/docker/files/early-docker.service
```diff
@@ -1,13 +1,17 @@
 [Unit]
 Description=Early Docker Application Container Engine
 Documentation=http://docs.docker.com
-After=early-docker.socket
-Requires=early-docker.socket
+After=containerd.service early-docker.socket
+Requires=containerd.service early-docker.socket
```
marineam: You are using the same containderd for both docker and early-docker. I seem to recall @mischief discovering that this isn't possible. Is that not the case?

joeatwork: I think I was able to observe this working - @mischief when you're around, can you talk me through the issues you had? It's not impossible that I misread the results of my (very manual, ad-hoc) testing, or didn't test the right cases.

joeatwork:
Discussed IRL:

- Run flanneld with early-docker, as originally implemented, as a kola test. This will allow you to observe how early-docker actually behaves - networking conflicts between docker and early-docker may have been what shut this effort down back in the day

- Maybe just roll early-docker back to where it is on master if the current approach has *any* problems - early docker is deprecated and we don't need to add cool features to it.

- IF you have socket conflicts running early-docker with a child docker-containerd process, move the socket in use by the real-docker containerd process. This could have other beneficial effects!

app-emulation/docker/files/early-docker.service
joeatwork: Discussed, see notes below

joeatwork:
Superceeded by #2247

PR #2224, opened 2016-10-07
Containerize the GCE agent

dm0-:
This replaces the two existing `app-emulation/google-*` ebuilds with a new one.  The new ebuild is an update from the same upstream code repository, which now uses regular `/usr` paths instead of installing into `/usr/share/oem`.  It only installs the Python module and script entry points, not any of the init system files etc.

A new `coreos-oem-gce` meta-package ebuild has been added to install a functional CoreOS environment from its dependencies (a la the `coreos-dev` ebuild).  The resulting file system from installing this ebuild is packaged into an ACI file that will run the GCE agent.

The `oem-gce` ebuild has been modified to have no dependencies and to only install the OEM configuration files and a service file to start the container.  Note that its version number no longer needs to be synchronized with the agent's ebuild, since the agent gets installed into the ACI, while this gets installed onto the host.

Ignition was bumped to support the new service file.

I've updated the metadata since ebuilds were added and removed.

This requires coreos/scripts#593.

This fixes coreos/bugs#1468 and coreos/bugs#1543.


crawford:
I'm not clear on the purpose of the new coreos-oem-gce package. How does this work?

dm0-:
The SDK requires a base package to install an image.  It serves the same purpose as `coreos-base/coreos` or `coreos-base/coreos-dev`.

app-emulation/google-compute-engine/google-compute-engine-20160930.ebuild
```diff
@@ -0,0 +1,47 @@
...
+	sys-apps/shadow
+	sys-libs/glibc
+	sys-libs/nss-usrfiles
+"
+
+S="${WORKDIR}/compute-image-packages-${PV}"
+
+src_prepare() {
+	# Don't attempt to mess with our host keys.
+	sed -i -e '/set_host_keys/s/true/false/i' \
```
zmarano: Instead of modifying the python code, all you need to do is create a template file, /etc/default/instance_configs.cfg.template with the following to override (you can override any of the settings like this):

[InstanceSetup]
set_host_keys = false

dm0-: I've added a template file in #2235.

PR #2226, opened 2016-10-10
profiles: accept dev-libs/libgcrypt-1.7.3 for arm64

mischief:


glevand:
@mischief  Just to let you know, we we getting duplicate entries, so to avoid that I had put the entries in alphabetical order and put that comment at the top of the file.

PR #2230, opened 2016-10-14
kubelet-wrapper: delete previous pods on launch

euank:
Works around https://github.com/coreos/bugs/issues/1612

Perhaps it would be better to save the uuid in something like `/var/run/kubelet_wrapper.uuid` instead to sorta match pid files.

cc @jonboulle  @pbx0

### DIFF
--- a/app-admin/kubelet-wrapper/files/kubelet-wrapper
+++ b/app-admin/kubelet-wrapper/files/kubelet-wrapper
@@ -51,9 +51,16 @@ if [[ "${KUBELET_IMAGE%%/*}" == "quay.io" ]]; then
 	RKT_RUN_ARGS="${RKT_RUN_ARGS} --trust-keys-from-https"
 fi
 
+UUID_FILE="/var/lib/kubelet-wrapper/last-uuid"
+
+if [ -f "${UUID_FILE}" ]; then
+  /usr/bin/rkt rm --uuid-file="${UUID_FILE}" &>/dev/null || true
+fi
+
 mkdir --parents /etc/kubernetes
 mkdir --parents /var/lib/docker
 mkdir --parents /var/lib/kubelet
+mkdir --parents /var/lib/kubelet-wrapper
 mkdir --parents /run/kubelet
 
 RKT="${RKT:-/usr/bin/rkt}"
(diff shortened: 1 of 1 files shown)


crawford:
I'm going to absorb this into another PR where we are doing something [similar](https://github.com/coreos/coreos-overlay/pull/2205/files#diff-42cb6a4a730f8dd5893cb0f6138cb1f0R19). I'll move the logic from the services into the wrappers.

euank:
Cool, closing in favor of your pr.

crawford:
Actually, having thought about it more, I think it's more appropriate for this to live in the service units (as it's already written). I recommend doing something similar to the [etcd-member service](https://github.com/coreos/coreos-overlay/pull/2205/files#diff-42cb6a4a730f8dd5893cb0f6138cb1f0R19).

philips:
@crawford what version of the OS ships this?

crawford:
@philips https://coreos.com/releases/#1214.0.0.

euank:
@crawford this diff didn't ship and after thinking about it more, your change can't cover this because we don't actually ship a kubelet service file directly (rather simply documenting how we expect a customer to use it).

To sorta keep the detail of rkt well-encapsulated in this bash script and, more importantly, to not require users to update their service files more, I think we should reconsider this change.

SGTY?

crawford:
@euank sorry, the version I referred to shipped the etcd-member-change, which is what I thought @philips was referring to. The changes in _this_ PR haven't shipped.

> To sorta keep the detail of rkt well-encapsulated in this bash script and, more importantly, to not require users to update their service files more, I think we should reconsider this change.

That change would prevent multiple instances of the kubelet wrapper from being run. A user could inadvertently kill their service by manually running the service. I am not a fan. The rest of the services handle this directly instead of depending on the wrapper script. I expect kubelet to be able to do the same.

euank:
> A user could inadvertently kill their service by manually running the service.

A user would leak a pod, not kill their service. `rkt rm || true` will be a noop if the pod in question is running.

They could inadvertantly remove and already stopped pod, sure, but that doesn't seem so scary.

crawford:
Why do you want this functionality here instead of in the service? This just doesn't feel right.

euank:
Closing in favor of https://github.com/coreos/coreos-kubernetes/issues/759 per @crawford's preference.

PR #2231, opened 2016-10-17
sys-kernel/coreos-modules: add bpf and megaraid

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/1602.

I have no idea where these options are supposed to be physically located within the common config, so I just tossed them at the end.

crawford:
Fixes https://github.com/coreos/bugs/issues/1057.

PR #2234, opened 2016-10-18
app-emulation/rkt: bump to v1.17.0

crawford:


crawford:
Which version of CoreOS should be used for the PXE image?

/cc @lucab

dm0-:
The 1.17 release didn't change CoreOS versions for the stage1 ACI, so that shouldn't change.

https://github.com/coreos/rkt/tree/dbdd0758d856b208d8d60f9c5303afb93bc3ca35/stage1/usr_from_coreos

It will need to be updated for the next release, though.

crawford:
Fixes https://github.com/coreos/bugs/issues/1616.

lucab:
As @dm0- said, this is still at systemd v231 and coreos-alpha pxe 1151.0.0.

This information are already inside our makefile, is there any way for us to expose them in a way that you can directly use it inside the ebuild? Consuming it via `make print-foo` would be nice but I think it won't be possible here as it is before autogen+configure; any better options?

PR #2236, opened 2016-10-19
app-admin/locksmith: bump to v0.4.2

mischief:


mischief:
fixes coreos/bugs#1171

PR #2238, opened 2016-10-19
app-admin/kubelet-wrapper: bind var/log/containers

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/1614.

aaronlevy:
Is this something we want as part of default wrapper (e.g. one added -- pretty much won't change)?

pbx0:
Right now we bind mount in `/var/log` in the kubelet.service RKT_OPTS for coreos-kubernetes.

chancez:
You unfortunately need `/var/log` in order for the k8s logging addon to work properly by default. Docker container logs are symlinks within `/var/log/containers`, but the fluentd logging addon is storing it's cursors in `/var/log`. 

If we want people to be able to use the upstream logging addon with no changes, `/var/log` must be a mount as well. Otherwise if the fluentd pod get's re-created, it will lose it's cursor files in the container file system, and re-ingest logs it's already ingested.

crawford:
@chancez it sounds like `/var/log/containers` isn't core to the kubelet and therefore shouldn't be included. If the kubelet service contains that option, this should work fine. I'm going to go ahead and close this out.

robszumski:
FWIW, this comes up frequently as a complaint/bug from users.

PR #2243, opened 2016-10-20
sys-apps/systemd: include fix for zero-length msg

crawford:


joeatwork:
LGTM this commit is https://github.com/coreos/systemd/commit/c193398d41ce1594cfa5b730b2c69aeddb69bc52 as expected

PR #2244, opened 2016-10-20
sys-kernel/coreos-*: include recent fixes

crawford:


joeatwork:
LGTM Walked through with @crawford - Patches are familiar, kernel source revision number is correct

PR #2254, opened 2016-10-25
dev-libs/openssl: bump to 1.0.2j

crawford:
Addresses CVE-2016-8610.


mischief:
[LGTM](https://lgtm.in/i/mZ5BS4RWp)

marineam:
Something to keep in mind: this is going to disable sslv2 which was a mess in the past but I believe upstream finally fixed it so it the ABI isn't modified (sslv2 specific functions just become stubs). Probably not an issue but worth watching out for.

PR #2257, opened 2016-10-28
coreos-devel/kola-data: support data for testing with old docker

joeatwork:
PLEASE LOOK CLOSELY! This was a *very* trial and error process, I pretty much hammered the Jukebox on the side over and over until it started playing my song.


coreos-devel/kola-data/kola-data-9999.ebuild
```diff
@@ -0,0 +1,21 @@
...
+
+KEYWORDS="amd64"
+DESCRIPTION="Binary assets for use in kola tests"
+HOMEPAGE="https://github.com/coreos/mantle"
+LICENSE="Apache"
+
+SRC_URI="https://get.docker.com/builds/Linux/x86_64/docker-1.9.1.tgz"
+SLOT="0"
+
+QA_PREBUILT="*"
```
joeatwork: @marineam is this the right value for `QA_PREBUILT`?

joeatwork:
Crawford took a look, @marineam it's up to you :)

coreos-devel/kola-data/kola-data-9999.ebuild
joeatwork: Fixed in 8356311e0b498f644709adc88c8686b1c78a228f

marineam:
Note: for this to be available in the SDK by default you'll need to add it to the sdk-depends ebuild

joeatwork:
Thanks!

joeatwork:
Added to sdk-depends in https://github.com/coreos/coreos-overlay/pull/2258

PR #2260, opened 2016-10-31
sys-kernel/coreos-{sources,modules,kernel}: bump to v4.8.6

mischief:


mischief:
the patch series has been rebased.

PR #2262, opened 2016-11-01
coreos-base/coreos-metadata: bump to v0.6.1

crawford:


crawford:
This version adds an infinite retry for Azure leases. A part of me wanted to add a `Requires` and `After` on `network.target` so that systemd could prevent any deadlocks on Azure, but not all of the providers actually require the network. Worst case scenario, the system locks up for 90 seconds before systemd kills off `coreos-metadata.service` and unblocks the dependency chain.

PR #2275, opened 2016-11-16
app-admin/flannel-wrapper: use oneshot for opts

crawford:
docker.service shouldn't be allowed to run until
flannel-docker-opts.service _finishes_ (since it is actually writing the
flags for Docker).


crawford:
Fixes https://github.com/coreos/bugs/issues/1659.

PR #2277, opened 2016-11-16
coreos-sources: Add arm64 verity hash

glevand:
Adds a place to inject the verity hash value in the arm64 kernel image.

cc: @mjg59


glevand:
Rebased to latest.

glevand:
Rebased to latest.

PR #2280, opened 2016-11-22
app-admin/fleet-wrapper: add fleet-wrapper

dongsupark:
Replaces `app-admin/fleet` with `app-admin/fleet-wrapper`, which runs fleet under `rkt fly`, to make it easier for users to test a particular release.

Fixes: https://github.com/coreos/bugs/issues/1669


crawford:
We can't remove fleet from the OS just yet. The plan is to publish documentation showing the wrapper script and service unit. If the user wants to use that method, they can, but they'll have to place the necessary components on their system (this can be done via Ignition).

dongsupark:
@crawford Interesting. I thought that we should replace fleet with fleet-wrapper, just like it was done via kubelet-wrapper or flannel-wrapper. Some policy must have changed recently. (Scratching my head)

crawford:
We replaced the kubelet with the kubelet-wrapper, but that was a very jarring migration and wasn't backward compatible. The flannel-wrapper always existed (just in a different form), so there was no issue there. The change you proposed will likely break a lot of existing systems when they upgrade and we don't think it is worth the risk. The plan is to sunset fleet and remove it from the OS. Users can migrate to this script if they wish, but it will need to be a conscious effort.

dongsupark:
@crawford Yes, I see. That has been also my understanding: "sunset fleet from OS". Without https://github.com/coreos/bugs/issues/1669, I wouldn't have cared at all in the first place. Thanks anyway.

PR #2283, opened 2016-11-29
import qemu from coreos-overlay and apply vpc patch

mischief:
the bump to r7 includes a number of security patches.


crawford:
Can you break the commits into an import and patch+version bump?

mischief:
ptal

PR #2295, opened 2016-11-30
app-admin/etcd-wrapper: set data dir properly

crawford:


gyuho:
lgtm. We tested this and confirmed that it works.

Thanks.

PR #2296, opened 2016-12-01
systemd: add TODO for hostnamed polkit rule

marineam:


crawford:
Can you file a bug as well?

marineam:
Done

PR #2297, opened 2016-12-02
app-emulation/docker: use journald logger

DrMurx:
Re-apply commit 3aaf6715af392de36877710c5f0e8d02d60bb1fd which has been
reverted in f2c7d9d00af623e96294c12909baeefff065fd65 because of missing
support for `docker logs`.

Recent docker versions support journald as source for `docker logs`.


crawford:
Unfortunately, we cannot make this change since Kubernetes reads the Docker logs directly. I'm dying inside as well. For now, we'll have to document how to use the journald backend.

PR #2298, opened 2016-12-02
Enable ACPI in arm64 CONFIG_CMDLINE

glevand:
Enables ACPI for arm64 and bumps kernel rev to -r2.

In follow up to [grub.cfg: Enable arm64 ACPI](https://github.com/coreos/scripts/pull/610).


marineam:
Closing in favor of the GRUB change, I only suggested the kernel config because I thought there was already a way to enable the EXTEND behavior. Probably a mistaken assumption I made when I saw the FORCE option. Sorry for the extra trouble.

PR #2300, opened 2016-12-06
Stop networkd from interfering with Docker network interfaces

dm0-:
This fixes coreos/bugs#1638.

It can be used to address coreos/bugs#1678.

Note that the problem with Docker from coreos/bugs#1485 is corrected by this, but that workaround is still in place to avoid changing behavior.  See coreos/systemd#73.


SpComb:
Would this change also serve to avoid the use of IPv6 autoconf on the host-side veths?

I think that currently the CoreOS host node will send ICMPv6 Router Solicitation messages on each container veth created by Docker, which isn't ideal. I'm not sure if it's possible for Docker containers to affect the host machine's IPv6 networking configuration that way.

dm0-:
@SpComb This change effectively makes veths invisible to networkd.  I don't know offhand if anything else sends router solicitation messages, but they will no longer be triggered by networkd.  (An exception since #2306 is veths named `ve-*`, but they should only be created by systemd-nspawn.)

From the issue you linked above, also note that systemd will support excluding devices by e.g. `Driver=!bridge veth` since systemd/systemd#4809.  We don't plan to backport that feature to CoreOS, though, so it will be available with v233.

PR #2307, opened 2016-12-07
net-misc/openssh: sync with upstream latest stable

dm0-:
This fixes GLSA 201612-18.


mischief:
i think you can drop bindist from the ebuild, as it's not used.

dm0-:
Okay, it's gone.

PR #2308, opened 2016-12-08
Expose container logs out of kubelet-wrapper

SchwarzM:
The current version of the kubelet-wrapper does not expose the container logs to external utilities like fluentd.

### DIFF
--- a/app-admin/kubelet-wrapper/files/kubelet-wrapper
+++ b/app-admin/kubelet-wrapper/files/kubelet-wrapper
@@ -54,7 +54,6 @@ fi
 mkdir --parents /etc/kubernetes
 mkdir --parents /var/lib/docker
 mkdir --parents /var/lib/kubelet
-mkdir --parents /var/log/containers
 mkdir --parents /run/kubelet
 
 RKT="${RKT:-/usr/bin/rkt}"
@@ -68,7 +67,7 @@ exec ${RKT} ${RKT_GLOBAL_ARGS} \
 	--volume usr-share-certs,kind=host,source=/usr/share/ca-certificates,readOnly=true \
 	--volume var-lib-docker,kind=host,source=/var/lib/docker,readOnly=false \
 	--volume var-lib-kubelet,kind=host,source=/var/lib/kubelet,readOnly=false \
-	--volume var-log-containers,kind=host,source=/var/log/containers,readOnly=false \
+	--volume var-log,kind=host,source=/var/log,readOnly=false \
 	--volume os-release,kind=host,source=/usr/lib/os-release,readOnly=true \
 	--volume run,kind=host,source=/run,readOnly=false \
 	--mount volume=etc-kubernetes,target=/etc/kubernetes \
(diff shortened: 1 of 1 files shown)


app-admin/kubelet-wrapper/files/kubelet-wrapper
```diff
@@ -67,13 +67,15 @@ exec ${RKT} ${RKT_GLOBAL_ARGS} \
...
 	--volume var-lib-kubelet,kind=host,source=/var/lib/kubelet,readOnly=false \
+  --volume var-log-containers,kind=host,source=/var/log/containers,readOnly=false \
 	--volume os-release,kind=host,source=/usr/lib/os-release,readOnly=true \
 	--volume run,kind=host,source=/run,readOnly=false \
 	--mount volume=etc-kubernetes,target=/etc/kubernetes \
 	--mount volume=etc-ssl-certs,target=/etc/ssl/certs \
 	--mount volume=usr-share-certs,target=/usr/share/ca-certificates \
 	--mount volume=var-lib-docker,target=/var/lib/docker \
 	--mount volume=var-lib-kubelet,target=/var/lib/kubelet \
+  --mount volume=var-log-containers,target=/var/log/containers \
```
euank: nit: rest of the file is tabs

euank:
I believe that `/var/log/containers` will not exist initially and rkt will error out when a given volume source doesn't exist.

Is something else creating that directory? Will this work if kubelet wrapper is used without docker having been run?

SchwarzM:
Will check the directory ASAP and come back to you.

pbx0:
We do already mount `/var/log` in https://github.com/coreos/coreos-kubernetes/blob/master/single-node/user-data#L108. It appears we don't do this in our bootkube setup examples. Mounting in `/var/log` I think should solve the problem of it not being created before running the wrapper. Seems like this might be worth doing in the wrapper itself. What happens to user's already mounting in `/var/log` via `RKT_OPTS` if we update kubelet-wrapper?

euank:
@pbx0 one data-point in favour of all of `/var/log` is that the kubelet [here](https://github.com/kubernetes/kubernetes/blob/ee177dbbfe405b11460eca48e622b55bdcad87a0/pkg/kubelet/kubelet.go#L1214-L1216) exposes everything in that directory over http, not just `containers`.

I'm not a fan of that behaviour and I'm hopeful it'll change soon, but technically the kubelet has a dependency on all of `/var/log`

SchwarzM:
@bugroger as i am currently ill could you take a look at this ?

BugRoger:
Fixed spaces to be tabs. Also creating the missing `/var/lib/containers`...

euank:
We still have a pending question and comment from @pbx0 on this one, bump.

> Mounting in /var/log I think should solve the problem of it not being created before running the wrapper
> ...
> What happens to user's already mounting in /var/log via RKT_OPTS if we update kubelet-wrapper?

SchwarzM:
Hi,

sorry for the delay i was ill and could not check. 
Tried to check for extra rkt-opts with the following results:
same name, same mount -> works, second name takes precedence
same name, different mount -> works
different name, same mount -> works, first name takes precedence

as for using /var/log is fine by me but exposes much more than normally accessible.

kind regards
M. Schwarz

pbx0:
@euank @SchwarzM Lets do `/var/log` because https://github.com/coreos/coreos-kubernetes/issues/322#issuecomment-218278783. @chancez Cursor files still get stored outside `/var/log/containers` I presume?

Then, this LGTM.

chancez:
Yah, cursors are in `/var/log`: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-elasticsearch/fluentd-es-image/td-agent.conf#L115

PR #2312, opened 2016-12-09
sys-kernel/coreos-modules: support for xfs quotas

crawford:


crawford:
Should fix https://github.com/coreos/bugs/issues/1691.

vcaputo:
lgtm

PR #2313, opened 2016-12-11
Enable CONFIG_USB_ACM

xentac-google:
So that we can use other classes of usb serial devices.

Fixes coreos/bugs#1488.


xentac-google:
I have made the changes.  Thanks for looking at this!

crawford:
Thanks for the fix!

PR #2314, opened 2016-12-11
sys-auth/realmd: fix krb5-config detection

dm0-:


crawford:
Thought I saw this one before (https://github.com/coreos/coreos-overlay/pull/2304).

PR #2325, opened 2016-12-19
Active directory

mjg59:


mjg59:
https://github.com/coreos/bugs/issues/968

mjg59:
Updated based on @crawford's review

net-fs/samba/samba-4.5.1.ebuild
```diff
@@ -215,6 +208,11 @@ multilib_src_install() {
 		systemd_dounit "${FILESDIR}"/winbindd.service
```
crawford: I don't understand why I cannot leave a comment above this line...

We don't need the files installed on lines 201 and 202.

mjg59: Fixed.

net-fs/samba/samba-4.5.1.ebuild
```diff
@@ -204,11 +204,16 @@ multilib_src_install() {
...
 		systemd_dounit "${FILESDIR}"/samba.service
 	fi
-	rm ${D}/usr/bin/ldb*
-	rm ${D}/usr/lib/debug/usr/bin/ldb*
-	rm ${D}/usr/lib64/samba/ldb/*
-	rm ${D}/usr/bin/tdb*
-	rm ${D}/usr/lib/debug/usr/bin/tdb*
+	rm ${D}/usr/lib*/samba/ldb/*
+	mv ${D}/usr/bin/net /tmp
```
marineam: Should this mv be inside `if use minimal`? Also, copping a simple name like "net" into and out of /tmp sounds dangerous. Use "${T}" instead which is local to the build and thus won't collide with anything.

profiles/coreos/base/package.use
```diff
@@ -79,3 +79,5 @@ dev-cpp/glog gflags
 # enable rpc for rpc.rquotad
 sys-fs/quota rpc
 
+# don't ship the Samba daemons
+net-fs/samba minimal
```
marineam: Since this is a cusomized ebuild if you use `+minimal` in IUSE it'll turn on by default.

mjg59:
@marineam Updated.

PR #2329, opened 2016-12-20
sys-apps/systemd: Update to disable selinux permissions checks

mjg59:
Fix https://github.com/coreos/bugs/issues/1682


mjg59:
Ugh git add does not do what I want it to do when hitting symlinks. Fixed.

PR #2330, opened 2016-12-20
google-cloud-sdk: update to current release, replace gsutil

marineam:
Updating the existing gsutil and all of its dependencies is rather messy. Instead we can use gsutil from the google-cloud-sdk package and avoid most of the dependency mess.

Note: the gsutil command that gets installed does not make use of gcloud for authentication to preserve compatibility with existing behavior.


marineam:
Backstory: I wound up doing this because I briefly looked into adding the ppc64le cross toolchain which would need a new profile in coreos-overlay which would benefit from refreshing profiles in portage-stable which would be easier if we could clean up some outdated python packages...

I am glad I have all this yak fur to keep me warm this winter.

PR #2333, opened 2016-12-28
shadow: update for docker's userns-remap

euank:
We should have an automated test of this stuff, but I'd rather get all of --userns-remap working first.

This makes it so that the docker instructions of just touching `/etc/sub{u,g}id` and then starting docker with `--userns-remap=default` at least get a running dockerd.

- [ ] Verified the last sentence is true (pending image build)

Fixes the error seen in https://github.com/coreos/bugs/issues/1728


euank:
do-not-merge for the moment; I also have to update sys-apps/shadow so that such a version exists somewhere

euank:
This doesn't matter now that #2334 is merged, closing.

crawford:
I think we still want this though. It expresses an actual dependency.

PR #2334, opened 2016-12-28
sys-apps/shadow: update to 4.4 

euank:
This is to enable #2333.

All of the changes in shadow's ebuild are to get it closer to the upstream 4.4 ebuild.

cc @marineam


sys-apps/shadow/shadow-4.4.ebuild
```diff
@@ -117,15 +131,14 @@ src_install() {
 	newins etc/login.defs login.defs
 	dosym ../usr/share/shadow/login.defs /etc/login.defs
 
+	set_login_opt CREATE_HOME yes
```
euank: Note: This is a behavioural difference. The upstream ebuild added it as well.
I'm not sure whether we particularly care whether this is set or not, but happy to remove it or leave it either way.

sys-apps/shadow/shadow-4.4.ebuild
```diff
@@ -117,15 +131,14 @@ src_install() {
...
 	if ! use pam ; then
 		set_login_opt MAIL_CHECK_ENAB no
 		set_login_opt SU_WHEEL_ONLY yes
 		set_login_opt CRACKLIB_DICTPATH /usr/$(get_libdir)/cracklib_dict
 		set_login_opt LOGIN_RETRIES 3
 		set_login_opt ENCRYPT_METHOD SHA512
-
-		# CoreOS: increase the minimum password length to eight
-		set_login_opt PASS_MIN_LEN 8
```
euank: Note: this is a no-op change since we now use pam so this block never executes at all.

sys-apps/shadow/shadow-4.4.ebuild
marineam: If it is a behavior we want it would need to be in PAM since login is only used by getty and not SSH.

euank: I dug into it a bit and it looks like right now it's only used in `useradd` for defaulting `-m` vs `-M`.

There is a line in "TODO" to use it for "newusers", but no code related to that.

PR #2336, opened 2016-12-29
Profile updates/cleanups in preparation for a portage-stable update

marineam:
The only thing particularly interesting here are the seccomp and xattr flags. Upstream has enabled them by default and it seems fair to follow suit. amd64 already has xatter on all packages and seccomp on some. arm64 is the bigger change as it had both completely disabled. The end result is that the big portage-stable update will not toggle any flags.

All the rest of the churn is just trivial cleanups or tweaks to make diffing emerge results before and after the portage-stable update easier.

### DIFF
--- a/profiles/coreos/base/package.use.force
+++ b/profiles/coreos/base/package.use.force
@@ -1,2 +1,5 @@
 # Copyright (c) 2014 The CoreOS Authors. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
+
+# Do not force this flag, we don't need XATTR_PAX
+sys-apps/portage -xattr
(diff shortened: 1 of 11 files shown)


profiles/coreos/base/package.use.force
```diff
@@ -1,2 +1,5 @@
 # Copyright (c) 2014 The CoreOS Authors. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
+
+# Do not force this flag, we don't need XATTR_PAX
+sys-apps/portage -xattr
```
crawford: Why is this needed in both `package.use` and `package.use.force`?

marineam: The package.use.force removes the flag from a previous listing in package.use.force in the hardened profile as that overrides everything. The entry in package.use then can actually disable the flag since it is listed in $USE. The forced flag mechanism is basically an independent system from normal $USE and package.use system.

PR #2350, opened 2017-01-05
app-emulation/docker: remove early docker

crawford:


philips:
This should be reverted. Early-docker is crufty but we should deprecate it on a planned rollout. https://twitter.com/BenReser/status/837699258971107328

crawford:
This was announced five months ago (https://groups.google.com/d/msg/coreos-user/DqOcGM9EXzo/HLnyYLLiCAAJ).

PR #2353, opened 2017-01-05
sys-kernel/coreos-*: bump to 4.8.15

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/1703 and https://github.com/coreos/bugs/issues/1730.

PR #2355, opened 2017-01-05
Revert "Active directory"

crawford:
Reverts coreos/coreos-overlay#2325


crawford:
Triggers https://security.gentoo.org/glsa/201701-03 and needs https://github.com/coreos/coreos-overlay/pull/2354.

PR #2357, opened 2017-01-06
dev-lang/python-oem: hardcode distname to coreos

crawford:


euank:
Can we add a comment to the patch  referencing when we can remove it?

crawford:
I added some context to the commit message.

PR #2358, opened 2017-01-06
app-admin/toolbox: bump to swap docker for rkt

dm0-:
This fixes coreos/bugs#1704.

I've added a runtime dependency on rkt for completeness, but I suppose it isn't *strictly* required ... if the user knows how to manually configure and construct an appropriate root directory.


glevand:
This broke toolbox on arm64.  See https://github.com/coreos/rkt/issues/3526.
We can either wait for the rkt fix, or put a temporary conditional in the toolbox script to use docker for arm64.

PR #2359, opened 2017-01-07
sys-kernel/coreos-sources: patch dev ref leak on IPv6 DAD failure

bgilbert:


crawford:
Correct. Can you also open a PR against the build-1235 branch?

PR #2381, opened 2017-01-18
Upgrade to Docker 1.13

dm0-:


app-emulation/docker-proxy/docker-proxy-9999.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2016 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
+EGO_PN="github.com/docker/libnetwork"
+
+COREOS_GO_PACKAGE="${EGO_PN}"
```
marineam: should probably explicitly set the coreos go version too.

dm0-:
Oops, amended the wrong commit.  Updated.

euank:
@dm0- https://github.com/docker/docker/blob/v1.13.0/hack/dockerfile/binaries-commits#L4-L5

We probably need to check if there's anything meaningful there too

PR #2393, opened 2017-01-24
sec-policy/selinux-*: Two small updates

crawford:
Ensure that containers can append to fifos so stderr works under Docker,
and quieten wake_alarm AVCs.


crawford:
Fixes https://github.com/coreos/bugs/issues/1757 and https://github.com/coreos/bugs/issues/1739.

PR #2395, opened 2017-01-25
Increase CONFIG_NR_CPUS to support Intel Xeon Phi

jakub-d:
Fixes https://github.com/coreos/bugs/issues/1771


glevand:
It would be nice to see the increase in kernel size with this change.  If possible, could you post the before and after output of the toolchain's ```size``` utility?  In the SDK that would be ```x86_64-cros-linux-gnu-size```.

glevand:
The upstream defconfig for arm64 does not specify a CONFIG_NR_CPUS value, and so defaults to 64.  I think that, or specifying 128 would be enough for arm64.

jakub-d:
> It would be nice to see the increase in kernel size with this change. If possible, could you post the before and after output of the toolchain's size utility? In the SDK that would be x86_64-cros-linux-gnu-size.

@glevand - against which file you want me to run the `x86_64-cros-linux-gnu-size`?

Here are the Kernel sizes:

```
 # zgrep CONFIG_NR_CPUS /proc/config.gz
CONFIG_NR_CPUS=128
# ls -l /boot/coreos/vmlinuz-a
-rwxr-xr-x. 1 root root 34415576 Jan 25 21:58 /boot/coreos/vmlinuz-a
```

```
 # zgrep CONFIG_NR_CPUS /proc/config.gz
CONFIG_NR_CPUS=512
# ls -l /boot/coreos/vmlinuz-a
-rwxr-xr-x. 1 root root 34923864 Jan 26  2017 /boot/coreos/vmlinuz-a
```

This PR does not bump Kernel revision. I don't know how to do it.

jakub-d:
@glevand-- is there anything I can do to move forward with this PR?

crawford:
@glevand Is there concern around the image size? I'm inclined to just bump it for both platforms, but I could be easily convinced to keep them seperate.

@jakub-d Both coreos-modules and coreos-kernel need to be bumped. `coreos-modules-4.8.17.ebuild` (which is implicitly `coreos-modules-4.8.17-r0.ebuild`) needs to be renamed to `coreos-modules-4.8.17-r1.ebuild`. The same goes for coreos-kernel.

glevand:
@crawford I think it is OK to set them both at 512. We can always set it smaller later.

jakub-d:
@crawford @glevand - thank you for the instructions.  Let me know if the change is now correct.

crawford:
@jakub-d looks good. Can you just update the commit message to match our format? Something like "sys-kernel/coreos-kernel: blah blah".

crawford:
Superseded by https://github.com/coreos/coreos-overlay/pull/2410. I want to get this in to the next Alpha.

PR #2414, opened 2017-02-05
sys-libs/ncurses: add rxvt-unicode-256color

sherter:
Every once in a while, I stumble upon this error message because the terminfo for `rxvt-unicode-256color` is missing in CoreOS:
```
WARNING: terminal is not fully functional
-  (press RETURN)
```

`rxvt-unicode` is a quite popular terminal emulator and, at least on [Arch](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/rxvt-unicode#n32), compiled with [`--enable-256-color`](http://cvs.schmorp.de/rxvt-unicode/README.configure?view=markup#l240). I figured it's worth a try to get this fix upstream instead of working around this problem over and over again...


sherter:
@euank I've changed the filename (this is all I have to do, right?) and squashed the commits.

PR #2416, opened 2017-02-07
app-emulation/containerd: use the upstream default timeout

dm0-:
I think this will be less problematic for drop-in customizations as opposed to completely replacing `ExecStart=`, especially since this command syntax has had backwards-incompatible changes in the past.


app-emulation/containerd/files/containerd.service
```diff
@@ -4,7 +4,7 @@ Documentation=http://github.com/docker/containerd
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/containerd --listen unix:///var/run/docker/libcontainerd/docker-containerd.sock --shim /usr/bin/containerd-shim --state-dir /var/run/docker/libcontainerd/containerd --runtime /usr/bin/runc
+ExecStart=/usr/bin/containerd --listen unix:///var/run/docker/libcontainerd/docker-containerd.sock --shim /usr/bin/containerd-shim --state-dir /var/run/docker/libcontainerd/containerd --runtime /usr/bin/runc $CONTAINERD_OPTS
```
crawford: Why not break all of these flags out?

What do you intend to add to CONTAINERD_OPTS?

dm0-: What do you mean by "break them out"?  Give them separate environment variables?  I didn't have a reason to change them.

My immediate use for this is to set the container start timeout, which is 15 seconds by default, which is not enough time when multiple ARM systems are being emulated on the same host in parallel.

crawford: > What do you mean by "break them out"?

Similar to what we do with Docker: https://github.com/coreos/coreos-overlay/blob/53aa60b0f35744f856860be6794270efd0cb04da/app-emulation/docker/files/docker.service#L14

dm0-:
I've added separate options for the other flags, though I'm not sure how useful that is here since the Docker variables seemed to be intended for flannel to use.  This fixes coreos/bugs#1854.

app-emulation/containerd/files/containerd.service
```diff
@@ -4,7 +4,13 @@ Documentation=http://github.com/docker/containerd
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/containerd --listen unix:///var/run/docker/libcontainerd/docker-containerd.sock --shim /usr/bin/containerd-shim --state-dir /var/run/docker/libcontainerd/containerd --runtime /usr/bin/runc
+ExecStart=/usr/bin/containerd --listen unix:///var/run/docker/libcontainerd/docker-containerd.sock $CONTAINERD_OPTS $CONTAINERD_RUNTIME $CONTAINERD_SHIM $CONTAINERD_START_TIMEOUT $CONTAINERD_STATE_DIR
+
+# define default options that can be overridden
+Environment="CONTAINERD_RUNTIME=--runtime /usr/bin/runc"
+Environment="CONTAINERD_SHIM=--shim /usr/bin/containerd-shim"
+Environment="CONTAINERD_START_TIMEOUT=--start-timeout 2m"
```
crawford: While you are in here, can you fix https://github.com/coreos/bugs/issues/1854?

dm0-: I did.

euank:
This locks us a little more into having docker/containerd as separate units which I'm still not convinced we want.

Are there specific customizations we're looking to allow here (besides fixing the timeout)?

app-emulation/containerd/files/containerd.service
crawford: Wow, I can't read.

crawford:
@euank while I'm not convinced we want it either, is there a downside to this setup?

dm0-:
The only customization I wanted is the timeout, and that is the only one needed for the linked issue.

euank:
@crawford I thought it might make `--live-restore` or `--add-runtime`, but it looks like I'm wrong and those both work fine.

I'm not certain that `live-restore` is working exactly as intended in that I think dockerd can restart, but containerd can't.

app-emulation/containerd/files/containerd.service
```diff
@@ -4,7 +4,13 @@ Documentation=http://github.com/docker/containerd
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/containerd --listen unix:///var/run/docker/libcontainerd/docker-containerd.sock --shim /usr/bin/containerd-shim --state-dir /var/run/docker/libcontainerd/containerd --runtime /usr/bin/runc
+ExecStart=/usr/bin/containerd --listen unix:///var/run/docker/libcontainerd/docker-containerd.sock $CONTAINERD_OPTS $CONTAINERD_RUNTIME $CONTAINERD_SHIM $CONTAINERD_START_TIMEOUT $CONTAINERD_STATE_DIR
+
+# define default options that can be overridden
+Environment="CONTAINERD_RUNTIME=--runtime /usr/bin/runc"
```
euank: This doesn't matter if the user configures the runtime via `--add-runtime / --default-runtime` on dockerd, right?

I don't see why this knob needs to be turned on containerd ever.

dm0-: So it should drop the `--runtime` option completely?

euank: I think so

dm0-: Okay, amended, but untested.

dm0-: All kola tests pass with it.

euank:
What I was trying to say is I'd prefer just bumping the timeout without providing any options yet if we can get away with it to avoid locking us into keeping them split.

dm0-:
The reason I created the PR in the first place is to be able to configure the timeout, but I guess just increasing it is a start.  It's updated to replace `--runtime` with `--start-timeout`.

PR #2419, opened 2017-02-09
Upgrade pesign

dm0-:
This doesn't install pesign into any CoreOS targets by default.  It just makes it available for manually emerging a la `sys-firmware/edk2-armvirt` in Jenkins.

Part of coreos/portage-stable#523.

### DIFF
--- a/profiles/coreos/base/package.accept_keywords
+++ b/profiles/coreos/base/package.accept_keywords
@@ -108,3 +108,6 @@ dev-util/checkbashisms
 
 # xfsprogs 4.5.0 has install conflicts, and versions in between fail to link
 =sys-fs/xfsprogs-4.9.0 **
+
+# pesign has no stable version at the moment, so accept the one with pesigcheck
+=app-crypt/pesign-0.112 **


dm0-:
Ignore this; it was fixed upstream now.

dm0-:
(Repurposing this PR to just accept the new version instead of importing ebuild.)

glevand:
This seems useful enough to put into the SDK.  Can a host version check files of a different arch?

dm0-:
@glevand It should theoretically be able to verify different architectures, but I haven't yet gotten `pesigcheck` to validate anything correctly.  I'm digging through it to find where the actual failure is.

There is another command in the SDK `sbverify --no-verify --cert $pem $efi` that you can use as a substitute for now with any architecture (where `--no-verify` just means you're not forced to add the Secure Boot CA to the system's trusted bundle).  If `pesigcheck` works, we can probably drop `sbsigntool`.

ajeddeloh:
Needs a rebase. Should this be in `profiles/coreos/targets/sdk` instead of `profiles/coreos/base` so it errors if we accidently pull it into the image?

dm0-:
@ajeddeloh Since we're dropping the Secure Boot patches from the kernel, we should probably just close this until it's revisited.

@glevand I have since fixed the `pesigcheck` verification upstream, and I believe I did successfully test it against different architectures at the time.

PR #2426, opened 2017-02-15
sys-kernel/coreos-sources: Add script to revbump and update patches

bgilbert:


bgilbert:
The script should bump the kernel and modules ebuilds too, but that's for another day.

PR #2429, opened 2017-02-18
app-emulation/containerd: fix cache problem

euank:
Prior to this, "${P}" would match upstream gentoo's distfile cache of
containerd, and that tarball would be used regardless of our SRC_URI
changing as we bumped the commit hash.

That resulted in us having an incorrect version of containerd installed
(and lying about the commit hash in --version to boot. Yikes!)

This fixes the ebuild version to capture that it is not a tagged version.


euank:
@marineam it seems like there should be a better way to do this. Is there?

marineam:
The git hash was changed but only -r1 to r2 was bumped? The issue is that -r is the wrong version to bump when the upstream code (e.g. tarball) changes. Tarballs *must* be considered immutable otherwise mirroring and caching are impossible. Normally this all just works out because most ebuilds do not have to rename tarballs but the exception is this sort of vcs-snapshot ebuilds since the upstream tarball doesn't have a sensible name to begin with.

The key to remember -r is for local ebuild revisions while the rest of the version string is the upstream version, which is why -r is left out of $P. To version upstream patch/snapshot/etc changes that don't have a release version the _p suffix can be used. The value can be either just an arbitrary incremental patch version or the date the snapshot is from. So 0.2.5_p20170218-r1 for example.

The fact that we often bump -r for code changes in addition to ebuild changes is more of a fluke of using cros-workon where git is used directly so there isn't a tarball to name. But perhaps we should stop doing that to avoid this sort of confusion in the future. Over the years we have migrated from perpetual -r rev bumping to real version numbers for a number of things but that transition isn't universal yet.

euank:
I picked a `_p` number scheme that makes sense to me for these sorts of ebuilds. PTAL @marineam

euank:
@marineam ping

PR #2430, opened 2017-02-20
Update to glibc 2.23

dm0-:
This updates glibc to the latest Gentoo stable version which fixes the GLSA.


sys-libs/glibc/files/eblits/src_prepare.eblit
```diff
@@ -58,6 +58,8 @@ eblit-glibc-src_prepare() {
 	find . -type f '(' -size 0 -o -name "*.orig" ')' -delete
 	find . -name configure -exec touch {} +
 
+	eprefixify extra/locale/locale-gen
```
marineam: We don't ship locale-gen, but I presume this is just modifying the file in ${S} and doesn't impact the installed system right?

dm0-: Yes, the `eprefixify` function just runs `sed -i -e "s|@GENTOO_PORTAGE_EPREFIX@|${EPREFIX}|g"` on the given files.  That line was added upstream and was left there during the sync.

PR #2439, opened 2017-02-22
Patch the DCCP CVE for beta

dm0-:
Try again.


euank:
@dm0- is "arm64-coreos-verity-hash.patch" intentionally be included here?

bgilbert:
For the record: it's a sync from the v4.8.17-coreos branch of coreos/linux, which has that patch in a previous alpha but not currently in beta.  It's just adding padding in the ARM image and has been fine in alpha, so adding it here for simplicity.

PR #2434, opened 2017-02-21
Fix some recent GLSAs

dm0-:
This is part of coreos/portage-stable#525.

PR #2455, opened 2017-03-02
app-emulation/qemu: sync with latest upstream stable

dm0-:
This is the same commit stolen from #2434 (for a GLSA), but we now know that the failures were from kola using the old `bios.bin` instead of the current default `bios-256k.bin`.


marineam:
Upstream has since stabilized 2.8.0-r3 which includes a new pile of security patches.

dm0-:
I've replaced it with the `-r3` ebuild.

PR #2463, opened 2017-03-07
coreos-base/oem*: update bug report url

crawford:
This changes the bug URL from GitHub to a subdomain under coreos.com.
This will allow us in the future to move to another bug tracker. As it
exists today, this new subdomain just redirects to GitHub so there
should be no change in behavior.


jonboulle:
random drive-by note: you might consider changing links in github.com/coreos/*/README.md too, for the projects that point to github.com/coreos/bugs (not sure how many that affects)

PR #2464, opened 2017-03-08
app-admin/flannel-wrapper: update to 0.7.0 again

euank:
This properly updates the image tag.

This should have happened in #2390 

This'll conflict with #2459; I'll rebased when I need to.


dm0-:
I've merged #2459.  Sorry, I was hoping to get confirmation that it fixed the issue.

euank:
I don't see a reason this really needs to be rushed to beta; it's just making things faster, not more correct.

PR #2469, opened 2017-03-09
dev-libs/openssl: generate /etc/ssl at boot

dm0-:
This fixes a few `openssl` commands such as `req` on PXE at least and takes out the sample CA scripts.

Note that this also wipes out `/etc/ssl` in the SDK (aside from `certs`), but I haven't hit any build failures from it yet.  We might want to handle that case at some point anyway.

### DIFF
--- a/dev-libs/openssl/openssl-1.0.2k-r1.ebuild
+++ b/dev-libs/openssl/openssl-1.0.2k-r1.ebuild
@@ -4,7 +4,7 @@
 
 EAPI="5"
 
-inherit eutils flag-o-matic toolchain-funcs multilib multilib-minimal
+inherit eutils flag-o-matic toolchain-funcs multilib multilib-minimal systemd
 
 MY_P=${P/_/-}
 DESCRIPTION="full-strength general purpose cryptography library (including SSL and TLS)"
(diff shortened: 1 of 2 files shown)


dev-libs/openssl/openssl-1.0.2k-r1.ebuild
```diff
@@ -242,6 +237,12 @@ multilib_src_install_all() {
 	dodir /etc/sandbox.d #254521
 	echo 'SANDBOX_PREDICT="/dev/crypto"' > "${ED}"/etc/sandbox.d/10openssl
 
-	diropts -m0700
-	keepdir ${SSL_CNF_DIR}/private
+	# Don't keep the sample CA files and their ilk in /etc.
```
glevand: `s/ilk/link`?

dm0-: That phrase refers to the other odd scripts installed in `/etc/ssl/misc`, like the ones that just print out fields of a certificate.  This PR is following an offline discussion with @marineam about distributing the CA demo scripts, but I saw no reason to keep any of the other stuff in there.

PR #2470, opened 2017-03-10
eclass/coreos-kernel.eclass: Verify enablement of kconfig options

bgilbert:
If any kconfig options are set to `y` or `m` in the defconfig or commonconfig, but are set to `n` after `make olddefconfig` runs, fail the build. I've verified that this would have caught https://github.com/coreos/bugs/issues/1843.

Fix up affected options, mostly on ARM. Also fix a few warnings from `make olddefconfig`.


bgilbert:
Thanks @glevand. I think those were the only wireless options.

PR #2471, opened 2017-03-10
Properly support nscd

dm0-:
The actual `nscd` binary has been installed since sssd was added, but none of the systemd configuration files were included.  This resulted in a binary that would not start due to a missing `/run` directory.

The upstream `nscd.conf` is installed in `/usr` and linked into `/etc`.


sys-libs/glibc/files/eblits/src_install.eblit
```diff
@@ -166,10 +166,21 @@ toolchain-glibc_src_install() {
 	# Clean out any default configs
 	rm -rf "${ED}"/etc
 
+	cd "${S}"
+
+	# Install misc network config files
+	insinto /usr/share/baselayout
+	doins nscd/nscd.conf || die
```
euank: Any particular reason this should be here vs in coreos/baselayout?

dm0-: It is the unmodified file from the glibc source.  We're not enabling the service by default, so I don't see a need to modify it at the moment.

PR #2473, opened 2017-03-11
Fix SSSD on PXE

dm0-:
This fixes coreos/bugs#1813.

I've added the last commit to make the service behave more like everything else in the OS, but it's not related to the bug and changes behavior.  It can be dropped if we don't want to change it now.


dm0-:
I bumped the ebuild in the first commit in case we'd want to drop the last one.

crawford:
Ninja edit. LGTM.

dm0-:
Alright, its revision is incremented again.

PR #2494, opened 2017-03-22
coreos-base/coreos-init: fix motdgen ordering

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/1774.

PR #2509, opened 2017-04-05
profile: Enable arm64 perl

glevand:
Perl now builds for arm64.


glevand:
Updated `package.accept_keywords` as you suggested.  Seems to build OK.

PR #2513, opened 2017-04-08
coreos-base/coretest: Drop package

bgilbert:
Subsumed by kola.


euank:
cc @marineam if you want to have an opinion

PR #2514, opened 2017-04-10
app-emulation/docker: update to 17.04-ce

euank:
This is the simplest change to get us to 17.04.0.

I manually verified it built and passed kola tests.

I'd like to followup with the additional changes we should implement (ebuild improvements, docker-init working). I think all those changes also applied to 1.13.x as well, and we just didn't get to them then.

If you'd rather me try to get all of them at once instead I can do that.


app-emulation/docker/docker-9999.ebuild
```diff
@@ -27,7 +27,7 @@ SLOT="0"
 IUSE="apparmor aufs +btrfs contrib +device-mapper experimental +overlay seccomp
 	+selinux vim-syntax zsh-completion +journald"
 
-# https://github.com/docker/docker/blob/master/hack/PACKAGERS.md#build-dependencies
+# https://github.com/docker/docker/blob/v17.04.0-ce/project/PACKAGERS.md#build-dependencies
```
coresolve: https://github.com/docker/docker/blob/master/project/PACKAGERS.md?

euank: Want to clarify what you're trying to comment?

euank:
Good catch @dm0-. I'll run this through jenkins to make sure it works on a clean workspace since I think my dirty workspace is what allowed the runc thing to work.

PR #2517, opened 2017-04-12
sys-process/tini: add tini ebuild

euank:
Even though there's an upstream gentoo ebuild, it depends on cmake
(which had trouble cross-compiling for arm64).

tini is a simple enough program that writing our own makefile for it is
pretty easy, and it'll pay off in reduced buildtime and dependencies for
now since we can avoid dragging in cmake.

Note that we don't bother providing the static useflag and just make it
always static since we're packaging this for docker regardless.


sys-process/tini/tini-0.13.2.ebuild
```diff
@@ -0,0 +1,28 @@
...
+HOMEPAGE="https://github.com/krallin/tini"
+SRC_URI="https://github.com/krallin/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="amd64 arm64"
+
+
+src_prepare() {
+	mkdir -p "${S}/src"
```
dm0-: Is this necessary if it's in the source archive?

PR #2511, opened 2017-04-06
Reenable flannel on arm64

glevand:
Also adds missing ebuild metadata.

PR #2520, opened 2017-04-13
Enable etcd-wrapper on arm64.

glevand:
Also adds missing ebuild metadata.

Same ebuild structure as https://github.com/coreos/coreos-overlay/pull/2511 (Reenable flannel on arm64).

To work as-is on arm64 the etcd-wrapper needs https://github.com/coreos/etcd/pull/7714 and an arm64 release to quay.io/coreos/etcd.

For testing, an etcd systemd drop-in with something like this should work:

    [Service]

    Environment="ETCD_IMAGE=docker://quay.io/glevand/etcd:v3.1.0-rc.1-arm64"
    Environment="RKT_RUN_ARGS=--uuid-file-save=/var/lib/coreos/etcd-member-wrapper.uuid --insecure-options=image"
    Environment="ETCD_UNSUPPORTED_ARCH=arm64"


glevand:
@crawford Did you want to rename `etcd-wrapper-0.0.4.ebuild` to `etcd-wrapper-3.0.10.ebuild` so the ebuild package version matches the etcd release?  Then we can use ${PV} in the ebuild, and it makes it easier to know which etcd version the package uses.

crawford:
Seems like a reasonable idea. I'd be fine with it in a separate commit in this PR.

glevand:
Added patch `	etcd-wrapper: Change ebuild version to etcd release`.

glevand:
Rebased to lastest.
Dropped patch `etcd-wrapper: Change ebuild version to etcd release`, which became duplicated in  https://github.com/coreos/coreos-overlay/pull/2549 (app-admin/etcd-wrapper: bump to etcd 3.1.6).

coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -75,9 +75,7 @@ RDEPEND="${RDEPEND}
 	)
 	etcd_protocols_2? (
 		dev-db/etcd:2
-		amd64? (
-			app-admin/etcd-wrapper
-		)
+		app-admin/etcd-wrapper
```
dm0-: The `coreos` ebuild revision number should probably be bumped for this.  (The symlink in the same directory.)  Maybe the `coreos-base/coreos` changes could be in their own commit.

app-admin/etcd-wrapper/etcd-wrapper-3.1.6.ebuild
```diff
@@ -22,8 +22,15 @@ RDEPEND=">=app-emulation/rkt-1.9.1[rkt_stage1_fly]"
 S=${WORKDIR}
 
 src_install() {
+	local tag="v3.1.6"
```
dm0-: Since the ebuild has the same version now, maybe make this value `"v${PV}"` so updates only require a rename with no edits.

glevand:
Used `${PV}` for etcd release tag.
Moved coreos-base/coreos changes to separate patch and bumped ebuild version.

PR #2524, opened 2017-04-18
coreos-metadata: move provider into environment variable

dgonyeo:
By moving specifying how to get the provider into an environment
variable, it can be overridden via an environment varaible (which will
be necessary on openstack).

Related things: https://github.com/coreos/bugs/issues/1917 https://github.com/coreos/container-linux-config-transpiler/pull/58


dgonyeo:
Renamed the symlink due to oob conversation with @crawford

coreos-base/coreos-metadata/files/coreos-metadata.service
```diff
@@ -3,7 +3,8 @@ Description=CoreOS Metadata Agent
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/coreos-metadata --cmdline --attributes=/run/metadata/coreos
+Environment=COREOS_METADATA_PROVIDER_OPTS=--cmdline
```
crawford: Let's call this `COREOS_METADATA_PROVIDER` instead.

dgonyeo:
Fix applied

PR #2526, opened 2017-04-19
Attempt Active Directory again

dm0-:
This is for coreos/bugs#968 and coreos/bugs#1910.  I haven't tested it yet, just fixed the build from the last attempt and added `adcli`.


dm0-:
It worked for me with amd64 and arm64 after checking out this branch.  Did you sync portage-stable since yesterday?

euank:
did a `repo sync` before emerging, I'll do a full `build_packages` pass too.

euank:
Looks like that fixed it; I'm not sure what was wrong with that eclass/portage

PR #2535, opened 2017-04-24
Revert "Merge pull request #2308 from sapcc/master"

euank:
This reverts commit 14be7e884c211d222c168daedfc75487f86d4dfb, reversing
changes made to 144812b4f9e5104b09a5122a49933501308a16b9.

See https://github.com/coreos/bugs/issues/1892

Fixing that issue correctly turns out to be more complicated than expected (requiring patching rkt fly), so let's hold this fix out of stable for one more cycle.

It's better to have the state of the world before this (kubelet-wrapper boots, but doesn't have logs plumbed out unless you manually add mounts) than after (kubelet-wrapper fails to boot if you had configured logs).


euank:
@dm0- autogenerated `git revert ... -m` message, just telling you which git parent it was thinking about when it reverted the merge I think.

PR #2540, opened 2017-04-24
app-emulation/containerd: build with go1.7

dm0-:
This is the Docker 1.12 version of the commit in #2514 for stable.


dm0-:
This fixes coreos/bugs#1909.

PR #2549, opened 2017-04-26
app-admin/etcd-wrapper: bump to etcd 3.1.6

dm0-:
Tests are still running on an image with this, but the service looks okay when run manually.

This fixes coreos/bugs#1935.


dm0-:
It shouldn't hurt.  I've bumped the version to match.

euank:
One random gotcha here is that migrating from 2.x can only be done to 3.0.x, not to 3.1.x

This means any migration instructions we might have specific to Container Linux that refernece etcd-wrapper might need to be updated with an intermediate 3.0.x step

PR #2553, opened 2017-04-27
app-admin/flannel-wrapper: update timeouts & update version

euank:
See https://github.com/coreos/bugs/issues/1833 where this was requested.

This has become more important with the more recent flannel releases
including kubernetes and becoming larger as well.

The value of 5 minutes is arbitrarily chosen as a reasonable increase
over the default 90s.


euank:
Oops, didn't bump the ebuild revision. One sec...

PR #2554, opened 2017-04-28
sys-boot/grub: bump to 2.02

dm0-:
Should something be done for existing SDKs to detect this version is actually newer?


marineam:
Portage understands the _beta suffix so the upgrade should work properly.

dm0-:
Upstream added the subslot, and I left it there when messing with it to see if it would rebuild.  It can be dropped if it's not needed.

PR #2557, opened 2017-05-03
dev-db/etcd: remove etcd 0.4.9

crawford:


coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -100,6 +86,7 @@ RDEPEND="${RDEPEND}
 	)"
 
 RDEPEND="${RDEPEND}
+	amd64? ( app-admin/etcd-wrapper )
```
dm0-: Was this dependent on the architecture before?

coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -9,7 +9,7 @@ HOMEPAGE="http://coreos.com"
 LICENSE="GPL-2"
 SLOT="0"
 KEYWORDS="amd64 arm arm64 x86"
-IUSE="etcd_protocols_1 etcd_protocols_2 selinux"
```
dm0-: The arm64 profile still references `etcd_protocols_1`, which should be dropped.

coreos-base/coreos/coreos-0.0.1.ebuild
dm0-: Actually, as of #2520, it looks like it is intended to be enabled on arm64.

glevand: Yes, we want etcd-wrapper on arm64.  We should get an arm64 quay.io image in the next etcd release.
See https://quay.io/repository/coreos/etcd?tab=tags.

crawford: Ah, I branched before #2520 was merged. I've updated it to unconditionally include etcd-wrapper.

PR #2562, opened 2017-05-08
deprecate dockerd script, misc fixups

euank:
cc @dm0- @crawford 

The motivation for removing the script is twofold. It lets us trust docker's defaulting more (which gets us overlay2 on 13.00+! Woo!). It also simplifies thie stuff a bit


euank:
Reworked the last commit, removing the dockerd script inadvertently would have flipped off the defaulting of `--selinux-enabled` for people who had copied the old docker.service to /etc.

The full explanation is in the commit message.

PR #2564, opened 2017-05-09
app-emulation/docker: update to 17.05

euank:
Waiting for a local build+kola cycle to pass on it still.


euank:
The diff between our version and that version of tini is i386 support in their build scripts, so I'm not concerned about skewing a little there.

PR #2566, opened 2017-05-15
Allow kubelet to load missing kernel modules

databus23:
The `kubenet` network plugin uses `ebtables` to set up some [dedup rules](https://github.com/kubernetes/kubernetes/blob/0480917b552be33e2dba47386e51decb1a211df6/pkg/kubelet/network/kubenet/kubenet_linux.go#L808). `ebtables` tries to automatically load missing `ebt_*` kernel modules when first used. Without access to `/lib/modules` this fails and `kubenet` subsequently fails to ensure the ebtables rules.

This PR replaces #2565


pbx0:
@databus23 Could you change the volume name to `coreos-lib-modules` per github.com/coreos/coreos-overlay/pull/2570

databus23:
@pbx0 done

pbx0:
cc @dm0-  @lucab 

@databus23 Thanks for the quick update! I believe this needs a rebase as well now that some other kubelet-wrapper changes have gone in, if you don't mind.

databus23:
@pbx0 squashed and rebased

pbx0:
ping @euank

PR #2567, opened 2017-05-16
Add the torcx generator

dm0-:
This is the overlay side of moving Docker to torcx.  You can't really do anything with this without the scripts changes to build the torcx images into `/usr/share/torcx/store` and edit their units.  I haven't run complete tests on it yet, but I was able to run Docker containers without issue using both versions (by writing `/etc/torcx/profiles/vendor.json` with Ignition to override the version).

The big problem with this is that there is no backwards compatibility for command paths etc.  I'm not sure how far to go with that.  A dropin for `/etc/profile.d` could just amend `$PATH` for user interactions, but there are probably user services pointing at the majority of the Docker binaries' paths.

There are other minor annoyances like the generator writing errors to the journal every `daemon-reload`.

/cc @lucab @euank @crawford


app-torcx/docker/files/docker-1.12.6-manifest.json
```diff
@@ -0,0 +1,23 @@
...
+            "/bin/docker-containerd-shim",
+            "/bin/docker-proxy",
+            "/bin/docker-runc",
+            "/bin/dockerd",
+            "/bin/runc"
+        ],
+        "service": [
+            "/lib/systemd/system/containerd.service",
+            "/lib/systemd/system/docker.service",
+            "/lib/systemd/system/docker.socket",
```
lucab: I was planning to only have service (and dependency units) in here, but you are right that sockets (and timers and some other things) go under `system/` as well. Do we want to a) split a separate section for sockets b) rename "service" into "system" or something equally generic c) do nothing?

The rationale for having separate sections in this assets manifest is to (possibly later) be able to aggregate metadata about torcx addons without having to unpack every single image.

app-arch/torcx/files/vendor.json
```diff
@@ -0,0 +1,11 @@
+{
+    "kind": "profile-manifest-v0",
+    "value": {
+        "images": [
+            {
+                "name": "docker",
+                "reference": "17.05.0"
```
lucab: We probably want to drop the final `.0` in there, or at least symlink this into `17.05` and/or some well known reference like `com.coreos.cl`.

This is part of a larger discussion about user-facing guarantees though, so perhaps we can defer till beta-promotion.

/cc @euank @crawford @philips

lucab:
Great! I've left a couple of comments for future items and captured some of your comments above: https://github.com/coreos/torcx/issues/51 https://github.com/coreos/torcx/issues/52 https://github.com/coreos/torcx/issues/53

Migration items that I've seen so far includes:
 * docker binaries in PATH under `/usr`: I guess those can be replaced with a symlink (or a script, not my preference) pointing to the torcx-ified ones
 * docker users/groups: those can be probably moved to sysusers.d, but I haven't investigated yet the implications of this

app-torcx/docker/files/docker-1.12.6-manifest.json
dm0-: Ignition configuration uses the term "units" for this to cover specifying any of `.service`, `.socket`, `.mount`, etc.  I think that would work here, too.

app-arch/torcx/files/vendor.json
dm0-: Okay, it makes sense to set the vendor reference to `com.coreos.cl` which would be a symlink to `17.05` for alpha.  Then when it rolls to beta, we can just point the symlink at `1.12` instead for Kubernetes support, for consistency with what we're doing now.

dm0-: Noting that image symlinks don't work.  I get `apply failed: image docker:com.coreos.cl not found`.  It does work if I just copy the file to have the same name.

dm0-:
I've updated this to use `com.coreos.cl` refs in `vendor.json`, to drop the least significant version numbers from the Docker packages, and to add compatibility wrapper scripts in `/usr` paths.  I decided to go with wrapper scripts over symlinks because it allows somewhat better error messages for why commands fail to execute, and because `TORCX_BINDIR` is supplied as a runtime variable that otherwise would need to be hard-coded into read-only `/usr` with symlinks.

app-torcx/docker/files/docker-1.12.6-manifest.json
lucab: Agreed, Iet's go for `units`.

lucab:
> I decided to go with wrapper scripts over symlinks because it allows somewhat better error messages for why commands fail to execute, and because TORCX_BINDIR is supplied as a runtime variable that otherwise would need to be hard-coded into read-only /usr with symlinks.

Another (crazy) idea would be having an additional mount unit inside the package to bindmount torcx-docker over usr-docker, taking `TORCX_BINDIR` from the metadata environment file. This still needs a placeholder file in place as a target and I'm not sure all the systemd pieces are in place for this to work, so fine with wrapper scripts now.

dm0-:
As discussed earlier today, this just adds the generator.  I also installed an empty `vendor.json` so errors aren't logged on boot.  With this (and ignoring coreos/scripts#685), torcx just creates its `bin` and `unpack` directories with nothing in them.  I've pushed the previous contents of this PR to a branch `torcx-docker` in case you want to reference using Docker in torcx.

app-arch/torcx/torcx-9999.ebuild
```diff
@@ -0,0 +1,50 @@
...
+
+	exeinto "${generatordir}"
+	newexe "${S}/bin/${ARCH}/torcx" torcx-generator
+#	mkdir -p "${ED}${generatordir}"
+#	ln -fns ../../../bin/torcx "${ED}${generatordir}/torcx-generator"
+	systemd_dounit "${FILESDIR}/torcx.target"
+
+	insinto "${vendordir}/profiles"
+	doins "${FILESDIR}/vendor.json"
+	mkdir -p "${ED}${vendordir}/store"
```
crawford: Why is `ED` used?

dm0-: It was to install it in the buildroot correctly.  I've just changed it to `dodir` instead.

PR #2568, opened 2017-05-16
app-emulation/rkt: bump to 1.26.0

dm0-:


lucab:
As per https://github.com/rkt/rkt/pull/3666, this has a breaking change on duplicate volume names so it would be better to consider adding a `coreos-` vendor prefix to volumes in our wrappers.

PR #2585, opened 2017-05-24
Move Docker to torcx

dm0-:
Since we've now built an alpha with torcx and no profiles or images, this PR (together with coreos/scripts#685) should satisfy the second stage of coreos/torcx#47: shipping Docker images with an empty profile.  As is, it bundles both Docker 1.12 and 17.whatever despite 17 still being present in the alpha image.

/cc @lucab @euank


dm0-:
Since it was decided to not include this in the previous alpha, I'm repurposing the PR to finish the Docker migration again.  I'd like to bump the torcx version to a new alpha tag before merging.

euank:
I did some manual testing. I can't figure out any usecase that breaks with this.
All the kola tests pass still happily enough.

Let's do this! 🌟 🚀 🚫 (combining character) 🐳

PR #2593, opened 2017-06-03
profiles: build lz4 support into journald

dm0-:
This fixes coreos/bugs#1988.


euank:
The one concern to mention with this:

I believe with +LZ4 systemd actually defaults journals to lz4, so the logs will start being written in a different format.

This could hit people who roll back and then try to read their logs.

That being said, I think this is enough of an edge case that we can not worry our heads about it.

lucab:
This may also impact rkt, we hit something similar in the past https://github.com/rkt/rkt/issues/3104.

PR #2594, opened 2017-06-03
app-admin/etcd-wrapper: bump to 3.1.8

dm0-:
My SDK is tied up at the moment so I didn't run tests against an image with this version yet, but it at least produces the correct unit.


dm0-:
So there isn't an arm64 container for 3.1.8, but there isn't one for 3.1.6 either.  The first supported versions seem to be the 3.2 RCs.  I'll merge this without the arm64 version working out of the box, since it didn't appear to be working previously.

PR #2602, opened 2017-06-14
coreos-base: Add oem-vagrant-virtualbox

AlexNPavel:
This commit adds the new oem-vagrant-virtualbox provider, which is a
stripped down version of oem-vagrant for use with virtualbox and ignition.
It also sets the oem id to virtualbox instead of vagrant so it can be
handled correctly by ignition.


coreos-base/oem-vagrant-virtualbox/oem-vagrant-virtualbox-0.0.1.ebuild
```diff
@@ -0,0 +1,24 @@
...
+
+# no source directory
+S="${WORKDIR}"
+
+src_install() {
+	insinto "/usr/share/oem"
+	doins -r "${FILESDIR}/box"
+	doins "${FILESDIR}/grub.cfg"
+
+	into "/usr/share/oem"
```
dm0-: Is this supposed to affect something?  If not, drop it.

PR #2605, opened 2017-06-16
sys-kernel/bootengine: update to include vfat and networkd changes

dgonyeo:


dgonyeo:
I lack write access to merge this

PR #2628, opened 2017-06-26
Sync with Packet provisioner

bgilbert:
Move configuration currently performed by the Packet provisioner into the OEM.  This set of PRs implements things a little differently than the provisioner, but the result should be equivalent.

There are a few other pieces:
- https://github.com/coreos/coreos-metadata/pull/51
- https://github.com/coreos/bootengine/pull/119
- https://github.com/coreos/baselayout/pull/66

This PR will be updated with revbumps after those PRs merge.


coreos-base/oem-packet/files/units/packet-phone-home.service
```diff
@@ -6,8 +6,9 @@ After=coreos-metadata.service
 
 [Service]
 EnvironmentFile=/run/metadata/coreos
-Type=oneshot
```
crawford: Does anything order itself `After` this service? The change from `oneshot` to `simple` will change when the dependent service runs.

coreos-base/oem-packet/files/units/packet-phone-home.service
```diff
@@ -1,13 +1,14 @@
...
 ConditionFirstBoot=true
 Requires=coreos-metadata.service
 After=coreos-metadata.service
 
 [Service]
 EnvironmentFile=/run/metadata/coreos
-Type=oneshot
-ExecStart=/usr/bin/curl --request POST "${COREOS_PACKET_PHONE_HOME_URL}"
+ExecStart=/usr/bin/curl --header "Content-Type: application/json" --request POST "${COREOS_PACKET_PHONE_HOME_URL}"
+Restart=on-failure
```
dm0-: I think this will only fail and restart when it is unable to reach the server.  Do HTTP response codes matter?  If so, `curl` should have `--fail`.

coreos-base/oem-packet/files/units/packet-phone-home.service
bgilbert: Not in the OS, and I wouldn't expect in user configs either.  It's a leaf service.

coreos-base/oem-packet/files/units/packet-phone-home.service
bgilbert: I'm not sure.  The script installed by the Packet provisioner doesn't have `--fail`.

bgilbert:
Rebased and revbumps added.

PR #2645, opened 2017-07-05
app-emulation/rkt: ship upstream API units (disabled)

dm0-:


lucab:
I guess they can be shipped disabled just to make sure they get automatically updated on new release, but yes I don't think there will be much usecases for those in the feature and you may have to deprecate them again soon.

I don't think it is worth shipping them in the OS, unless somebody explicitly asked for them or you are fine with a sudden deprecation/removal in the next future.

dm0-:
Closing since the k8s warnings about these are apparently supposed to be ignored.

PR #2648, opened 2017-07-12
profiles: upgrade to dracut 045

dm0-:
Part of coreos/portage-stable#563.


dm0-:
I've added a placeholder commit for when coreos/bootengine#120 is merged, for amending purposes.

PR #2651, opened 2017-07-14
Upgrade Linux in build-1409 to 4.11.11

coreosbot:
From: https://jenkins-os.prod.coreos.systems/job/os/job/update-linux/1/


crawford:
Tests passed.

PR #2654, opened 2017-07-20
Add meson

dm0-:
This is an intermediate step that just makes Python 3 and meson available to the SDK.  Migrating systemd can happen later.

Part of coreos/portage-stable#564.


dm0-:
This enables Python 3 support for nearly everything other than portage, which has bootstrapping problems during the transition.  Python 2 support is still built for everything, and it is used as the default interpreter.

PR #2656, opened 2017-07-20
sec-policy/selinux-virt: add spc_t type

euank:
See http://danwalsh.livejournal.com/74754.html

People expect spc_t to work. I think this accomplishes that.

I'll admit my understanding is a bit shaky, but since our selinux policy
is already fairly permissive/small in scope, I think this type is
suitably privileged to roughly match the expectations of spc_t.


euank:
Abandoning, Container Linux's selinux needs more elaborate help, which will presumably resolve this.

PR #2665, opened 2017-07-27
sys-firmware/edk2: New package

glevand:
Create a new EDK2/UEFI ebuild sys-firmware/edk2 to replace the current
per-arch edk2-armvirt and edk2-ovmf ebuilds.  Updates to use the latest
EDK2 UDK2017 and openssl-1.1.0f releases.

For amd64 builds UDK2017 requires an update to nasm-2.10 or greater.

Fixes UEFI startup problems like these when running Container Linux in
a KVM guest on arm64 machines:

  Synchronous Exception at 0x00000000BF033004

Depends on https://github.com/coreos/portage-stable/pull/566 (bump(dev-lang/nasm): sync with upstream).


glevand:
Incremented revisions as requested.

PR #2703, opened 2017-08-11
app-emulation/containerd: add restart directive to unit

ajeddeloh:
Containerd can crash and should be restarted if it does. Kola test: https://github.com/coreos/mantle/pull/671


euank:
@ajeddeloh You'll also want to bump the ebuild revisions, otherwise a cached binpackage with the old service file could be installed.

Pretty much any change to an ebuild should have an associated version/revision bump.

ajeddeloh:
Fixed

app-torcx/docker/docker-17.05.ebuild
```diff
@@ -12,7 +12,7 @@ KEYWORDS="amd64 arm64"
 # Explicitly list all packages that will be built into the image.
 RDEPEND="
 	=app-emulation/docker-17.05.0
-	=app-emulation/containerd-0.2.7_p2
+	=app-emulation/containerd-0.2.7_p2-r1
```
bgilbert: @dm0-, does this require the docker ebuild to be bumped?

ajeddeloh: I think it _should_ but it looks like since this is getting built for torcx and since the torcx packages seem to be built every `build_packages` run it wasn't causing issues.

dm0-: I think it should update `app-emulation/docker`, not `app-torcx/docker`.  Updating the torcx package version would require a separate PR for the scripts repo.

euank:
Since both `containerd` ebuilds reference that service file, and both want this change, both should have their revision bumped.

docker 1.12 and 17 also need to be bumped. I agree with moving the `torcx` bump to a different PR.

dm0-:
I wasn't suggesting moving the torcx bump to a different PR; I was suggesting not bumping the torcx package(s).  It's not really important since torcx packages are always built from source and not really installed in the image from portage's point of view, and keeping the versions in sync between the scripts and overlay repos seems error-prone.  Although I've made the torcx build script drop revision numbers from torcx image names so users shouldn't be able to tell the difference, so it probably doesn't matter.

ajeddeloh:
@dm0- you mean not bumping the app-torcx/docker ebuild revision, yes? We still need to bump the contents of the app-torcx/docker ebuild since the containerd need to be bumped for the app-emulation/docker ebuilds (which also should be bumped, yes?)

dm0-:
Yes, what you have now looks correct, except the dependencies need to be updated in the torcx ebuilds since they use exact versions.

ajeddeloh:
Updated dependencies

dm0-:
It is missing the `-r1` for 17.05.

ajeddeloh:
Added -r1 for app-emulation/docker-17.05, updated app-torcx/docker ebuild dependencies to point to correct app-emulation/docker versions

euank:
One more nit, the commit prefix should be `app-emulation/containerd: ` (per the git log and upstream/gentoo way of doing things).

bgilbert:
Per discussion with @euank, this seems pretty harmless, and additional testing is unlikely to uncover much.  Merging for 1506.

PR #2707, opened 2017-08-12
sys-cluster/ipvsadm: add package to coreos

ajeddeloh:
Depends on https://github.com/coreos/portage-stable/pull/575


coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -164,4 +164,5 @@ RDEPEND="${RDEPEND}
 	sys-libs/timezone-data
 	sys-process/lsof
 	sys-process/procps
+	sys-cluster/ipvsadm
```
dm0-: This list is alphabetical.

ajeddeloh:
Review bump. ipvsadm didn't have the arm64 keyword at all but it appears to build and run.

euank:
wrt the "appears to build and run", can we get a kola test alongside this to codify that?

ajeddeloh:
Fixed build issues. Builds without `libnl` or `ipvsadm` being installed locally in the SDK on both arm64 and amd64.
@euank What were you looking for in a kola test? Just that it runs at all or something more exhaustive?

euank:
@ajeddeloh I think we should get into the habit of adding "smoke test" like kola tests for new binaries.

So yes, verify basic functionality. In this case, just adding, listing, and then deleting a few ipvsadm "services" would be a good minimum check to have.

ajeddeloh:
Added INSTALL_MASK and PKG_INSTALL_MASK for libnl

coreos/config/env/dev-libs/libnl
```diff
@@ -0,0 +1,2 @@
+INSTALL_MASK+=" /etc/libnl/ /etc/libnl/classid /etc/libnl/pktloc"
+PKG_INSTALL_MASK+=" /etc/libnl/ /etc/libnl/classid /etc/libnl/pktloc"
```
dm0-: Can you just add `/etc/libnl` to `profiles/coreos/targets/generic/make.defaults`?  It's better to keep these centralized so we don't lose track of them.

ajeddeloh:
Fixed.

profiles/coreos/targets/generic/make.defaults
```diff
@@ -33,6 +33,7 @@ INSTALL_MASK="${INSTALL_MASK}
   /etc/sudoers
   /etc/wgetrc
   /etc/xinetd.d
+  /etc/libnl
```
dm0-: This list is alphabetical.  (Or it should be ... `lvm` does not go before `logrotate`.)

PR #2712, opened 2017-08-15
coreos-base/oem-vagrant-virtualbox: drop box files

bgilbert:
They were never actually shipped in the OEM partition: `scripts/build_library/vm_image_util.sh` has always moved them into the `.box` file during the build.

Part of https://github.com/coreos/scripts/pull/729.

cc @AlexNPavel


bgilbert:
Thank you for submitting this PR. Unfortunately, we don't think we'll end up addressing it in Container Linux.

We're now working on [Fedora CoreOS](https://coreos.fedoraproject.org/), the successor to Container Linux, and we expect most major development to occur there instead. Meanwhile, Container Linux will be fully maintained into 2020 but won't see many new features. We appreciate your taking the time to submit this PR and we're sorry that we won't be able to address it.

PR #2713, opened 2017-08-15
sys-fs/multipath-tools: add to sdk dependencies

ajeddeloh:
kola mkimage requires kpartx which is in sys-fs/multipath-tools. The sdk should have the things required for mantle


ajeddeloh:
Fixed. The title line wasn't quite long enough to spell out `sys-fs/multipath-tools` so I dropped the `sys-fs`

PR #2718, opened 2017-08-19
Upgrade to Docker 17.06.1

dm0-:
This is the transition to the new CE structure.  It needs testing (and probably more cleaning in the `docker` ebuild), but I'm posting this early for anyone who wants to try it.


dm0-:
All tests pass with this iteration.

dm0-:
Note there is one minor package difference from this update that I ignored:  Syncing `docker` with the upstream ebuild installs vim syntax files (among other things, but those are caught by `INSTALL_MASK` paths).  We aren't masking `/usr/share/vim`, so they'd make it into the file system, but torcx doesn't include that path in the image.  It will only matter if someone builds an image that drops torcx and adds Docker.

app-emulation/docker/docker-9999.ebuild
```diff
@@ -236,24 +233,42 @@ src_compile() {
 		fi
 	done
 
-	go_export
+	pushd components/engine || die
 
-	# verbose building
-	export BUILDFLAGS="-x -v"
+	if use hardened; then
```
dm0-: Previously, it added `-fno-PIC` based on the value of `gcc-specs-pie` instead of the `hardened` flag.  The `gcc-specs-pie` function evaluates to `true` in our SDK, so I don't think we're in a different position.  Our amd64 profile inherits from Gentoo's hardened profile, so it seems these kinds of options are intended to be used.

PR #2719, opened 2017-08-21
profiles: sync bzip2 on arm64 for GLSAs

dm0-:
Part of coreos/portage-stable#576.

### DIFF
--- a/profiles/coreos/arm64/package.accept_keywords
+++ b/profiles/coreos/arm64/package.accept_keywords
@@ -2,6 +2,7 @@
 # Keep these in alphabetical order.
 
 =app-admin/sudo-1.8.20_p2 ~arm64
+=app-arch/bzip2-1.0.6-r8 ~arm64
 =app-crypt/mit-krb5-1.14.2 ~arm64
 =app-text/asciidoc-8.6.9-r3 ~arm64
 =dev-cpp/gflags-2.1.2 ~arm64


profiles/coreos/arm64/package.accept_keywords
```diff
@@ -2,6 +2,8 @@
 # Keep these in alphabetical order.
 
 =app-admin/sudo-1.8.20_p2 ~arm64
+=app-arch/bzip2-1.0.6-r8 ~arm64
+=app-arch/unrar-5.5.8 ~arm64
```
euank: Given `unrar` isn't emerged in the sdk or image, do we need to include this at all yet>

dm0-: It is not used by anything we ship right now, so it doesn't really matter.  It had an accepted ebuild for arm64 before the update, so that line is just to keep compatibility.

euank: I'd personally probably not accept it until needed; including it here makes it seem intentional and imply it might be used somewhere.

glevand: I'd say leave `unrar` line out.  If the package is removed this would need to be cleaned up, which seems would be easy to forget.

dm0-:
If we don't want unrar here, there's not much sense in keeping the ebuilds at all: coreos/portage-stable#577.

euank:
Good point, thanks!

PR #2720, opened 2017-08-21
coreos-base/coreos: use etcdctl v3

euank:
See the individual commit messages.

This is possible as a result of https://github.com/coreos/coreos-overlay/pull/2557


euank:
Existing etcd tests pass, which has coverage for `etcdctl cluster-health` (and exactly zero other etcdctl commands..)

I'm writing a few more kola tests around this before merging I think.

euank:
I'm gonna go ahead and merge this for the nightly coverage.
I claim kola test will still be added before the release 👼

PR #2730, opened 2017-08-25
sys-kernel/coreos-*: bump to 4.13-rc6

bgilbert:
4.12 will be EOL by the time CL 1520 reaches stable, so let's start working toward 4.13.

defconfig changes:

- LOCKUP_DETECTOR was split into HARDLOCKUP_DETECTOR and SOFTLOCKUP_DETECTOR in torvalds/linux@05a4a95279311c3a4633b4277a5d21cfd616c6c7.

- CIFS_SMB2 was made mandatory (under CIFS) in torvalds/linux@2a38e12053b760a8f5e85030eb89512660077c15.


sys-kernel/coreos-sources/coreos-sources-4.12.8.ebuild
```diff
@@ -3,18 +3,29 @@
 
 EAPI="5"
 ETYPE="sources"
+
+# -rc releases should be versioned L.M_rcN
+# Final releases should be versioned L.M.N, even for n == 0
+
+# Only needed for RCs
+K_BASE_VER="4.12"
```
dm0-: If this is only for RCs (as `kernel-2.eclass` seems to indicate), can the value just be `${PV%%_*}` so we don't have to edit this file for updates?

dm0-: Ah, nevermind, I see that this same 4.12 version is carried over for 4.13 for the tar version.

PR #2731, opened 2017-08-25
profiles: accept awscli for the sdk

euank:
Given our tooling, such as kola and ore, interacts with AWS, it's also
handy to have the awscli laying around to do things like emergency
cleanup, misc bash scripting, and other poking around.

This does add the yet-unused `profiles/coreos/targets/sdk/package.accept_keywords` file because that seemed like an appropriate place to accept the aws cli. Maybe there's a reason it wasn't used or that I should prefer something else?

Part of https://github.com/coreos/portage-stable/pull/580


euank:
Updated to be just the profile change

PR #2734, opened 2017-08-26
Fix the net-tools install race

dm0-:


bgilbert:
It looks like we can avoid the patch by having the ebuild make `update` instead of `install`.

dm0-:
The patch was sent upstream.

PR #2741, opened 2017-09-05
app-emulation/docker: build reproducibly

euank:
The docker client and engine both include a 'BuildTime' variable set in
their build scripts.
Overriding that to a consistent value is sufficient for them to build
reproducibly as best I can tell.

This CLI's build scripts have a mechanism for doing this. The engine has
an upstream patch (included starting with 17.07) that allows doing the
same.

This modifies the build to apply the above build patch, and set a build
time for both.

It's expected that the build time will be set by the ebuild author each
time the ebuild is modified, thus turning the 'build time' output to
really be the 'package created time', which I think is a reasonable
difference.


app-emulation/docker/docker-9999.ebuild
```diff
@@ -198,13 +203,21 @@ pkg_setup() {
 	enewgroup docker
 }
 
+src_prepare() {
```
dm0-: Can't this just have a `PATCHES` array instead of adding a `src_prepare` function?

euank:
@dm0- updated

PR #2742, opened 2017-09-05
app-backup/casync: add initial ebuild

euank:
Part of the torcx packaging work, this will be used by changes in scripts to determine a torcx package's digest.


app-backup/casync/casync-2.ebuild
```diff
@@ -0,0 +1,34 @@
...
+SRC_URI="https://github.com/systemd/casync/archive/v${PV}.tar.gz"
+
+inherit meson eutils
+
+LICENSE="LGPL-2.1"
+SLOT="0"
+KEYWORDS="amd64"
+IUSE="fuse selinux doc"
+
+DEPEND="app-arch/zstd
```
dm0-: We don't seem to be carrying this.

app-backup/casync/casync-2.ebuild
```diff
@@ -0,0 +1,34 @@
...
+fuse? ( sys-fs/fuse )
+"
+RDEPEND="${DEPEND}"
+
+src_prepare() {
```
dm0-: This function should be replaceable with a `PATCHES` array.

app-backup/casync/casync-2.ebuild
```diff
@@ -0,0 +1,34 @@
...
+	eapply_user
+}
+
+src_configure() {
+	local emesonargs=(
+		-Dfuse=$(usex fuse true false)
+		-Dselinux=$(usex selinux true false)
+		-Dman=$(usex doc true false)
+	)
+	meson_src_configure
```
dm0-: I wasn't sure if this was how the function was intended to be used, but I see other Gentoo ebuilds doing this, so it should be fine.  From the way the `meson_src_configure` function is defined, if `emesonargs` is defined in the global scope, the `src_configure` definition could be omitted.  I suppose from the convention of capitalized variables in the global scope, this is correct, though.

app-backup/casync/files/fix-selinux-build.patch
```diff
@@ -0,0 +1,126 @@
+From ab7ef83b5aef07af6e8628eeab353a3fb043fadd Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Zbigniew=20J=C4=99drzejewski-Szmek?= <zbyszek@in.waw.pl>
+Date: Wed, 26 Jul 2017 09:52:40 -0400
+Subject: [PATCH 1/2] Fix build without selinux and refuse --with=selinux when
```
dm0-: Just to verify, there is no intended patch `2/2`?

euank: I got this from https://github.com/systemd/casync/commit/ca268c88cc225a33be569ccabab7292e71576935.patch

Since the PR has a second patch fixing a segfault, github generated a `1/2` subject, but I'm fine not picking up 2/2 until the next casync release since it's in an area of the code we don't use currently.

app-backup/casync/casync-2.ebuild
euank: oops, I have a portage-stable pr I forgot to make.

euank:
Updated

PR #2745, opened 2017-09-07
sys-kernel/coreos-modules: enable FORTIFY_SOURCE

bgilbert:
Enable some buffer overflow checking added in torvalds/linux@6974f0c4555e285ab217cee58b6e874f776ff409.


bgilbert:
The change adds 40 KiB to the amd64 vmlinuz image, presumably due to inlining, and 17 KiB at runtime in qemu (excluding modules).  I haven't benchmarked, and it could conceivably affect really hot paths, but: it's only adding a test or two to str*/mem* calls that operate on a constant buffer, and those tests should branch-predict well.

PR #2751, opened 2017-09-08
Update to Docker 17.06.2

dm0-:


app-emulation/docker/docker-9999.ebuild
```diff
@@ -19,7 +19,7 @@ else
 	else
 		MY_PV="$PV-ce"
 	fi
-	DOCKER_GITCOMMIT="874a737"
+	DOCKER_GITCOMMIT="cec0b72"
 	# Unix timestamp to use for this build. Set to `date +%s` each time an ebuild
 	# is edited.
 	# This is required for a reproducible build
```
euank: Missing the 'BUILD_DATE' bump below. Sorry :(

@bgilbert suggested that maybe grabbing the timestamp of a file in docker itself (maybe the mtime of the `VERSION` file?) could be better.

dm0-: Does that actually need to be updated for anything other than cosmetic reasons?  Regardless, I've changed it to use the tar file mtime.

euank: It's cosmetic only

PR #2761, opened 2017-09-12
Support building Rust packages in the SDK

dm0-:
I'm making this to run test Jenkins builds.  The final version should be the same with the test package dropped.  Actual Rust packages can be PRed separately.

The Gentoo rust package supports the LLVM cross-compiler targets we need, and the rustlib package added here provides the cross-compiled std crate.  (I've made @sdemos the author of that commit since he wrote the initial version.)  It blocks the rust package to avoid building the compiler for the different architectures.

The rustlib package should match the version of the rust package.  We could enforce this with `EAPI=5-hdepend`, but that would require reimplementing the `cargo` eclass in the ebuild since it only supports `EAPI=6`.  For now, just remember update both packages together.

Depends on coreos/portage-stable#578.


profiles/coreos/base/make.defaults
```diff
@@ -35,7 +35,7 @@ USE="${USE} -openmp"
 POLICY_TYPES="targeted mcs mls"
 
 # Disable packages or optional features with distribution issues.
-ACCEPT_RESTRICT="* -bindist -mirror"
```
crawford: Should this be on `profiles/coreos/sdk/make.defaults` instead?

dm0-: This changes it everywhere.  We could lift the restriction only in the SDK since cargo shouldn't be installed anywhere else, but is there a reason to?  It looked like the intent of restricting this originally was just to get our attention.  There doesn't seem to be any affect on what actually gets mirrored.

dm0-:
A nightly compiler shouldn't be used anywhere.  The `rustc` program is taken from the unmodified Gentoo ebuild, and the same version is used for the libstd source.

sdemos:
I found the environment variable that the compiler uses to cheat when it's bootstrapping itself so that they can use `#!feature`s in the std libraries (it's `RUSTC_BOOTSTRAP`, it's checked [here](https://github.com/rust-lang/rust/blob/master/src/libsyntax/feature_gate.rs#L1681) in the rustc source code). The rust std library ebuild just sets that instead of having to use a nightly build for anything.

dm0-:
The test build succeeded, and I verified the cross-compiled `marker` program works in the built prod images on both architectures.  I've dropped the test package from this PR and updated metadata so it should be ready for merge.

PR #2784, opened 2017-09-22
app-arch/torcx: update to 0.1.2

bgilbert:
Add profiles for /etc/coreos/docker-1.12 flag file.

https://github.com/coreos/torcx/pull/90


bgilbert:
@euank Yep, that's coming next.

PR #2787, opened 2017-09-22
app-emulation/docker: apply ebusy overlayfs patch

euank:
See coreos/bugs#2127 and other related issues.

The docker ebuild's  patch files have been split into more folders, plus some manual eapply
calls, to allow moby/moby patches to be used unmodified against
docker-ce as well.


app-emulation/docker/files/patches/engine/revert-make-overlay-home-dir-private.patch
```diff
@@ -0,0 +1,111 @@
...
+ 	if err != nil {
+ 		return nil, err
+@@ -227,11 +222,10 @@ func (d *Driver) GetMetadata(id string) (map[string]string, error) {
+ 	return metadata, nil
+ }
+ 
+-// Cleanup any state created by overlay which should be cleaned when daemon
+-// is being shutdown. For now, we just have to unmount the bind mounted
+-// we had created.
++// Cleanup simply returns nil and do not change the existing filesystem.
```
arithx: Should `do` be `does`?

euank: Seems like a comment for the upstream PR. I didn't do any editorializing of my revert to try and make it easier to review / in-line with the commit message, but I could toss another commit on over there to fix the spelling.

app-emulation/docker/docker-9999.ebuild
```diff
@@ -212,6 +215,9 @@ src_unpack() {
 		DOCKER_GITCOMMIT=$(git -C "${S}" rev-parse HEAD | head -c 7)
 		DOCKER_BUILD_DATE=$(git -C "${S}" log -1 --format="%ct")
 	fi
+	pushd "${S}"/components/engine || die
```
dm0-: Does `eapply` support `-d` to specify the directory?  Might be cleaner.

euank: It took a few tries, but yes it does, so long as you don't have a space... and it didn't work after the `--` either. Weird.

euank:
For the record, I'm not totally thrilled about merging/shipping patches that haven't been properly reviewed upstream yet.
I'm merging now to give these coverage in our nightlies, and hopefully upstream will have had a look before the alpha release.
If not, I still think i'd be okay shipping these based on my understanding of things, it's just a bit less than ideal.

PR #2790, opened 2017-09-25
Enable SELinux on arm64

dm0-:
I'm creating these PRs to build test images.  Just about all of these packages could use updating, but this just attempts to get them working as is first.


dm0-:
The current state of this is that everything builds, the userland tools work, and all `kola` tests pass on arm64.  The policy itself may not be built or installed correctly, so it's not restricting everything it should.  The journal says `SELinux: the above unknown classes and permissions will be allowed`.  So, this should be investigated before merging, but the PRs can be used as is for building test images with clean `kola` runs.

euank:
@dm0- we have:

```

dm0-:
Yes, it's a much bigger set.

euank:
@ajeddeloh was looking at possibly updating our amd64 selinux policy. @ajeddeloh, want to review this as well?

ajeddeloh:
Sure, For the commits changing configuration flags (e.g. https://github.com/coreos/coreos-overlay/pull/2790/commits/f722cbcbfe155246bed4492554bd222817275eb5) can you add a commit body about why that change was needed? It would be helpful for when I go through and update these pacakges in the future as well.

glevand:
@ajeddeloh @euank I've already done some work to update the selinux related packages.  Let's try to minimize duplicated work.   See:
https://github.com/glevand/coreos--coreos-overlay/commits/master
https://github.com/glevand/coreos--portage-stable/commits/master

dm0-:
Replaced by #3100 and #3155

PR #2797, opened 2017-09-29
Update Docker

dm0-:
This updates 17.06 to 17.09, and it also adds packages for 17.03 (the latest version supported by Kubernetes 1.8).

An SELinux-related runc patch did not apply, and it was skipped in the interest of getting a buildable test image in a reasonable timeframe.  It should be revisited later to see if it applies.


euank:
This LGTM apart from the selinux thing. I'll investigate that in a bit to see whether I think it's still needed. I'll hold off on a proper approval review until I've had a chance to look into that too.

PR #2802, opened 2017-10-04
app-emulation/rkt: Bump to 1.29.0

dm0-:


app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -17,11 +17,11 @@ if [[ "${PV}" == "9999" ]]; then
...
 else
 	KEYWORDS="amd64 arm64"
-	CROS_WORKON_COMMIT="3abde24bc284b7ded5784c56b4e8184c28999641" # v1.28.1
+	CROS_WORKON_COMMIT="6de500a70706403c8c611d80491aea64019141b0" # v1.29.0
 fi
 
-PXE_VERSION="1235.0.0"
-PXE_SYSTEMD_VERSION="v231"
+PXE_VERSION="1478.0.0"
+PXE_SYSTEMD_VERSION="v233"
```
bgilbert: Why do we want to stay on 233?

dm0-: That is the version in 1478.0.0.
https://github.com/rkt/rkt/blob/v1.29.0/stage1/usr_from_coreos/coreos-common.mk#L12

bgilbert: It may be the other way around: 1478 is the last release with systemd 233.  But okay, if that's what upstream's doing.

lucab: See https://github.com/rkt/rkt/pull/3787 for details.

PR #2808, opened 2017-10-05
sys-kernel/coreos-*: add overlay ebusy fix

euank:
Goes along with https://github.com/coreos/linux/pull/96


euank:
Added that change too

euank:
Thanks for the reminder. Revert PR: https://github.com/coreos/docker/pull/43

euank:
This built + worked locally. Hopefully I didn't miss anything else.

app-emulation/docker/docker-9999.ebuild
```diff
@@ -212,7 +210,7 @@ src_unpack() {
 		DOCKER_GITCOMMIT=$(git -C "${S}" rev-parse HEAD | head -c 7)
 		DOCKER_BUILD_DATE=$(git -C "${S}" log -1 --format="%ct")
 	fi
-	eapply -d"${S}"/components/engine "${ENGINE_PATCHES[@]}"
+	[ -n "${ENGINE_PATCHES}" ] && eapply -d"${S}"/components/engine "${ENGINE_PATCHES[@]}"
```
dm0-: It's minor, but I think I'd rather this test for a non-empty string using the `[*]` subscript, or just test the length of the array.  As is, it references the non-existent first array element (which is an error with `-u`).

PR #2815, opened 2017-10-12
coreos-base/update_engine: tectonic detection

crawford:


bgilbert:
It looks as though we usually tag a release when we want to bump update_engine.

crawford:
Tagged the release.

PR #2820, opened 2017-10-12
coreos-base/{update-ssh-keys, coreos-metadata}: add rust packages

sdemos:
this adds ebuilds which build the rust version of both the `update-ssh-keys` project and the `coreos-metadata` project. they both use the `coreos-cargo` eclass. 

right now coreos-metadata requires you to manually package the update-ssh-keys crate and insert it into your cache, hence WIP. I'm going to wait until coreos/update-ssh-keys#15 and coreos/coreos-metadata#62 get merged to move to the more permanent solution of grabbing update-ssh-keys from a git tag.


coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -1,36 +1,204 @@
...
+
+inherit systemd eutils coreos-cargo
 
+DESCRIPTION="a simple tool for collecting instance metadata from various providers"
+HOMEPAGE="https://github.com/coreos/coreos-metadata"
+KEYWORDS="amd64 arm64"
 LICENSE="Apache-2.0"
 SLOT="0"
-IUSE=""
+VERSION="47390b8f433c7e4a6be305c3460f7f75cfb46105"
```
dm0-: Should we just keep the `cros_workon` setup for this package?  (Or add it for `update-ssh-keys`?)

@euank Do you have a preference what repos are added to the manifest?

sdemos: I tried to add `cros_workon`, but it breaks because both `cros_workon` and the `cargo` eclass try and create a `src_unpack`, and the last one seems to win

sdemos:
I updated the package version for coreos-metadata to `1.0.0`

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
dm0-: I pushed a patch onto this branch that uses `cros_workon` for `update-ssh-keys`.  (Feel free to force-push it off again; I just figured this was the easiest way to demonstrate a patch.)

dm0-:
I've added a commit to build `coreos-metadata` with `cros_workon` as well.  It also builds with `update-ssh-keys` from GitHub instead of having to create a crate manually; that part should be dropped in favor of using the usual crate setup when it's ready.

Note that your `Cargo.toml` still says it's `0.1.0`.  We'll probably want that to match the ebuild.

If we're going to use `cros_workon` for `update-ssh-keys`, that project should be added in the manifest repo.

coreos-base/coreos-init/coreos-init-9999.ebuild
```diff
@@ -3,14 +3,14 @@
...
-CROS_WORKON_PROJECT="coreos/init"
+CROS_WORKON_PROJECT="sdemos/init"
 CROS_WORKON_LOCALNAME="init"
 CROS_WORKON_REPO="git://github.com"
 
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="c48b1fdd52fa78c7e67db33861dabaa93f3e73e1"
+	CROS_WORKON_COMMIT="ed478168f0133cc140baa08d4d77e4333e8991d7"
```
dm0-: This would need to bump the stable ebuild version.  (And obviously reference the merge commit in the real `init` repo when ready.)

coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -110,6 +110,7 @@ RDEPEND="${RDEPEND}
 	coreos-base/coreos-cloudinit
 	coreos-base/coreos-init
 	coreos-base/coreos-metadata
+	coreos-base/update-ssh-keys
```
dm0-: This needs to bump the stable ebuild version.

sdemos:
once the three relevant prs get merged I'll force-push the new commit hashes and point the `CROS_WORKON_PROJECT` back at the coreos org instead of my fork. Other than that, are there any other necessary changes I should include?

dm0-:
Add `update-ssh-keys` to the manifest repo.

We'll also want to reference a tag for `update-ssh-keys` source in the `coreos-metadata` ebuild instead of a branch (or do some other source fetching method), but other than that, I think this should be good.

sdemos:
`update-ssh-keys` has been added to the manifest repo in coreos/manifest#143

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -1,36 +1,222 @@
-# Copyright (c) 2015 CoreOS, Inc.. All rights reserved.
+# Copyright (c) 2017 CoreOS, Inc.. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=5
-CROS_WORKON_PROJECT="coreos/coreos-metadata"
+EAPI=6
+
+CROS_WORKON_PROJECT="sdemos/coreos-metadata"
```
dm0-: `coreos/coreos-metadata`

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -1,36 +1,222 @@
...
 CROS_WORKON_LOCALNAME="coreos-metadata"
 CROS_WORKON_REPO="git://github.com"
-COREOS_GO_PACKAGE="github.com/coreos/coreos-metadata"
-inherit coreos-go cros-workon systemd
 
 if [[ "${PV}" == 9999 ]]; then
```
dm0-: The quotes around `${PV}` can be dropped here.

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -1,36 +1,222 @@
...
-IUSE=""
 
-src_compile() {
-	export GO15VENDOREXPERIMENT="1"
-	GO_LDFLAGS="-X main.version=$(git describe --dirty)" || die
-	go_build "${COREOS_GO_PACKAGE}/internal"
+src_unpack() {
+	cros-workon_src_unpack "$@"
+	coreos-cargo_src_unpack "$@"
+	unpack update-ssh-keys-0.1.0.tar.gz
```
dm0-: Let's make this version a variable defined at the top near the `CROS_` settings.  It appears enough times that updating it will be error prone.  So this line could be e.g. `unpack "update-ssh-keys-${UPDATE_SSH_KEYS_VERSION}.tar.gz"`.  A `grep || die` line could also be added to check that this variable actually matches what is in `Cargo.toml` to make sure required updates aren't missed.

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -1,36 +1,222 @@
...
+}
+
+src_prepare() {
+	default
+
+	sed -i "s;^update-ssh-keys.*;update-ssh-keys = { path = \"${WORKDIR}/update-ssh-keys-0.1.0\" };" Cargo.toml
```
dm0-: Use a version variable here.

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -1,36 +1,222 @@
...
+winapi-build-0.1.1
+winutil-0.1.0
+ws2_32-sys-0.2.1
+xml-rs-0.3.6
+"
+# not listed:
+# update-ssh-keys-0.1.0
+
+SRC_URI="$(cargo_crate_uris ${CRATES})
+https://github.com/coreos/update-ssh-keys/archive/v0.1.0.tar.gz -> update-ssh-keys-0.1.0.tar.gz
```
dm0-: Use a version variable here twice.

coreos-base/coreos-init/coreos-init-9999.ebuild
```diff
@@ -3,14 +3,14 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI="4"
-CROS_WORKON_PROJECT="coreos/init"
+CROS_WORKON_PROJECT="sdemos/init"
```
dm0-: `coreos/init`

sdemos:
@dm0- fixed the things you pointed out in your comments. now that all the prs this depends on are merged, this should be good to go.

coreos-base/coreos-metadata/coreos-metadata-9999.ebuild
```diff
@@ -1,36 +1,228 @@
...
+	coreos-cargo_src_unpack "$@"
+	unpack update-ssh-keys-${UPDATE_SSH_KEYS_VERSION}.tar.gz
+}
+
+src_prepare() {
+	default
+
+	# check to make sure we are using the right version of update-ssh-keys
+	# our Cargo.toml should be using this version string as a tag specifier
+	grep "${UPDATE_SSH_KEYS_VERSION}" Cargo.toml || die "couldn't find update-ssh-keys version ${UPDATE_SSH_KEYS_VERSION} in Cargo.toml"
```
dm0-: The command should probably be something like `grep -q "^update-ssh-keys.*tag.*${UPDATE_SSH_KEYS_VERSION}"` so it matches the line of the expected package.

sdemos: done.

PR #2826, opened 2017-10-17
 coreos-base/coreos: add socat to base image 

euank:
Goad along with https://github.com/coreos/portage-stable/pull/588


euank:
Updated to add socat as well; somehow that slipped my mind before.

euank:
Finally got my sdk happy enough with all the binutils to manually verify this seems happy enough; sneaking this in :)

PR #2828, opened 2017-10-18
Address etcd-member race.

coresolve:
In some cases etcd is able to start before systemd-resolved has
populated the /etc/resolv.conf file. This commit addresses this in a
couple of ways. It adds the --dns=host flag to ensure that the contents
of /etc/resolv.conf remain in sync between the node and the etcd pod. We
also change the dependencies to include network-online.target.


euank:
> It adds the --dns=host flag to ensure that the contents
of /etc/resolv.conf remain in sync between the node and the etcd pod

As a matter of commit message accuracy, we should also correct this.

What the `--dns=host` flag does is create a bindmount of `/etc/resolv.conf` at the time the container is created. This is ont the same thing as keeping them in sync.

systemd-resolved does [atomic renames](https://github.com/systemd/systemd/blob/v235/src/resolve/resolved-resolv-conf.c#L270-L273) of the resolv.conf file it manages.
That means that as soon as resolved updates its idea of `resolv.conf`, the rkt pod will be out of sync (since it will have a bindmoutn of the unlinked previous version until restarted).

The only real way to fix this is to use the stub resolver resolved provides, but doing so is much more involved.
I don't have a better realistic solution than `network-online`.

app-admin/etcd-wrapper/files/etcd-member.service
```diff
@@ -16,7 +17,7 @@ Environment="ETCD_IMAGE_TAG=@ETCD_IMAGE_TAG@"
 Environment="ETCD_NAME=%m"
 Environment="ETCD_USER=etcd"
 Environment="ETCD_DATA_DIR=/var/lib/etcd"
-Environment="RKT_RUN_ARGS=--uuid-file-save=/var/lib/coreos/etcd-member-wrapper.uuid"
+Environment="RKT_RUN_ARGS=--uuid-file-save=/var/lib/coreos/etcd-member-wrapper.uuid --dns=host"
```
euank: If we want to change this by default it should be in the wrapper script, not in the service.

I also don't think we should change it because this flag is deceptive; for stage1-fly it operates almost the same as the default option (bindmount vs copy).
Including this flag implies it does something, when the reality is I don't think it is effectively different in any notable way.

coresolve: I've pulled the --dns=host because it doesn't do what I expected here. 

In my testing I added the arg and then changed the resolvers that were provided by dhcp. I then waited for the new lease to pick up the resolvers. When `--dns=host` is configured I see systemd-resolved complaining with:
```
Oct 19 18:18:12 node001.metal.k8s.work systemd-resolved[9442]: Failed to move private resolv.conf file into place: Device or resource busy
Oct 19 18:18:12 node001.metal.k8s.work systemd-resolved[9442]: Failed to move private resolv.conf file into place: Device or resource busy
Oct 19 18:18:12 node001.metal.k8s.work systemd-resolved[9442]: Failed to move private resolv.conf file into place: Device or resource busy
Oct 19 18:18:12 node001.metal.k8s.work systemd-resolved[9442]: Failed to move private resolv.conf file into place: Device or resource busy
Oct 19 18:18:12 node001.metal.k8s.work systemd-resolved[9442]: Failed to move private resolv.conf file into place: Device or resource busy
```

coresolve:
@dm0- @euank Comments Addressed.

PR #2839, opened 2017-10-20
app-admin/kubelet-wrapper: Specify hosts-entry to mount hosts file

weikinhuang:
This fixes an issue with the `v1.8.x` hyperkube images switching to debian base and missing an internal hosts file. Which causes lookups to `localhost` to fail.

Refs issue: https://github.com/kubernetes-incubator/bootkube/issues/740


weikinhuang:
Quoting @euank from mentioned issue:

> I suspect the root cause of that hosts file difference is this kubernetes change: kubernetes/kubernetes#48535
>
>Rather than bindmounting through `/etc/hosts`, the rkt option `--hosts-entry=host` should copy over the hosts entry you have in `/etc/hosts`.
>
>We'd be happy to take a PR against the kubelet-wrapper in the coreos-overlay repo if you'd like to (do note you'll need to bump the ebuild revision too if you do make such a PR).
>
>Other fixes could be for the upstream hyperkube image to continue including a proper `/etc/hosts` file, or for the kubelet to only use `127.0.0.1` instead of relying on localhost to resolve to it.
>
>I might make that last change when I get time.
>
>I do think that this is not a bootkube issue but rather a kubernetes+hyperkube or container-linux+kubelet-wrapper issue.

weikinhuang:
done

dghubble:
Works for me to address port-forward being broken when running on GCE.

There are more details in the issue @weikinhuang linked, but whether `port-forward` works depends on whether kubelet localhost can be resolved, which varies depending on the network / platform. Some of them mask this issue. Fix LGTM

weikinhuang:
thanks!

euank:
Thank you.

This should end up in the next set of releases; in the meanwhile overriding the option in the unit will be easier, but at least it can just be a temporary hack.

dghubble:
Yeah, there are several places we can patch this in the mean time. On it.

PR #2840, opened 2017-10-20
coreos-base/update-ssh-keys: req newer coreos-init

sdemos:
coreos-init versions older than 0.0.1-r152 still have the
update-ssh-keys shell script which means we the update-ssh-keys package
fails to install. to fix this, we make coreos-base/update-ssh-keys
depend on a new enough version of coreos-init.


sdemos:
reworded the commit message and rebased onto `master`.

euank:
I think this should also bump the revision of the symlink.

sdemos:
bumped the revision version.

PR #2853, opened 2017-10-25
Move iproute2 back to portage-stable

ajeddeloh:


profiles/coreos/base/package.accept_keywords
```diff
@@ -75,3 +75,6 @@ dev-util/checkbashisms
 
 # CVE-2017-1000117
 =dev-vcs/git-2.14.1
+
+# iproute2 4.13 has the patch we used to maintain in the overlay
+sys-apps/iproute2 ~amd64 ~arm64
```
bgilbert: We generally pin to a specific package version.

ajeddeloh:
Pinned to 4.13

profiles/coreos/base/package.accept_keywords
```diff
@@ -75,3 +75,6 @@ dev-util/checkbashisms
 
 # CVE-2017-1000117
 =dev-vcs/git-2.14.1
+
+# iproute2 4.13 has the patch we used to maintain in the overlay
```
bgilbert: A link to the upstream commit would be helpful.

ajeddeloh:
Updated commit messages to have upstream commit, ~arm64. [Upstream commit](https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/commit/?id=d6a4076b6ba6547d7e52c377a7c58c56eb5ea16e) if anyone's curious,

profiles/coreos/base/package.accept_keywords
euank: I think the comment makes more sense not to explain why it's not in overlay, but to explain why we want that version. E.g:

`# 4.13 includes a patch to avoid leaking netns mounts in rkt https://git.kernel.org/pub/scm/linux/kernel/git/shemminger/iproute2.git/commit/?id=d6a4076b6ba6547d7e52c377a7c58c56eb5ea16e`

As-is, it's basically a pointer to dig more vs the actual reason we need to accept it... and I don't think the additional context that it was in overlay for a bit really matters in this file.

ajeddeloh:
@euank lgty?

PR #2859, opened 2017-10-30
readme: update readme to be more useful

ajeddeloh:
- Point to portage-stable for packages we don't change.
 - Point out important packages sdk-depends and coreos

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,32 @@
-This overlay contains CoreOS specific packages and Gentoo packages
+# Overview
+
+This overlay contains Container Linux specific packages and Gentoo packages
 that differ from their upstream Gentoo versions.
+
+See the [portage-stable](https://github.com/coreos/portage-stable) repo
+for packages which do not have Container Linux specific changes.
 
 Licensing information can be found in the respective files, so consult
 them directly. Most ebuilds are licensed under the GPL version 2.
 
 Upstream Gentoo sources: https://gitweb.gentoo.org/repo/gentoo.git
+
+# Important packages
+
+`coreos-base/coreos` is the package responsible for everything that gets
+built into a production image and is not OEM specific.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,7 +1,16 @@
-This overlay contains CoreOS specific packages and Gentoo packages
+This overlay contains Container Linux specific packages and Gentoo packages
 that differ from their upstream Gentoo versions.
 
+See the [portage-stable](https://github.com/coreos/portage-stable) repo
+for packages which we do not have Container Linux specific changes to.
```
bgilbert: `for packages which do not have Container Linux specific changes.`

ajeddeloh:
Fixed, also updated to include `board-packages` which I only just found.

README.md
```diff
@@ -1,7 +1,19 @@
...
+for packages which do not have Container Linux specific changes.
+
 Licensing information can be found in the respective files, so consult
 them directly. Most ebuilds are licensed under the GPL version 2.
 
 Upstream Gentoo sources: https://gitweb.gentoo.org/repo/gentoo.git
+
+`coreos-base/coreos` is the package responsible for everything that gets
+built into an image.
+
```
bgilbert: Maybe also document `coreos-base/coreos-dev`?

README.md
```diff
@@ -1,7 +1,19 @@
...
+`coreos-devel/board-packages` is everything that could be built into a
+development or production image, plus any OEM specific packages.
```
bgilbert: There are a bunch of OEM packages not listed in `board-packages`.  Should we add them?

ajeddeloh: Uh probably? I just found this but that seems to fit. Maybe @dm0- knows better?

ajeddeloh:
Updated to include an overview of more packages, including the distinction between `oem-*` and `coreos-oem-*`.

PR #2867, opened 2017-10-31
coreos-devel: add sdk-extras

ajeddeloh:
Add a metapackage like `sdk-depends` to track useful things we want in portage-stable or coreos-overlay, but aren't actually required for building an image. This list is currently incomplete.

The end goal is for every package in `coreos-overlay` or `portage-stable` to be in, or be a dependency of something in `board-packages`, `sdk-depends`, or `sdk-extras`. This will make maintainence of those repos easier.

### DIFF
--- a/coreos-devel/sdk-extras/sdk-extras-0.0.1.ebuild
+++ b/coreos-devel/sdk-extras/sdk-extras-0.0.1.ebuild
@@ -1,0 +1,45 @@
+# Copyright 2017 The CoreOS Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# This isn't meant to be installed directly. It mostly exists so we can better keep track of
+# dependencies within the SDK.
+
+EAPI=5
+
+DESCRIPTION="Meta ebuild for everything that isn't needed in the SDK, but might be useful"
+HOMEPAGE="http://coreos.com/docs/sdk/"
+SRC_URI=""
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="amd64"
+IUSE=""
+
+RDEPEND="
+	app-admin/python-updater
+	app-backup/casync
+	app-crypt/efitools
+	app-crypt/tpm-tools
+	app-editors/emacs
+	app-editors/nano
+	app-portage/eix
+	app-portage/gentoolkit-dev
+	app-portage/repoman
+	app-misc/screen
+	app-misc/tmux
+	app-text/tree
+	app-text/dos2unix
+	dev-util/cscope
+	dev-util/perf
+	dev-util/strace
+	dev-util/valgrind
+	dev-go/glide
+	dev-go/godep
+	dev-python/awscli
+	sys-apps/ed
+"
+
+# Needed for Jenkins
+RDEPEND+="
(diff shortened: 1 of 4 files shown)


coreos-devel/sdk-extras/sdk-extras-0.0.1.ebuild
```diff
@@ -0,0 +1,35 @@
...
+DESCRIPTION="Meta ebuild for everything that isn't needed in the SDK, but might be useful"
+HOMEPAGE="http://coreos.com/docs/sdk/"
+SRC_URI=""
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="amd64"
+IUSE=""
+
+DEPEND="
```
dm0-: Is there a reason to make any of these build-time dependencies?  It seems like they could all be brought in at install time.

dm0-: Also, lists like this should be alphabetical.

coreos-devel/sdk-extras/sdk-extras-0.0.1.ebuild
```diff
@@ -0,0 +1,35 @@
...
+	dev-python/awscli
+	app-portage/repoman
+	app-portage/eix
+	dev-go/glide
+	dev-go/godep
+	app-misc/screen
+	app-misc/tmux
+	app-text/tree
+	virtual/awk
+	app-editors/nano
```
dm0-: If you're including editors, there is also a usable emacs ebuild.

coreos-devel/sdk-extras/sdk-extras-0.0.1.ebuild
```diff
@@ -0,0 +1,35 @@
...
+	coreos-devel/kola-data
```
dm0-: This is already in `sdk-depends`.

coreos-devel/sdk-extras/sdk-extras-0.0.1.ebuild
ajeddeloh: I'm not sure it really, matters, since this is mostly just to keep track of packages. I'll change it to RDEPEND though, no reason not to.

coreos-devel/sdk-extras/sdk-extras-0.0.1.ebuild
ajeddeloh: It needs `dev-libs/libunistring` (for `gnutls`). I'll add that package to portage stable I guess.

dm0-: Oh, yes, I've installed it with that `USE` flag unset.  It's probably good to get gnutls buildable, though, since we may revisit enabling it for systemd again at some point.

ajeddeloh:
Fixed nits, depends on https://github.com/coreos/portage-stable/pull/603 for emacs stuff

ajeddeloh:
Updated to throw in a couple more things that were in portage-stable we might want to keep around (e.g. tpm-tools). I didn't include `pesign` since it needs use flag changes as well.

PR #2875, opened 2017-11-03
profiles/arm64: remove ref to udev-init-scripts

ajeddeloh:
We do not use udev-init-scripts. Remove configuration for it.

cc @glevand: this was added in b37d1376cd13290ea708bd2e8a17dcce5e87064a. Do you know why? Were we including that package then?


ajeddeloh:
It's not in `@system`. I'm testing the change to remove it (and some other cruft) from `portage-stable` now.

PR #2877, opened 2017-11-04
Update portage to 2.3.8

ajeddeloh:
This should shave some time off the build in exchange for taking up more space.

DO NOT MERGE: still testing to see if this breaks other packages.


profiles/coreos/base/make.defaults
```diff
@@ -94,3 +94,6 @@ PORTAGE_WORKDIR_MODE="0755"
 
 # Don't let GRUB whine about /boot
 DONT_MOUNT_BOOT=1
+
+# Use fast compression for binpkgs since we mostly compress then immediately uncompress them
```
dm0-: That is not true.  Binary packages are always uploaded and downloaded between workspaces in every CI build.  In addition, they are documented as being available to users in their developer images, containers, and SDKs, so it will affect some end-user download experience as well.

I disagree about compression speed being the right priority for binary packages, but not enough to block this if someone else wants to approve it.

ajeddeloh: Hmm. It should be much faster locally (rebuilding things now). Maybe don't change make.defaults by default, but still update portage and document it in sdk-tips-and-tricks? If network bandwidth is an issue and we can figure out how to not break users, switching to xz might also be a good idea.

dm0-: I would be okay with changing the default to prioritize minimal file size, so xz would be okay.  (I believe Fedora still uses xz for RPM payload compression.)

ajeddeloh:
Portage 2.3.8 breaks the autoconf ebuild, due to https://gitweb.gentoo.org/repo/gentoo.git/commit/sys-devel/autoconf?id=a91f8fc1b6a4e72de75e50dc0579141ebb5122f1. I'll update it and see if anything else is broken.

dm0-:
Looking at this a bit, it seems you can just set the environment variable `BINPKG_COMPRESS=lzop` in your SDK so local development uses faster compression.

ajeddeloh:
Yup. I dropped the commit that changes the default. I'll throw up a PR to the docs to document it. I'm still waiting to see if this breaks more packages.

ajeddeloh:
Confirmed that `autoconf` is the only package that breaks with portage 2.3.8 and that updating it fixes the breakage. Got lzop working (with the patches I just pushed). But it also needs a change in `build_library/prod_image_util.h` since we use tar to extract gcc's libraries and it can only autodetect compression based on file extension (ironically, when using the autodetect flag, it fails and tells me what flag to use). On the bright side, with `--jobs 12` build_image takes less than 5 minutes locally.

ajeddeloh:
Should I remove the oldest with this? (keep the number of versions the same by removing one when we add one)?

dm0-:
I don't see why we'd keep any old ones, but I'd be okay merging as is and cleaning next time.

PR #2878, opened 2017-11-06
sys-apps/systemd: "Fix" telinit in the SDK

dm0-:
This appears to be causing hangs on the build server, and it should not be needed in the SDK.  The SDK is just a chroot and so doesn't have an active init process.  If a user of the developer image attempts to upgrade the host glibc, they might need to run this manually.


ajeddeloh:
From slack:
> We shouldn't need it in our build, but maybe it's an indication of something else breaking.

dm0-:
I've replaced the glibc change with a more general telinit fix.

glevand:
Fixed my SDK build breakage.  Thanks.

PR #2881, opened 2017-11-08
coreos-base/update-ssh-keys: bump to v0.1.1

sdemos:
this fixes a problem with a dependency of update-ssh-keys where it
didn't correctly parse authorized_keys files.


coreos-base/update-ssh-keys/update-ssh-keys-9999.ebuild
```diff
@@ -10,7 +10,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ ${PV} == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm64"
 else
-	CROS_WORKON_COMMIT="fcbbf62deb7bbe1df527d71c8b6c0231cb1090cb" # v0.1.0
+	CROS_WORKON_COMMIT="967f65a1a6301949d8a59951d6144048c4061dcb" # v0.1.0
```
bgilbert: `# v0.1.1`

dm0-: We traditionally update the comment to indicate the named tag of this commit.

sdemos:
it will have to update the dependency at some point, yes. however, I don't think it's as much of a problem since this use case is already not supported by all cloud provider metadata services.

PR #2882, opened 2017-11-08
coreos-base/update-ssh-keys: bump to v0.1.1 for 1576

sdemos:
this fixes a problem with a dependency of update-ssh-keys where it
didn't correctly parse authorized_keys files.


coreos-base/update-ssh-keys/update-ssh-keys-0.1.1.ebuild
```diff
@@ -0,0 +1 @@
+update-ssh-keys-9999.ebuild
```
bgilbert: Remember to drop the old symlink.

PR #2886, opened 2017-11-08
Upgrade glibc to 2.25

dm0-:
This renders eblits extinct.


dm0-:
For the record, there are currently two problems I hit in testing this PR.

- Locales were somehow broken, manifesting later in a Python 3 script failing to convert to an ASCII string.
- Installation randomly fails for board installs due to https://bugs.gentoo.org/627378.

The first issue appears to be fixed by just dropping the special handling of `locale-gen`.  I'll look into a workaround for the second.

dm0-:
So I accidentally pushed to this remote URL from the wrong repo, and it seems GitHub permanently closes PRs when that happens.  It says "no common history" with the bad branch, and when I fix it, it says it can't reopen because it was force-pushed.  So opening #3161 instead.

PR #2895, opened 2017-11-14
sys-kernel/arm64_defconfig: Remove CONFIG_CMA

glevand:
CONFIG_CMA is really unneeded, and was causing network problems
on Cavuim machines due to the default CONFIG_DMA_CMA of 16MiB
being too small.

cc @bgilbert


glevand:
Rebased to latest and bumped kernel, modules ebuilds.

PR #2900, opened 2017-11-16
Upgrade Rust to 1.21.0

dm0-:
This accepts the new `virtual/rust` package while unstable instead of changing its `KEYWORDS` so it can be moved back to portage-stable when Gentoo is updated.


lucab:
I rebased and updated my local chroot on top of this (and the linked portage-stable) and can confirm that both updating the toolchain and building target binaries work fine.

PR #2906, opened 2017-11-18
net-fs/nfs-utils: fix rpcbind race

ajeddeloh:
Make rpcbind an After= for nfs-mountd. This is fixed upstream but
hasn't been marked as stable on amd64/arm64 by gentoo. Should fix the
nfs races in kola tests.

### DIFF
--- a/net-fs/nfs-utils/files/nfs-utils-1.3.2-rpcbind-after.patch
+++ b/net-fs/nfs-utils/files/nfs-utils-1.3.2-rpcbind-after.patch
@@ -1,0 +1,18 @@
+Fixed in 2.2.1 commit 907426b00bdcd69d9a56ac1870990e8ae8c6fe9f
+use rpcbind.target instead of .socket (as the upstream commit uses)
+because the future versions of nfs-server.service activate the socket
+while ours activate the service.
+
+diff --git a/systemd/nfs-mountd.service b/systemd/nfs-mountd.service
+index d908afe..587112b 100644
+--- a/systemd/nfs-mountd.service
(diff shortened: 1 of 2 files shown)


net-fs/nfs-utils/files/nfs-utils-1.3.2-rpcbind-after.patch
```diff
@@ -0,0 +1,18 @@
...
+
+diff --git a/systemd/nfs-mountd.service b/systemd/nfs-mountd.service
+index d908afe..587112b 100644
+--- a/systemd/nfs-mountd.service
++++ b/systemd/nfs-mountd.service
+@@ -3,6 +3,7 @@ Description=NFS Mount Daemon
+ Requires=proc-fs-nfsd.mount
+ After=proc-fs-nfsd.mount
+ After=network.target
++After=rpcbind.target
```
dm0-: Actually, now that I realize only HTTPS is broken, http://git.linux-nfs.org/?p=steved/nfs-utils.git;a=commitdiff;h=907426b00bdcd69d9a56ac1870990e8ae8c6fe9f says `rpcbind.socket`.

ajeddeloh: As per the comment at the top of the diff, the socket unit doesn't get started. I had that originally and it doesn't fix the issue, but .target does.

dm0-: Oh, I somehow missed the top of the patch file.  The LGTM still stands.

PR #2914, opened 2017-11-22
Upgrade to Docker 17.11

dm0-:
The only subcomponent upgrade is containerd, which required adding Go 1.9.


dm0-:
Since I might not finish this until after the holiday: The current PR is not suitable because containerd has switched to a configuration file instead of command-line options.  I will probably push commits with updated service files + torcx package Monday.

dm0-:
I couldn't figure out a good way to package the containerd 1.0.0 config file for torcx, so the containerd service file just includes hard-coded `TORCX_` variables that will be empty for non-torcx installs.  Maybe one day we can implement support for torcx to detect if a tmpfiles path is in the image and prefix it with `$TORCX_UNPACKDIR/$image_name` for copying/linking.

The containerd config file uses our existing paths rather than the new defaults in case anyone depends on them.  See https://github.com/containerd/containerd/blob/992280e8e265f491f7a624ab82f3e238be086e49/docs/ops.md for the specific commit used by Docker 17.11.0.

PR #2921, opened 2017-11-29
Enable LDAP support for sudoers

dm0-:
This brings in sudo from portage-stable just to drop the Perl dependency.  (It also bumps to the next minor version.)  Production images mask the Perl script, but it will still be installed in other targets.

If we want to enable this, I can submit another PR to drop the package from portage-stable.


dm0-:
I'll close in favor of recommending LDAP groups + sudoers rules unless there is some greater value in this.

PR #2922, opened 2017-11-29
coreos-base/coreos-metadata: update to v1.0.4

sdemos:
this fixes incompatibilities with the golang-style single-hyphen flags
and fixes a bug on digital ocean where private ip attributes would be
incorrectly indexed if there were public ips as well.


sdemos:
backported to 1576 in #2923 and 1590 in #2924 

\cc @bgilbert @dm0-

bgilbert:
I'll go ahead and merge this one so we get nightly coverage tonight.

PR #2936, opened 2017-12-04
coreos-base/oem-*: move Ignition base/default configs into OEMs

bgilbert:
For https://github.com/coreos/ignition/pull/475. Configs are as of coreos/ignition@dc992fff72cf45fbed14d2144d05a9df91eb2bd6. Tested with QEMU, PXE, ISO, Azure, DigitalOcean, EC2, GCE, OCI, Packet, and VMware Workstation. Not tested with OpenStack or Brightbox.

Note that this will be the first time the changes in coreos/ignition@222108bdf6a1fd0d5ddabca26e6a84fb78432ea6 reach master.


coreos-base/oem-gce/files/files/google-cloud-sdk.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+alias gcloud="(docker images google/cloud-sdk || docker pull google/cloud-sdk) > /dev/null;docker run -ti --rm --net="host" -v $HOME/.config:/root/.config -v /var/run/docker.sock:/var/run/docker.sock google/cloud-sdk gcloud"
+alias gsutil="(docker images google/cloud-sdk || docker pull google/cloud-sdk) > /dev/null;docker run -ti --rm --net="host" -v $HOME/.config:/root/.config google/cloud-sdk gsutil"
```
dm0-: Maybe drop the quotes from `--net="host"`.  They aren't needed, and they unquote the alias string.  It makes no difference in this case, but it's reminiscent of actual problems.

bgilbert: Done.

bgilbert:
Updated with bootengine bump.

PR #2941, opened 2017-12-06
Update the default Go version to 1.9

dm0-:
We're currently using Go 1.7 as the default, which stopped receiving security updates when Go 1.9 was released.  This sets the default version to 1.9 (specifically 1.9.2 at the moment), and I'd like to get in the practice of keeping the default version set to the current release.

Other commits here try to distinguish between packages where we do or don't care about their Go version.  (An etcd bump was also thrown into the mix to roll into the metadata update.)  With these changes, there are only two sets of packages that specify Go versions:

  - All Docker-related packages built into torcx images specify the same version used by upstream for the best supported configuration.
  - The etcd packages use their upstream-tested Go versions (etcd2 has 1.7; etcd3 has 1.8), since these projects are not developed by the OS team and seem like they'd be prone to performance-affecting changes in Go versions.

The packages that had a pinned Go version removed are all developed by the OS team, so we are best able to handle any needed maintenance.


dm0-:
It looks like this breaks the `coreos.locksmith.tls` test.  Everything else looks okay in a full build and test.  I'll try to isolate this problem before merging.

         cluster.go:117: "sudo locksmithctl --endpoint https://localhost:2379 --etcd-cafile /etc/ssl/etcd/ca-etcd-cert.pem --etcd-certfile /etc/ssl/etcd/locksmith-cert.pem --etcd-keyfile /etc/ssl/etcd/locksmith-key.pem status" failed: output , status Process exited with status 1

PR #2942, opened 2017-12-07
Enable SCSI device handlers as modules

redbaron:
These are mostly used when accessing SAN over iSCSI protocols
using multipath configurations


redbaron:
Addresses https://github.com/coreos/bugs/issues/2281

sys-kernel/coreos-modules/files/commonconfig-4.14
```diff
@@ -514,6 +514,10 @@ CONFIG_SCSI_QLA_FC=m
 CONFIG_SCSI_LPFC=m
 CONFIG_SCSI_VIRTIO=m
 CONFIG_SCSI_DH=y
+CONFIG_SCSI_DH_RDAC=m
+CONFIG_SCSI_DH_HP_SW=m
```
bgilbert: Per the [Kconfig text](https://github.com/torvalds/linux/blob/968edbd93c0cbb40ab48aca972392d377713a0c3/drivers/scsi/device_handler/Kconfig#L24-L26) (from 2008!) let's leave this one disabled.

redbaron: removed

redbaron:
All done.  Would you consider merging similar PR to current stable and beta?

bgilbert:
We generally prefer to let new features promote to stable via alpha and beta.  The next alpha will very likely promote to beta and then to stable in late January, so hopefully the wait shouldn't be too excessive.

Thanks again for the contribution!

PR #2954, opened 2017-12-16
coreos-base/update-ssh-keys: bump to v0.1.2

sdemos:
warn on invalid keys instead of throwing an error


sdemos:
should these changes be backported at all?

sdemos:
backported to stable and beta in #2955 and #2956 respectively

PR #2970, opened 2017-12-27
kernel: Enable throttled background buffered writeback (CONFIG_BLK_WBT)

ezequielgarcia:
From the commit log:
    
    This feature prevents the background writeback thread from stalling I/O,
    e.g. reads.
    
    For instance, without this option enabled, it's possible to stall reads
    setting a writeback dirty threshold (/proc/sys/vm/dirty_writeback_centisecs)
    sufficiently long.
    
    Quoting the author of the feature:
    
    [..] If you are sick of Linux bogging down when buffered
    writes are happening, then this is for you, laptop or server. The
    patchset is fully stable, I have not observed problems. It passes full
    xfstest runs, and a variety of benchmarks as well. It works equally well
    on blk-mq/scsi-mq, and "classic" setups.
    
    This feature adds a simple blk-wb code that keeps limits how much buffered
    writeback we keep in flight on the device end. [..] it should be pretty
    much auto-tuning.

See the patchset for the gory details [1]. Given this feature is not turned
on by default, distributions will start doing so [2].

While here, I bite the bullet and tried to consolidate some kernel options (as per https://github.com/coreos/bugs/issues/2137).

[1] https://lwn.net/Articles/704739/
[2] https://forum.manjaro.org/t/please-enable-writeback-throttling-by-default-linux-4-10/18135/3


ezequielgarcia:
@dm0- Thanks for the review. How about that? (Wasn't sure which commit to put the -r1 rename).

ezequielgarcia:
@dm0- That is fine. No rush at all.

ezequielgarcia:
@bgilbert Gentle ping :-). We need to rebase this on top of master, but it would be good to know if there are more requested changes.

sys-kernel/coreos-modules/files/commonconfig-4.14
```diff
@@ -931,3 +931,11 @@ CONFIG_MODULE_SIG_KEY="certs/modules.pem"
 # CONFIG_XZ_DEC_ARMTHUMB is not set
 # CONFIG_XZ_DEC_SPARC is not set
 CONFIG_NR_CPUS=512
+CONFIG_ACPI=y
+CONFIG_HW_RANDOM=y
```
bgilbert: This unmodularizes HW_RANDOM on amd64.  Is there a reason to do that?

sys-kernel/coreos-modules/files/commonconfig-4.14
```diff
@@ -46,6 +46,7 @@ CONFIG_MODULE_UNLOAD=y
 CONFIG_MODULE_SIG=y
 CONFIG_MODULE_SIG_SHA256=y
 CONFIG_BLK_DEV_THROTTLING=y
+CONFIG_BLK_WBT=y
```
bgilbert: Should we be enabling `BLK_WBT_SQ` as well?

ezequielgarcia: Well, I do not know. A few notes about BLK_WBT_SQ:
* It is supposed to enable writeback throttle by default on single-queue devices. However, it did not work for me. 
* Since writeback throttle can be enabled for any device (regardless of BLK_WBT_SQ) via a simple udev rule, instead of debugging why BLK_WBT_SQ did not work, I simply used a udev rule to enable writeback throttle, using ignition like this:

```
    - path: /etc/udev/rules.d/60-wbt-enable.rules
      filesystem: root
      mode: 0644
      contents:
        inline: |
          ACTION=="add|change", KERNEL=="sd[b-z]*", ATTR{queue/wbt_lat_usec}="75000"
```

sys-kernel/coreos-modules/files/commonconfig-4.14
ezequielgarcia: Was there any reason to build it as a module on amd64? It seems to me that this gives us consistent behavior across platforms, and saw no harm in making this option builtin.

sys-kernel/coreos-modules/files/commonconfig-4.14
```diff
@@ -931,3 +931,11 @@ CONFIG_MODULE_SIG_KEY="certs/modules.pem"
 # CONFIG_XZ_DEC_ARMTHUMB is not set
 # CONFIG_XZ_DEC_SPARC is not set
 CONFIG_NR_CPUS=512
+CONFIG_ACPI=y
+CONFIG_HW_RANDOM=y
+CONFIG_PCI=y
+CONFIG_PCI_MSI=y
+CONFIG_PCI_IOV=y
+CONFIG_KALLSYMS_ALL=y
+CONFIG_KEXEC=y
```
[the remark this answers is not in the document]
ezequielgarcia: I don't run CoreOS on ARM boxes, but again: it seems to me that if it's wanted on x86, it's wanted on any platforms. Maybe I'm missing anything? Or in other words, what is the current use of kexec on x86?

ezequielgarcia:
Hi @bgilbert 
> Apologies for the delay. There have been one or two things going on.  :smile:
>

No problem. It's been a rough couple weeks for everyone :-)

> Thanks for the PR and cleanup. When adding config options, please add them in the order make *config would write them out. We're not perfect about that now, but it'd be great not to make things worse.
>

OK got it. Being a "common" config, was not sure how to do that. Let me give a try.

> SCHEDSTATS and DEBUG_FS are both implicitly enabled on amd64, so this doesn't change anything there (but is harmless).

Indeed. It still seems a good idea to have them in the common file, no? IMO, the slimer the per-platform configs, the cleaner and better the configs will be.

sys-kernel/coreos-modules/files/commonconfig-4.14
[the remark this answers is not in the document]
glevand: Sure, KEXEC for arm64 is OK.

glevand: @ezequielgarcia If possible, could you post a diff of the generated kernel config files with and without this change?  That would make it a lot easier to review.

ezequielgarcia:
@glevand @bgilbert diff for x86_64:

```
--- amd64-defconfig-old	2018-01-11 17:58:24.784415388 -0300
+++ amd64-defconfig-new	2018-01-11 17:57:01.467750924 -0300
@@ -46,6 +46,7 @@
 CONFIG_MODULE_SIG=y
 CONFIG_MODULE_SIG_SHA256=y
 CONFIG_BLK_DEV_THROTTLING=y
+CONFIG_BLK_WBT=y
 CONFIG_PARTITION_ADVANCED=y
 CONFIG_BSD_DISKLABEL=y
 CONFIG_MINIX_SUBPARTITION=y
@@ -768,6 +769,7 @@
 CONFIG_IPMI_SSIF=m
 CONFIG_IPMI_WATCHDOG=m
 CONFIG_IPMI_POWEROFF=m
+CONFIG_HW_RANDOM=y
 CONFIG_HW_RANDOM_TIMERIOMEM=m
 CONFIG_HW_RANDOM_VIRTIO=m
 CONFIG_NVRAM=m

```

ezequielgarcia:
diff for ARM64:
```
--- arm64-defconfig-old	2018-01-11 18:03:59.811073197 -0300
+++ arm64-defconfig-new	2018-01-11 18:04:55.171071732 -0300
@@ -49,6 +49,7 @@
 CONFIG_MODULE_SIG=y
 CONFIG_MODULE_SIG_SHA256=y
 CONFIG_BLK_DEV_THROTTLING=y
+CONFIG_BLK_WBT=y
 CONFIG_PARTITION_ADVANCED=y
 CONFIG_BSD_DISKLABEL=y
 CONFIG_MINIX_SUBPARTITION=y
@@ -94,6 +95,7 @@
 CONFIG_ZSMALLOC=m
 CONFIG_SECCOMP=y
 CONFIG_PARAVIRT_TIME_ACCOUNTING=y
+CONFIG_KEXEC=y
 CONFIG_XEN=y
 CONFIG_RANDOMIZE_BASE=y
 CONFIG_ARM64_ACPI_PARKING_PROTOCOL=y
```

sys-kernel/coreos-modules/files/commonconfig-4.14
ezequielgarcia: Good idea.

glevand:
Seems OK for arm64. 
@ezequielgarcia If you want to test, something like this should work:
```
./build_packages --board=arm64-usr --nousepkg
./build_image --board=arm64-usr prod
./image_to_vm.sh --from=../build/images/arm64-usr/latest --board=arm64-usr --prod_image --format=qemu_uefi
../build/images/arm64-usr/latest/coreos_production_qemu_uefi.sh
```

ezequielgarcia:
@glevand @bgilbert @dm0- Updated the branch. How does it look?

dm0-:
Looks okay to me.  New feature merges are currently frozen to prepare for a release this week, so it might not be merged until next week.

ezequielgarcia:
@glevand @bgilbert Gentle ping.

glevand:
As I said, it seems OK for arm64.  Nothing more to add from me.
@bgilbert has a pending change request.

ezequielgarcia:
@bgilbert Done. How do we launch a ci job?

dm0-:
It's a manual process that involves writing manifest modifications at the moment.  I'll run the tests on it.

dm0-:
Tests looked okay.  Merging, thanks!

PR #2971, opened 2018-01-02
kernel: Enable BFQ I/O scheduler (for blk-mq)

ezequielgarcia:
From the commit log:

BFQ is a proportional-share storage-I/O scheduler for the
blk-mq layer. Its main features are:
    
* Low latency for interactive applications
* Low latency for soft real-time applications
* Higher speed for code-development tasks
* High throughput
* Strong fairness, bandwidth and delay guarantees
    
Note that the default blk-mq scheduler is "None", so
the BFQ scheduler must be explicitly chosen for a given
storage device.
    
Note also that BFQ only supports blk-mq layer, so drivers
using legacy blk layer are not affected by this change.
    
In particular, both HDD and SSD storage drives use
the legacy blk layer by default.

This PR is based on https://github.com/coreos/coreos-overlay/pull/2970. It has to be rebased once https://github.com/coreos/coreos-overlay/pull/2970 is merged.


bgilbert:
Please rebase.

ezequielgarcia:
@bgilbert rebased.

PR #2973, opened 2018-01-02
Upgrade Docker to 17.12

dm0-:
This pins 17.12 to Go 1.9.


app-emulation/docker/docker-9999.ebuild
```diff
@@ -211,9 +211,12 @@ src_unpack() {
 		DOCKER_BUILD_DATE=$(git -C "${S}" log -1 --format="%ct")
 	fi
 	[ "${#ENGINE_PATCHES[@]}" -gt 0 ] && eapply -d"${S}"/components/engine "${ENGINE_PATCHES[@]}"
+	# XXX: fix the insane version in the 17.12.0 tag -- DELETE THIS ON UPDATE
+	echo "${PV}-ce" > "${S}/VERSION"
```
bgilbert: For safety, it might be worth checking for the current (incorrect) contents of the file, and failing the build if some other value is found.

dm0-: Okay, I'll note to replace that line with such a test when they have new release with a usable value.

bgilbert: That might also be useful, but I meant in this PR, to make sure we don't accidentally continue to clobber the version after we no longer need to.

dm0-: Oh, okay, I had just planned on dropping it as soon as there is an update.  (I grep for the version every update so I wouldn't miss that line.)

PR #3008, opened 2018-01-12
sys-apps/ignition: v0.20.0 -> v0.20.1

dgonyeo:


ajeddeloh:
@dm0, the cros_workon stuff is smart enough to figure out this isn't on `master`, right?

dgonyeo:
Good to merge

dm0-:
As long as the commit specified in the ebuild exists under a branch or tag, it should all work.

The repo will still be on master by default after `repo sync`, so that will be what is used by cros_workon until you check out something else.  That can be changed in the manifest, but since this is against the overlay master branch, I think it should remain pointing at Ignition's master branch.

PR #3024, opened 2018-01-24
Upgrade to Rust 1.23.0

dm0-:
The cargo ebuild requires `virtual/rust`, which requires its version to match `dev-lang/rust`, but Gentoo hasn't updated the virtual package yet, so we're stuck with it.  We should probably wait to merge this until we're closer to the alpha freeze to delay adding huge build times to all new SDKs.  If the virtual package is updated by then, I'll remove it from the overlay as well.

Part of coreos/portage-stable#637


dm0-:
Updated the virtual package in portage-stable.

PR #3027, opened 2018-01-26
sys-apps/systemd: bump for networkd fix

ajeddeloh:
This fixes a bug where specifying classful routes caused the gateway
route to be ignored.


ajeddeloh:
Whoops, fixed. We really ought to have a bot for that.

PR #3030, opened 2018-01-29
sys-kernel/coreos-*: add 4.15.0

bgilbert:
No config changes are required.


sys-kernel/coreos-kernel/coreos-kernel-4.15.0.ebuild
```diff
@@ -0,0 +1,95 @@
+# Copyright 2014-2016 CoreOS, Inc.
```
dm0-: Should we bump the year in new files?

bgilbert: They're copies of other files and haven't really changed at all, so in this case I'd say no.

PR #3040, opened 2018-02-03
Enable installing compressed kernel modules

dm0-:
Testing this to see what happens.

Fixes coreos/bugs#2339


dm0-:
Before:

    $ du -sh /build/amd64-usr/lib/modules 
    148M	/build/amd64-usr/lib/modules

After:

    $ du -sh /build/amd64-usr/lib/modules
    97M	/build/amd64-usr/lib/modules

dm0-:
Just noting that a test build had some boot problems on a few platforms, so this will need some more investigation first.

dm0-:
It was 42M, now 39M.

dm0-:
It was decided that too many containers are based on distros without kmod compression support, so this will not be merged in the near future.  I'll close this, but it can be resurrected if the landscape changes.

PR #3052, opened 2018-02-07
coreos-base/hard-host-depends: build fewer things

ajeddeloh:
Remove packages we don't use so we stop building them every time.


ajeddeloh:
Added the removal of go-bindata to address https://github.com/coreos/bugs/issues/2349

ajeddeloh:
Tested with a full manifest run. There was a failure with libpcre but it looks unrelated and replaying the jobs worked. Good for review.

The go-bindata part was not tested as part of the manifest run (added later), but looks pretty harmless.

coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
```diff
@@ -148,11 +130,6 @@ RDEPEND="${RDEPEND}
 	dev-python/gdata
 	"
 
-# Host dependencies for HWID processing
-RDEPEND="${RDEPEND}
-	dev-python/pyyaml
```
bgilbert: Should this package be dropped?

coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
```diff
@@ -52,17 +51,8 @@ RDEPEND="${RDEPEND}
 	dev-libs/libxslt
 	dev-libs/libyaml
 	dev-libs/protobuf
-	dev-python/cherrypy
 	dev-python/ctypesgen
-	dev-python/m2crypto
 	dev-python/mako
-	dev-python/netifaces
-	dev-python/pyinotify
-	dev-python/pyopenssl
```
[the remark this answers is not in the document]
ajeddeloh: No, that's a mistake in 642. Will fix that.

coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild
ajeddeloh: I _think_ so. Everything built fine without it. `hard-host-depends` is the only thing that depends on it and `repo grep "import yaml" ` and `repo grep "from yaml"` don't turn anything up. [git blame](https://github.com/coreos/coreos-overlay/blame/master/coreos-base/hard-host-depends/hard-host-depends-0.0.1.ebuild#L153) shows it's from the initial import from ChromeOS

PR #3059, opened 2018-02-09
Upgrade GCC to 7.3.0

dm0-:


bgilbert:
Presumably the upstream kernel was tested with 7.3; why do we hit the fallthrough warning and they don't?

dm0-:
I don't know yet.  The two points of failure both have `/* FALLTHROUGH */` comments.

PR #3062, opened 2018-02-10
net-misc/google-cloud-sdk: bump to 188.0.1

ajeddeloh:
Bump google-cloud-sdk to latest. Remove python3 files from yet another
place in src_prepare. Do not "doman" since there are no longer any man
pages.


ajeddeloh:
Would be nice to get this merged early in the cycle so we can ensure it doesn't break jenkins and have time to fix things if it does.

PR #3068, opened 2018-02-13
Bump 4.14.18 to 4.14.19 in beta

dm0-:
This doesn't change the patches so that it doesn't reference the GCC 7 warning in the new patch set.


dm0-:
For simplicity, I'm just adding the beta Docker tag fix here.

PR #3082, opened 2018-02-20
app-emulation/containerd: Fix the 1.0.* manifest entries

dm0-:
The old entries were for the release archives, which apparently worked at one point, since they're the versions that were downloaded into our source mirror.  This changes the entries to use the commit archives instead, which just have different parent directories.

### DIFF
--- a/app-emulation/containerd/containerd-1.0.0.ebuild
+++ b/app-emulation/containerd/containerd-1.0.0.ebuild
@@ -11,11 +11,13 @@ if [[ ${PV} == *9999 ]]; then
 	EGIT_REPO_URI="https://${GITHUB_URI}.git"
 	inherit git-r3
 else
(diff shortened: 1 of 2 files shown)


euank:
We should probably update our ebuilds to not rename to `${P}` too.

I think those files were likely mirrored from gentoo's distfiles, and if we're going to depart from what they've got mirrored, it would be nice to have a different filename too

dm0-:
In that case, should we try to go the other way and change our `SRC_URI` to use the release archives?  Docker seems to be using containerd tags in their releases lately, and containerd isn't set up for `cros_workon` where we'd need to reference a commit otherwise.

euank:
Yeah, changing in that direction also makes sense to me and will look a bit nicer.

dm0-:
Updated.

PR #3095, opened 2018-02-28
Update LTS Docker to 17.12.1

dm0-:


redbaron:
Is it coming to beta?

dm0-:
It was suggested offline that I not backport this to the current beta branch, so it will be three weeks until it promotes to beta normally.  (If Docker 18.04 is released by then, that beta would get 18.03 instead.)  Does this version fix a problem for you in beta?

redbaron:
We've seen PLEG timeouts with 17.12.0 which looks very much like https://github.com/moby/moby/issues/36010. 17.12.1 changelog suggests that it was fixed with runc update.

redbaron:
If updating it in Beta is not really going to happen (why not, BTW?), then with https://gist.github.com/dm0-/66f8375418a8f2369070a2f5902ec5f0 I can try to replace Docker torcx package myself, but for that 17.12.1 pkg need to be published somewhere. I found https://console.cloud.google.com/storage/browser/builds.developer.core-os.net/torcx/pkgs/amd64-usr/docker/ , but checking each to find if there is 17.12.1 published at all is going to take some time :)

dm0-:
It sounded like backporting wasn't considered because there didn't seem to be big security fixes or reported issues, but since you're hitting an issue, it can be reconsidered.

Since 17.12.1 was released after our most recent alpha was branched, it's not built for any releases yet.  (We might not even have a build from test images yet, waiting on #3105 to resume those.)  You could build it yourself with the `build_torcx_store` script in the SDK until then, if needed.

dm0-:
Here is a 17.12.1 package from a test build today: http://builds.developer.core-os.net/torcx/pkgs/amd64-usr/docker/f0b3be513bc1b699a41ae787df73c477b5211219cdfdaeaf4326638fba5b610aa7faa8c2a73e8ec0893b1f4bb4bb0b864f2cf455bdeb67b403e1af2e65d5722c/docker:17.12.torcx.tgz

dm0-:
Docker 17.12.1 was backported to beta in #3108, and we should release it next Wednesday (if no issues are discovered that require a minor release before then).

redbaron:
Thanks, already using dev build from your secret bucket  :)

PR #3113, opened 2018-03-08
Bump to systemd v238

dm0-:
Depends on coreos/portage-stable#648


sys-apps/systemd/systemd-9999.ebuild
```diff
@@ -10,7 +10,7 @@ if [[ ${PV} == 9999 ]]; then
 	# Use ~arch instead of empty keywords for compatibility with cros-workon
 	KEYWORDS="~amd64 ~arm64 ~arm ~x86"
 else
-	CROS_WORKON_COMMIT="26a2f8c6f74d2b12b0b31610c35f48148c3e4264" # v237-coreos
+	CROS_WORKON_COMMIT="c20679823e34346352ec2a619d888f33bcb32389" # v238-coreos
```
lucab: After the backport, this is `dd42d3d148eef87f65666199c85905308a95bc53`.

sys-kernel/coreos-kernel/coreos-kernel-4.14.26.ebuild
```diff
@@ -19,7 +19,7 @@ DEPEND="${RDEPEND}
 	sys-apps/less
 	sys-apps/sed
 	sys-apps/shadow
-	sys-apps/systemd[cryptsetup]
+	sys-apps/systemd:=[cryptsetup]
```
lucab: As a learning occasion for me: the effect of this would be to always rebuild in case of slot updates, is that correct?

dm0-: Thanks for pointing this out.  Yes, it rebuilds for slot updates, but I didn't notice the systemd ebuild has a hard-coded `SLOT` value, not `0/${PVR}` like most of our other live ebuilds, so this won't work as intended.  I'll just drop that commit so we stick the the upstream `SLOT` value, and we'll have to remember to bump the kernel revision for systemd upgrades to get it in the initramfs.

PR #3136, opened 2018-03-20
profiles/coreos: update gnupg use flags

euank:
Goes along with https://github.com/coreos/portage-stable/pull/651

Newer versions of gnupg have an 'ssl' use flag, which is effectively the
same as the gnutls flag we already disabled.

This PR also enables 'usb' in addition to 'smartcard' for the sdk. Some
smartcards are also usb devices.

### DIFF
--- a/profiles/coreos/arm64/package.accept_keywords
+++ b/profiles/coreos/arm64/package.accept_keywords
@@ -4,7 +4,6 @@
 =app-admin/sudo-1.8.20_p2 ~arm64
 =app-arch/bzip2-1.0.6-r8 ~arm64
 =app-arch/libarchive-3.3.1 ~arm64
-=app-crypt/gnupg-2.2.4 ~arm64
 =app-crypt/mit-krb5-1.14.2 ~arm64
 =app-text/asciidoc-8.6.9-r3 ~arm64
 =dev-cpp/gflags-2.1.2 ~arm64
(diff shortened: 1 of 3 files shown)


profiles/coreos/base/package.use.mask
```diff
@@ -1,5 +1,5 @@
 # We don't ship GnuTLS by default, and smartcard has a dep loop
-app-crypt/gnupg gnutls smartcard
+app-crypt/gnupg gnutls smartcard ssl
```
dm0-: Maybe drop `gnutls` here since it's not used anymore?  Also, you're enabling `smartcard`, so it should probably be dropped here.  Have you tested rebuilding the SDK and boards to see if the dependency loop is still an issue?

euank: Dropped the extraneous flag.

I haven't done a full sdk rebuild, but since the sdk had smartcard support before this change (just not usb+smartcard), I don't think it was relevant for the sdk already.

dm0-: Ah, okay, I missed how that flag was unmasked separately in the SDK profile.

profiles/coreos/arm64/package.accept_keywords
```diff
@@ -4,7 +4,7 @@
 =app-admin/sudo-1.8.20_p2 ~arm64
 =app-arch/bzip2-1.0.6-r8 ~arm64
 =app-arch/libarchive-3.3.1 ~arm64
-=app-crypt/gnupg-2.2.4 ~arm64
+=app-crypt/gnupg-2.2.4-r2 ~arm64
```
dm0-: I think this line can just be dropped.  That ebuild has both `amd64` and `arm64`.

euank:
@dm0- Addressed comments I think

PR #3169, opened 2018-04-10
pass /etc/machine-id from host to kubelet

databus23:
Otherwise kubelet/cadvisor picks up whatever is shipped in the container image, which results in all nodes reporting the same value for `status.nodeInfo.machineID`


lucab:
This looks fine to me, but I'd like to have @euank and @aaronlevy check for possible side-effects at the k8s level before merging this.

lucab:
@databus23 thanks for the patch, this is mostly ready to get in. Can you please also bump the ebuild version?

databus23:
@dm0- renamed ebuild file as requested

euank:
Thanks again!

PR #3205, opened 2018-04-30
coreos-base/oem-azure: update wa-agent to 2.2.25

ajeddeloh:
Update the azure agent to 2.2.25.

Still need to test this.


bgilbert:
Might be a good idea to resync `waagent.conf`.

PR #3210, opened 2018-05-01
Enable SSSD and LDAP sudoers support

dm0-:
As part of this, LDAP is enabled globally for arm64.  It seems to really only be used by the SSSD/AD packages until this point.  This is also an upgrade.


glevand:
LGTM.

PR #3219, opened 2018-05-03
sys-kernel/coreos-sources: add dep on bison, flex

ajeddeloh:
The 4.16.x line and up now need flex and bison to build the kernel. Add
it here so people using our dev container will pick it up automatically.


ajeddeloh:
You're right, I don't think it is either. Updated to not bump

PR #3221, opened 2018-05-08
profiles/arm64: sync wget with amd64

euank:
Part of https://github.com/coreos/portage-stable/pull/662, fixes CVE-2018-0494


euank:
Indeed, I missed that. Closing.

euank:
In retrospect I should have updated the message here. Anyways, moved to https://github.com/coreos/coreos-overlay/pull/3222 to drop it instead of update it.

PR #3258, opened 2018-05-30
Bump Linux 4.14.44 to 4.14.47 and 4.16.12 to 4.16.13

dm0-:


dm0-:
I've just added another LTS kernel bump commit to these PRs rather than squashing so that the last commit hash continues to exist for the Jenkins jobs that are still running on it.

PR #3286, opened 2018-06-12
Bump Python

dm0-:
Part of coreos/portage-stable#641


dm0-:
I've rebased these PRs and added a few more package syncs to get more Python 3.6 coverage.

There are still a few that don't have it enabled in Gentoo (e.g. `libnl` and `gdb`) that don't seem worth moving to the overlay to update, and updating `boost` has new arm64 build failures that I didn't dig into.  So if we ever decide to switch to Python 3 as the preferred version, we'll need to remember to do another round of package syncs to catch those, but there shouldn't be a problem as is.

vielmetti:
Hey @dm0- can you point specifically to the `boost` arm64 failures you are seeing? I'm happy to help get some eyes on that project via @worksonarm .

dm0-:
@vielmetti I didn't save the output and can't recall specifically what the problem was now, but it was most likely a failure with cross-compiling in general (not with arm64 specifically).  I made boostorg/build#196 for that issue a few versions ago, which will probably need to be extended.  I can retry the upgrade to get error messages later today to verify.

dm0-:
@vielmetti  Actually, it looks like it's cross-compiling okay now.  The problem arose because of build ordering, where the cross-compiled version depended on having the same version installed natively on the host, which I hadn't done yet.  While it's not great for the cross-compiling package to depend on the host package, it should still work for us since the host packages are upgraded first by our normal OS build process.  I can submit the boost update as a separate PR from this since our latest test results only cover these changes.

PR #3295, opened 2018-06-18
Bump Linux 4.14.49 to 4.14.50 and 4.16.15 to 4.16.16

dm0-:


dm0-:
Updated.  I didn't notice the command I reran to take patches from `format-patch` globbed for `000*.patch`.

PR #3299, opened 2018-06-19
updateservicectl: update to v2.1.0

sdemos:
there are several related prs - 
* https://github.com/coreos/updateservicectl/pull/137 needs to be merged to release v2.1.0 of `updateservicectl`
* https://github.com/coreos/scripts/pull/824 updates `updateservicectl` usage in `transfer.sh`
* the new version of CoreUpdate needs to be released and deployed.


ajeddeloh:
Typically we tag the merge commit on projects, and since the updateservicectl PR isn't merged, wont you need a different hash?

dm0-:
In this case, the PR contains the commit that bumps the version to a dev version, so the merge commit would have the wrong version for the release.  Handling it this way should be fine as long as the PR is based on the latest commit, and I believe we've done it this way before.

sdemos:
most of the projects I maintain tag a commit that isn't the merge commit when tagging a release. normally a commit specifically made to be a release commit. 

the release pr for updateservicectl is merged now, so this should be good to go as long as there will be time between now and the next release merge the scripts changes and deploy the new coreupdate.

PR #3310, opened 2018-06-28
Upgrade systemd to 241

dm0-:
Depends on coreos/portage-stable#678


profiles/coreos/targets/generic/make.defaults
```diff
@@ -1,8 +1,8 @@
 # Copyright (c) 2012 The Chromium OS Authors. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
 
-USE="cros-debug acpi usb symlink-usr usrmerge cryptsetup policykit"
```
cgwalters: Is that now the default?

dm0-: No, `usrmerge` was replaced by `split-usr` upstream, which has an inverted meaning.  https://github.com/gentoo/gentoo/commit/f602af077d978ba8982ced38021ab97e90b7fcca  Our systemd ebuild was the last place using `usrmerge`, so we can just drop it.

ajeddeloh:
Why the one commit for sync+bump instead of the normal 2?

dm0-:
No particular reason other than this package didn't use the two separate commits previously.  We can start doing that in the future, or I can rewrite this if you think it's better to start now.

ajeddeloh:
/shrug I think it'd be good in the future, but if we haven't done it in the past there's no need to rewrite this now.

dm0-:
Random dbus failures were encountered in test runs with this update; do not merge until it's resolved.

dm0-:
I've bumped the commit to include coreos/systemd#104.  We discussed waiting to merge until July 31st to skip the next stable cycle.

dm0-:
This is currently blocked on systemd/systemd#9656.

PR #3312, opened 2018-06-28
Add Travis job for running repoman

csssuf:
Getting some QA checks on ebuilds here would be nice. Our setup generates some false positives with the default configuration, though. In order to avoid them, disable the `keywords`, `pkgmetadata`, and `profiles` modules, and the `gentooheader` line check, since we have lots of non-Gentoo ebuilds in the overlay. For the in-depth reasoning for each module see the comment in `metadata/repoman/repository.yaml`.

I've also gone through and set a few checks as warnings instead of fatal errors; see `metadata/repoman/qa_data.yaml`. I've mostly just used my own judgment to do so, so if anything there needs to change it can.


csssuf:
For reference, here's the output of the Travis run for this branch:
https://travis-ci.org/csssuf/coreos-overlay/builds/397928358

There are a number of warnings generated, but I believe they're all valid.

.travis/setup_repoman.sh
```diff
@@ -0,0 +1,23 @@
...
+git clone https://github.com/coreos/portage-stable /usr/portage/
+cp .travis/coreos.conf /etc/portage/repos.conf/
+ln -s /usr/coreos-overlay/profiles/coreos/amd64/sdk /etc/portage/make.profile
+mkdir -p /usr/portage/metadata/{dtd,xml-schema}
+wget -O /usr/portage/metadata/dtd/metadata.dtd https://www.gentoo.org/dtd/metadata.dtd
+wget -O /usr/portage/metadata/xml-schema/metadata.xsd https://www.gentoo.org/xml-schema/metadata.xsd
+
+# Download portage.
+mkdir /tmp/portage && cd /tmp/portage
+wget -qO - "https://github.com/gentoo/portage/archive/portage-${PORTAGE_VER}.tar.gz" | tar xz
```
dm0-: Does this still work now that Gentoo's GitHub project is down?  Maybe use e.g. https://gitweb.gentoo.org/proj/portage.git/snapshot/portage-2.3.41.tar.gz to pull from gentoo.org's Git repo.

csssuf: Indeed it doesn't, using gentoo.org is likely a better solution in general anyway.

PR #3319, opened 2018-07-03
Implement google cloud login

ajeddeloh:
Mostly done. Still need to do a little cleanup and test it thoroughly.
 - [x] Clean up, especially the script
 - [x] Update CL gce docs to cover:
    - [x] What it is, how it works, link to upstream
    - [x] It's supported, on by default on newly provisioned systemd
    - [x] How to enable it manually on existing systems
    - [x] How to disable it on new systems
 - [x] Test
   - [x] Unauthorized users cannot login
   - [x] Authorized users can log in but do not have sudo powers
   - [x] Authorized admins can log in and do have sudo powers
   - [x] masking the unit prevents all oslogins
 - [ ] Add kola tests

Depends on:
 - https://github.com/coreos/portage-stable/pull/680
 - https://github.com/coreos/docs/pull/1249


sys-auth/google-oslogin/google-oslogin-20180611.ebuild
```diff
@@ -0,0 +1,39 @@
...
+	insinto /usr/lib64
+	doins libnss_cache_google-compute-engine-oslogin-1.3.0.so
+	doins libnss_google-compute-engine-oslogin-1.3.0.so
+
+	exeinto /usr/libexec
+	doexe google_authorized_keys
+
+	insinto /usr/lib64/security
+	doins pam_oslogin_admin.so
+	doins pam_oslogin_login.so
```
dm0-: There is a `dopammod` function for installing PAM modules.

sys-auth/google-oslogin/google-oslogin-20180611.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+src_unpack() {
+	S=${WORKDIR}/compute-image-packages-${PV}/google_compute_engine_oslogin
+	default
+}
+
+src_install() {
+	insinto /usr/lib64
+	doins libnss_cache_google-compute-engine-oslogin-1.3.0.so
+	doins libnss_google-compute-engine-oslogin-1.3.0.so
```
dm0-: There is a `dolib.so` function for installing shared libraries.

sys-auth/google-oslogin/google-oslogin-20180611.ebuild
```diff
@@ -0,0 +1,39 @@
...
+# TODO: flesh this out
+DEPEND="
+	net-misc/curl[ssl]
+	dev-libs/json-c
+	virtual/pam
+"
+RDEPEND="${DEPEND}"
+
+src_unpack() {
+	S=${WORKDIR}/compute-image-packages-${PV}/google_compute_engine_oslogin
```
dm0-: This is usually set outside functions, and I think that makes it easier to find when the source uses an odd directory.  That could also allow you to omit the `src_unpack` definition if it's just calling `default`.

sys-auth/google-oslogin/google-oslogin-20180611.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+DESCRIPTION="Components to support google cloud os login. This contains bits that belong in USR"
+HOMEPAGE="https://github.com/GoogleCloudPlatform/compute-image-packages/tree/master/google_compute_engine_oslogin"
+SRC_URI="https://github.com/GoogleCloudPlatform/compute-image-packages/archive/${PV}.tar.gz"
+
+LICENSE=""
```
dm0-: The project looks like it's `Apache-2.0`.

coreos-base/oem-gce/files/files/pam-sshd-config
```diff
@@ -0,0 +1,3 @@
+account    requisite    pam_oslogin_login.so
+account    optional     pam_oslogin_admin.so
+session    optional     pam_mkhomedir.so
```
dm0-: It looks like this file will override `/lib/pam.d/sshd`, which configures SSH to use the `system-login` settings.  That seems like something we don't want to throw out completely, so this will probably need to have the additional rules mixed into the existing configuration.

coreos-base/oem-gce/files/units/oem-gce-enable-oslogin.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+DefaultDependencies=false
+Requires=sysinit.target
+After=sysinit.target
+
+[Service]
+ExecStart=/usr/share/oem/scripts/enable-oslogin.sh
```
dm0-: This appears to be running every boot, and the script looks like it will continually be appending to `nsswitch.conf` and overwriting files in `/etc` which the user might have modified.

coreos-base/oem-gce/files/units/oem-gce-enable-oslogin.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+DefaultDependencies=false
+Requires=sysinit.target
+After=sysinit.target
```
dm0-: Does this need to be ordered before sshd can start?

ajeddeloh: Doesn't appear so. When I was first testing it I added the files manually and sshd seemed to be fine picking up the changes without restart. I still wanted it pretty early in the boot process just in case.

coreos-base/oem-gce/files/units/oem-gce-enable-oslogin.service
ajeddeloh: whoops, this was meant to have a `ConditionFirstBoot`. will fix.

coreos-base/oem-gce/files/units/oem-gce-enable-oslogin.service
dm0-: Since we're socket-activating sshd by default, there is still the possibility of a user making an SSH connection before this script has run.

coreos-base/oem-gce/files/scripts/enable-oslogin.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/bash
```
bgilbert: Any reason not to have this in `/usr/share/oem/bin`?

This script should probably either be idempotent or, if that's not worth doing, should bail out if run twice.

coreos-base/oem-gce/files/files/sudoers-file
```diff
@@ -0,0 +1 @@
+#includedir /var/google-sudoers.d
```
bgilbert: The FHS wants this in `/var/lib`.  How hard is it to change?

sys-auth/google-oslogin/google-oslogin-20180611.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+DESCRIPTION="Components to support google cloud os login. This contains bits that belong in USR"
```
bgilbert: please don't unnecessarily uncapitalize :sadface:

coreos-base/oem-gce/files/files/sudoers-file
ajeddeloh: wouldn't be too bad, I'll add it.

ajeddeloh:
Updated. @dm0 does that patch hack in the ebuild look reasonable?

dm0-:
It looks okay, but I think you can drop the `S` definition in `src_prepare` and use that for the global value.  The `eapply` function can take `-p2` so it applies in the subdirectory.

coreos-base/oem-gce/files/units/oem-gce-enable-oslogin.service
```diff
@@ -0,0 +1,13 @@
...
+Before=sshd.socket
```
dm0-: This should probably have `sshd.service` as well for users who disable socket-activation, since there is otherwise no guarantee that this script runs before `sshd` starts accepting connections.

coreos-base/oem-gce/files/bin/enable-oslogin.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -e
+
+# nsswitch.conf changes
+grep oslogin /etc/nsswitch.conf || sed -i '/^passwd:/ s/$/ cache_oslogin oslogin/' /etc/nsswitch.conf
+
+# pam.d entry
+cat /usr/lib/pam.d/sshd /usr/share/oem/files/pam-sshd-config > /tmp/pam_sshd_config
+mv /tmp/pam_sshd_config /etc/pam.d/sshd
+#cp /usr/share/oem/files/pam-sshd-config /etc/pam.d/sshd
```
bgilbert: commented-out code :cry:

coreos-base/oem-gce/files/bin/enable-oslogin.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# This is a stripped down version of the google oslogin control script
+
+set -e
+
+# nsswitch.conf changes
+grep oslogin /etc/nsswitch.conf || sed -i '/^passwd:/ s/$/ cache_oslogin oslogin/' /etc/nsswitch.conf
+
+# pam.d entry
+cat /usr/lib/pam.d/sshd /usr/share/oem/files/pam-sshd-config > /tmp/pam_sshd_config
```
bgilbert: tempfile vulnerability.

coreos-base/oem-gce/files/bin/enable-oslogin.sh
```diff
@@ -0,0 +1,24 @@
...
+mv /tmp/pam_sshd_config /etc/pam.d/sshd
```
bgilbert: Silently clobbers existing customizations, which doesn't seem great for sshd.

coreos-base/oem-gce/files/bin/enable-oslogin.sh
```diff
@@ -0,0 +1,24 @@
...
+#cp /usr/share/oem/files/pam-sshd-config /etc/pam.d/sshd
+
+# sshd_config changes
+cat /usr/share/ssh/sshd_config /usr/share/oem/files/sshd-config > /tmp/sshd_config
+mv /tmp/sshd_config /etc/ssh/sshd_config
```
bgilbert: tempfile vuln; clobbers sshd_config.

coreos-base/oem-gce/files/bin/enable-oslogin.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/bash
```
bgilbert: nit: Can we drop the `.sh` extension?  It's kind of ugly in `bin` directories.

coreos-base/oem-gce/files/bin/enable-oslogin.sh
dm0-: This script seems like it will overwrite files in `/etc` that the user might have configured with Ignition.  It should probably check if it's replacing anything other than an expected symlink.

sys-auth/google-oslogin/google-oslogin-20180611.ebuild
```diff
@@ -0,0 +1,45 @@
...
+
+src_prepare() {
+	# We only care about the oslogin part of the repo, but patches are generally
+	# generated for the repo as a whole. Do the patching then change $S to be the
+	# directory we actually want.
+	eapply "$FILESDIR/0001-pam_module-use-var-lib-instead-of-var.patch"
+	S=${WORKDIR}/compute-image-packages-${PV}/google_compute_engine_oslogin
+	default
+}
+
```
dm0-: Upon further inspection, this project is yet another non-standard build system, and so will need to inherit `toolchain-funcs` and add a `src_compile` to use dependencies from the board (unless you want to start installing `json-c` and `pam` in the SDK root):

```
src_compile() {
	emake CC="$(tc-getCC)" CXX="$(tc-getCXX)" JSON_INCLUDE_PATH="${ROOT%/}/usr/include/json-c"
}
```

ajeddeloh:
Reworked this substanially. It now ships whole configs (with notes saying to keep them in sync with their origin). It also now does everything via Ignition and symlinks. Here's a CLC to disable it (will update docs PR  with it once this is finalized) https://gist.github.com/ajeddeloh/ce90afb14dee6f83e1718c0982d6c2a3 Disabling kinda abuses Ignition's implementation in that appended configs can overwrite things in the base configs since the entries in `links` are processed in order and appending configs appends to the list.

There's also a gce image called `andrew-amd64-usr-andrew-oslogin-test` If you want to play around with it.

sys-auth/google-oslogin/google-oslogin-20180611.ebuild
```diff
@@ -0,0 +1,52 @@
...
+	dev-libs/json-c
+	virtual/pam
+"
+
+RDEPEND="${DEPEND}"
+
+S=${WORKDIR}/compute-image-packages-${PV}/google_compute_engine_oslogin
+
+src_prepare() {
+	eapply -p2 "$FILESDIR/0001-pam_module-use-var-lib-instead-of-var.patch"
```
csssuf: Is there a reason to do this instead of setting `$PATCHES` and adapting the patch to work with `-p1`?

ajeddeloh: Mostly because this makes generating the patch easy.

coreos-base/oem-gce/files/base/base.ign
```diff
@@ -28,6 +28,44 @@
...
         "mode": 292
       }
+    ],
+    "links": [
+      {
+        "filesystem": "root",
+	"path": "/etc/ssh/sshd_config",
+	"target": "/usr/share/google-oslogin/sshd_config",
+	"overwrite": true
+      },
```
csssuf: Tab/space issues with indentation above?

bgilbert:
Looks like the commit messages are stale.

Are you planning to copy the `# Keep this in sync with` comments into the reciprocal packages as well?  Those config files are more likely to be modified than the GCE-private ones.

Do you plan to have scripts cross-check the config files during image build?

ajeddeloh:
Updated to bump the agent as well. Tested. Doesn't have comments in the sources yet with the "keep this in sync" but I figured I'd put this up for review while I did that and worked on the docs.

Went back to the oneshot first-boot unit because of https://github.com/coreos/ignition/issues/608 (Ignition is not declarative).

cc @bgilbert for the script changes

One minor nit is we try to list failed units on login, but oslogin users do not have permissions to do that and get a "failed to list units: access denied" on login, but I'm not sure how to silence that.

dm0-:
Oh, I just noticed that `coreos-base/coreos` is missing a revision bump.

And I think I mentioned somewhere before that the oslogin service should probably be ordered before the SSH server units so early connections aren't racing with the configuration being written in place.  It's not a big issue since it only affects the first boot, and socket-activation should allow later connections to succeed, but it's still possible to be a source of random failures.

ajeddeloh:
Good catch, fixed.

ajeddeloh:
Lost the race with the second comment, fixed that now too.

dm0-:
I think it also needs to be ordered before `sshd.socket` to cover the socket-activated case.

ajeddeloh:
Fixed. I'm not 100% sure its necesary, but it can't hurt to have it.

coreos-base/coreos-init/coreos-init-9999.ebuild
```diff
@@ -2,6 +2,8 @@
 # Copyright (c) 2011 The Chromium OS Authors. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
 
+# NOTE: if /etc/sshd_config changes make sure to copy the change to sys-auth/google-oslogin
```
bgilbert: `/etc/ssh/sshd_config`?

I'm concerned that these comments are too subtle and will be missed.  I understand not putting them in the files themselves, but how about adding `README`s to the package directories?

ajeddeloh: In addition to or instead of?

bgilbert: Instead of, I think?  Either way.

ajeddeloh: :+1: I'm in favor of that as well. Means we don't have to keep the notes about keeping things in sync in sync

ajeddeloh: Fixed.

coreos-base/coreos-init/README
```diff
@@ -0,0 +1,2 @@
+If /etc/sshd_config changes make sure to copy the change to sys-auth/google-oslogin.
```
bgilbert: `/etc/ssh/sshd_config`?

Maybe s/copy/apply/.

net-misc/openssh/README
```diff
@@ -0,0 +1,2 @@
+If you change sshd.pam_include.2 make sure to copy the change to sys-auth/google-oslogin
```
bgilbert: Oddly, the phrasing here is different than the other two.

PR #3320, opened 2018-07-06
Add fero-client

csssuf:
Add an ebuild for `fero-client`, and add it to `sdk-depends`. Blocked on:
- [x] Publishing fero repository
- [x] Adding fero to manifest (coreos/manifest#157)


coreos-devel/sdk-depends/sdk-depends-0.0.1.ebuild
```diff
@@ -24,6 +24,7 @@ DEPEND="
 	app-shells/bash-completion
 	coreos-base/hard-host-depends
 	coreos-base/coreos-sb-keys
+	coreos-devel/fero-client
```
ajeddeloh: Maybe should in `sdk-extras` or jenkins will build it into the sdk every release. Makes more sense if just the people signing build it locally imo.

csssuf:
Now that v0.1.0 has been tagged, this should be good to go.

PR #3334, opened 2018-07-18
sys-kernel/bootengine: Bump to fix resolv.conf

dm0-:
The previous service file would leave an empty /etc/resolv.conf in place, preventing tmpfiles from creating the symlink.

Depends on coreos/bootengine#146


sys-kernel/bootengine/bootengine-9999.ebuild
```diff
@@ -10,7 +10,7 @@ CROS_WORKON_REPO="git://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="30b4f2a73ecde5abbd2c5dd7e78fbf4bdbff3722"
+	CROS_WORKON_COMMIT="03ebaa6c28031da152134dabcecad5c26896ba5d"
```
bgilbert: Should this be the merge commit?

dm0-: Yes, after it's merged.  I put the PR commit there so Jenkins can test this.

PR #3294, opened 2018-06-15
sys-libs/glibc: sync with gentoo stable, to 2.26

iaguis:
This bumps glibc to 2.26, which is the latest stable on gentoo.

I had to bump eclasses on portage-stable so this also bumps the EAPI of
dev-util/crosutils because EAPIs older than 4 are not allowed to use git-r3.

Part of coreos/portage-stable#674

PR #3348, opened 2018-08-02
Fix glibc upgrade problems

dm0-:
This corrects build problems for the glibc upgrade PR.  There is still an NFSv4 test failure to investigate, and an upgrade test on vmware is needed.

The glibc PR commits are cherry-picked onto this to be able to test it all in the same ref; they can be dropped to merge the other PR once all tests are working.

Required by #3294


dongsupark:
Today I happened to look into this PR, because of the glibc upgrade issue.
If I understand the description correctly, after having upgraded nfs-utils etc, the kola's nfsv4 test started to fail. That prevents this PR from being merged.
Then is it possible split out this PR into 2 parts, e.g. upgrading glibc and dealing with nfs-related dependencies?
If that's possible, then it would become easier to make a progress.

dm0-:
If we split the NFS commits into a separate PR from this, then both would be a dependency of the glibc upgrade PR, so I don't see a benefit in doing so.

We also need to verify the vmware agent doesn't break while upgrading to an OS version with these PRs, so an upgrade test needs to be run for each CL release that has a new version of the vmware agent (since it never gets upgraded on an installed system).  This PR is stalled because it is apparently impossible to get access to a vmware environment for testing at Red Hat.

dm0-:
The NFS error turned out to be trivial, so I've fixed it.  This just needs vmware testing now.

dongsupark:
@dm0- That's good to hear. :+1: 
Sorry, I'm afraid I'm not able to help with the vmware test.

app-emulation/open-vm-tools/open-vm-tools-10.3.0.ebuild
```diff
@@ -1,26 +1,22 @@
-# Copyright 1999-2014 Gentoo Foundation
+# Copyright 1999-2018 Gentoo Foundation
```
bgilbert: Per the README, `coreos-base/oem-vmware` should also be bumped.

PR #3355, opened 2018-08-03
sys-kernel/coreos-*: fix CVE-2018-14678 for 1855

bgilbert:


dm0-:
This alpha gets promoted, so it should normally add the 4.14 patch, but I'll assume we'll have new stable kernels before promotion so it will be handled then.

bgilbert:
Right, that was my assumption.

PR #3362, opened 2018-08-09
Bump Linux 4.14.61 to 4.14.62 and 4.17.13 to 4.17.14 in alpha

dm0-:
coreos/linux#236
coreos/linux#237


dm0-:
The metadata updates don't really matter, but I've been pushing them with kernel updates just to keep it up to date.  For the alpha and beta branches this time, I've just included the metadata updates in the kernel bump commits since only the kernels change lately in release branches.  I updated the metadata as a separate step in master in case other packages get updated (but I don't think there were any this time anyway).

PR #3381, opened 2018-08-22
Bump Docker to 18.06.1 in beta

dm0-:
Applying this to beta so it gets included in the release to bake the last LTS kernel update before stable promotion.

Skipping alpha assuming there will be no disasters warranting a release before the next scheduled one.

Backports #3380


lucab:
Do you know if it contains a fix for https://github.com/coreos/bugs/issues/2497?

dm0-:
There doesn't seem to be anything that sounds relevant in the changelog for it.

dm0-:
I'll merge this to have the latest fixes queued up, but we may end up reverting to 18.03 for stable again depending on looking further into that bug.

PR #3393, opened 2018-08-27
sys-kernel/coreos-source: add ACS Overrides patch

CoRfr:
This patch provides a way to override PCIe ACS rules.

It can be useful for people that are trying to pass PCIe devices to
VMs, while these devices are not properly isolated.

By default it does nothing, unless the override is explictly enabled
through the kernel cmdline.
For instance: pcie_acs_override=downstream,multifunction

The author of the patch (Alex Williamson), said on a Fedora discussion
thread:
> The patch is not expected to be accepted upstream but downstreams can
> obviously choose to carry it.  The argument upstream is that issues that
> arise from a user overriding known, hardware advertised device isolation
> can be subtle and incredibly difficult to debug.  The path forward to
> allowing configurations that are currently prevented is to work with the
> hardware vendors to determine whether devices are isolated and encourage
> future products to support PCI ACS so that the hardware advertises this
> isolation automatically.

So at this point we are still unlikely to receive this patch from
upstream. Since the above discussion (that occured in 2014), the only
thing that has changed in offer from the community to provide patched
kernels:
- linux-vfio on arch: https://aur.archlinux.org/packages/linux-vfio/
- https://queuecumber.gitlab.io/linux-acs-override/

Usage of this patch comes with some risks and can lead to system
instability. However by default it does nothing, unless it is explicitly
enabled by the person that deploys CoreOS.

Original source: https://lkml.org/lkml/2013/5/30/513
Source of this patch: https://aur.archlinux.org/packages/linux-vfio/

Other discussions and links:
- background: https://vfio.blogspot.com/2014/08/iommu-groups-inside-and-out.html
- thread for the inclusion in Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=1113399
- ArchLinux doc: https://wiki.archlinux.org/index.php/PCI_passthrough_via_OVMF#Bypassing_the_IOMMU_groups_.28ACS_override_patch.29
- https://www.reddit.com/r/VFIO/comments/63hr88/why_is_pcie_acs_overridedownstreammultifunction/

### DIFF
--- a/sys-kernel/coreos-sources/coreos-sources-4.18.5.ebuild
+++ b/sys-kernel/coreos-sources/coreos-sources-4.18.5.ebuild
@@ -39,4 +39,5 @@ UNIPATCH_LIST="
 	${PATCH_DIR}/z0002-tools-objtool-Makefile-Don-t-fail-on-fallthrough-wit.patch \
 	${PATCH_DIR}/z0003-4.17.x-won-t-boot-due-to-x86-boot-compressed-64-Hand.patch \
 	${PATCH_DIR}/z0004-Revert-net-increase-fragment-memory-usage-limits.patch \
+	${PATCH_DIR}/z0005-add-acs-overrides.patch \
 "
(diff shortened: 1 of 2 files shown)


CoRfr:
Doesn't look this is going to get in.

PR #3394, opened 2018-08-30
sys-kernel/coreos-modules: add mlxsw drivers to default kernel options

dongsupark:
There are use cases for running the kernel on Mellanox switches. So we need to enable kernel modules for `CONFIG_MLXSW_CORE` & `CONFIG_MLXSW_SPECTRUM`, and also enable network framework, `CONFIG_NET_SWITCHDEV`.

See also https://www.kernel.org/doc/Documentation/networking/switchdev.txt,
https://github.com/mellanox/mlxsw/wiki.


blixtra:
FYI, the particular devices we're needing support for are the Mellanox SN2100 and SN2700.

PR #3404, opened 2018-09-06
coreos-base/oem-gce: avoid dep loops with oslogin

ajeddeloh:
Set DefaultDependencies=false for oem-gce-enable-oslogin.service to
avoid dependency loops.


ajeddeloh:
Wasn't sure if I needed to bump the revision or not. It looks like it gets rebuilt regardless every time?

PR #3414, opened 2018-09-18
oem-gce: enable oslogin after tmpfiles

ajeddeloh:
Ensure that tmpfiles has run before enabling oslogin since it relies on
some of the files written out by tmpfiles.


ajeddeloh:
Tested this with kola to ensure it doesn't cause races with openssh. Looked good. Proper testing of the actual oslogin bit to come a little later.

PR #3456, opened 2018-10-26
sys-kernel/bootengine: bump to fix AWS/GCE disks

r7vme:
This change includes bootengine with fix for
AWS/GCE disks in ignition.

https://github.com/coreos/bootengine/pull/149

Related bug: https://github.com/coreos/bugs/issues/2481


lucab:
@dm0- do we need to introduce any dependency here between `bootengine` and `nmve-cli` ebuilds?

dm0-:
Yeah, I guess that would be a good idea.  Right now `nvme-cli` is only brought it by the overall `coreos` metapackage, so in theory, the kernel ebuild could try to generate an initramfs before it's installed.  (In practice, the kernel is usually one of the last packages to build because the modules package takes so long.)  So, specifically I think the dependency could be added on the two `sys-kernel/coreos-kernel` ebuilds.

r7vme:
Seems both `sys-kernel/coreos-kernel` already have bootengine dependencies

https://github.com/coreos/coreos-overlay/blob/7e5f5b43b719a7de039709f8eee20172ec407d1a/sys-kernel/coreos-kernel/coreos-kernel-4.14.78.ebuild#L30
https://github.com/coreos/coreos-overlay/blob/7e5f5b43b719a7de039709f8eee20172ec407d1a/sys-kernel/coreos-kernel/coreos-kernel-4.19.0-r1.ebuild#L30

Is it what we were looking for?

r7vme:
PTAL

Added `sys-apps/nvme-cli` as dependency to both `sys-kernel/coreos-kernel` ebuilds.

lucab:
No further comments from my side, @dm0- feel free to merge whenever suitable. Thanks a lot @r7vme for pushing this forward.

PR #3458, opened 2018-10-29
Drop our Rust bootstrapping customizations

dm0-:
We are currently bootstrapping Rust std libraries for each board to support cross-compiling our Rust projects (with wrappers handling the correct toolchain `ld`, `ar`, etc.) for prod images.  Now that we only support an amd64 board and our version of portage understands EAPI 7, we can use the host `rustc` and its std libraries to build the board packages, allowing the removal of our custom process.  Future Rust upgrades should be able to just sync `dev-lang/rust`, `dev-util/cargo`, `virtual/rust`, and `virtual/cargo` in portage-stable.


cgwalters:
Why does this change the `coreos-kernel` ebuild too?

cgwalters:
Sorry, missed the split out first commit.  I don't have enough expertise in that bit to review, but the rest looks sane!

dm0-:
The first commit is because I refreshed the metadata for all packages after syncing eclasses to catch any changed global dependencies and such.  The kernel metadata was affected because it wasn't updated in #3457 (but it doesn't actually touch the ebuild again).

coreos-base/update-ssh-keys/update-ssh-keys-9999.ebuild
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2017 CoreOS, Inc.. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=6
+EAPI=7
```
bgilbert: Do these changes want revision bumps for clarity?

dm0-: I had kept the versions since the package contents should not have changed (unless we had some different settings between the host and cross-compiler linkers or something like that), but they're updated now.

PR #3490, opened 2018-11-27
coreos-base/oem-hcloud: add hetzner cloud support

dedene:
Along with:
- https://github.com/coreos/coreos-metadata/pull/125
- https://github.com/coreos/ignition/pull/667
- https://github.com/coreos/scripts/pull/859
- https://github.com/coreos/container-linux-config-transpiler/pull/164
- https://github.com/coreos/bootengine/pull/150

This will greatly enhance working with CoreOS on the Hetzner Cloud.


dedene:
@LKaemmerling Would you mind taking a look at this from the Hetzner Cloud team? Feel free to suggest changes.

coreos-base/oem-hcloud/files/base/default.ign
```diff
@@ -0,0 +1,14 @@
+{
```
bgilbert: This file can just be dropped.

coreos-base/oem-hcloud/files/base/base.ign
```diff
@@ -0,0 +1,22 @@
...
+      {
+        "name": "coreos-metadata-sshkeys@.service",
+        "enabled": true
+      },
+      {
+        "name": "coreos-metadata.service",
+        "enabled": true
+      },
+      {
+        "contents": "[Unit]\nDescription=Setting the hostname from the metadata\n\n[Service]\nType=oneshot\nEnvironment=COREOS_METADATA_OPT_PROVIDER=--cmdline\nExecStart=/usr/bin/coreos-metadata ${COREOS_METADATA_OPT_PROVIDER} --hostname=/etc/hostname\n\n[Install]\nRequiredBy=multi-user.target",
```
bgilbert: Is it necessary to use coreos-metadata to set the hostname?  (Is it not set by DHCP?)

This service is being added as a dependency of a service that's approached late, and the service has no other dependencies, so there's no guarantee it'll do anything useful.  Other platforms that use coreos-metadata to set the hostname do it from the initramfs, such as [DigitalOcean](https://github.com/coreos/bootengine/blob/master/dracut/30ignition/coreos-digitalocean-network.service) or [Packet](https://github.com/coreos/bootengine/blob/master/dracut/30ignition/coreos-static-network.service).

dedene: AFAIK it is indeed not set by DHCP. But indeed, totally agree that that is a much better approach. I did not yet add network setup support to Ignition. Can I add a PR to https://github.com/coreos/bootengine just for setting the hostname from initramfs or should I add also support for the network setup?

bgilbert: If DHCP is sufficient for network setup, it's okay to have the initramfs only set the hostname.

dedene: I created a PR to add support in coreos/bootengine to set the hostname from the initramfs (https://github.com/coreos/bootengine/pull/150) and removed the empty default ignition config and the part related to the hostname from the base ignition config.

bgilbert:
LGTM.

bgilbert:
At this point Container Linux is in deep maintenance mode and we're not merging updates of this kind. I'd recommend that you consider contributing Hetzner Cloud support to Fedora CoreOS instead. Thanks for the PR.

PR #3512, opened 2018-12-12
Bump Linux 4.14.81 to 4.14.84 in stable

dm0-:


bgilbert:
Note for the record: cache was regenerated from scratch.

PR #3523, opened 2018-12-18
Modernize Go dependencies

dm0-:
Make all of our Go packages use the host Go compilers, so Go is no longer needlessly cross-compiled (three times in our current setup).

This skips Docker-related packages since they are only built for torcx, which ignores all dependencies.  The only Go packages in portage-stable that can't benefit from this change are `godep` and `glide`, which would only be installed on the SDK host anyway and aren't even installed by default.


app-emulation/rkt/rkt-9999.ebuild
```diff
@@ -79,15 +79,17 @@ function add_stage1() {
 }
 
 src_prepare() {
-	# ensure we use a CoreOS PXE image version that matches rkt's expectations.
-	local rkt_coreos_version
+	eapply_user
```
ajeddeloh: was this previously in `autotools-utils_src_prepare`?

dm0-: Yes, and it's required to call it since EAPI 6.

PR #3542, opened 2019-01-20
Updates to sync profiles

dm0-:
This changes our base upstream profile due to it being deprecated.

Part of coreos/portage-stable#709


dm0-:
The profile rebase disables `gdbm` support by default.  It was only built for `perl` and `man-db`, and nothing seems to depend on that, so we can try leaving it disabled unless someone hits issues from it.

PR #3661, opened 2019-05-02
sys-kernel/coreos-modules: Enable XDP sockets

dm0-:
Fixes coreos/bugs#2580


bgilbert:
Looks like this bumps -sources instead of -kernel.

dm0-:
Oops, wrong tab completion with no spare SDK or Jenkins to test it yet.  Updated.

PR #3708, opened 2019-07-10
sys-kernel/coreos-sources: fix overlayfs Docker breakage in 4.19.56+

bgilbert:
https://github.com/coreos/linux/pull/339


bgilbert:
Passed two kola runs on all platforms.

PR #3714, opened 2019-07-26
Upgrade Linux 4.19.60 to 4.19.61

coreosbot:
coreos/linux#344


ajeddeloh:
Tests fail due to ENOSPACE, clearing workspaces and restarting

PR #3719, opened 2019-08-01
coreos-devel/mantle: bump to v0.13.0

ajeddeloh:


coreos-devel/mantle/mantle-9999.ebuild
```diff
@@ -11,7 +11,7 @@ if [[ "${PV}" == 9999 ]]; then
 	COREOS_GO_MOD="vendor"
```
dm0-: I don't think the new version is going to build unless you move this variable to global scope (see the previous commit).

ajeddeloh: whoops. you're right. fixed.

PR #3729, opened 2019-08-12
oem-exoscale: thorough overhaul (version bumped to 0.0.6)

exo-cedric:
Hello CoreOS team,

We have gone through Exoscale OEM machinery again and addressed a few issues to further improve users experience:
- hostname is now properly set
- allow password login/reset
- use 169.254.169.254 as canonical datasource

We believe those changes are ready for inclusion upstream.

Thanks for your consideration and best regards


lucab:
@exo-cedric thanks for the update. For reference, in the meanwhile we have been trying to consolidate this kind all logic across all supported clouds and to share it with other distros as well. The result of that effort is https://github.com/coreos/afterburn, and I think it could be worth trying to fold most of this logic over there.

bgilbert:
The standard architecture for maintained OEMs is: [Ignition](https://github.com/coreos/ignition) fetches and applies userdata, and Afterburn running in the initramfs handles any non-DHCP-based network configuration.  coreos-cloudinit is deprecated, but existing OEMs can configure a default Ignition config which falls back to coreos-cloudinit if no Ignition config is provided.  So the preferred next step is to add Exoscale support to Ignition and Afterburn.

Since the Exoscale OEM is community-supported, we'd be willing to consider merging updates to it.  In general, though, CoreOS Container Linux is in maintenance mode and is receiving almost no new functionality.  Its successor is [Fedora CoreOS](https://github.com/coreos/fedora-coreos-tracker/), and you may prefer to focus on adding Exoscale support there instead.  Fedora CoreOS also uses Ignition and Afterburn.

exo-cedric:
Thank you for your quick coming back to us.
We will definitely reroute our efforts into Ignition and Afterburn, which will require a new development cycle.
Would you consider merging the current PR anyway ?

bgilbert:
At this point Container Linux is in deep maintenance mode and we're not merging updates of this kind.  I'd recommend that you consider contributing Exoscale support to Fedora CoreOS instead.  Thanks for the PR.

PR #3730, opened 2019-08-14
profiles: mask old binutils

ajeddeloh:
binutils < 2.32.0 cannot read files handled by binutils >= 2.32.0. Mask
the old one to prevent build failures.

untested atm


ajeddeloh:
Doesn't work

PR #3733, opened 2019-08-16
Upgrade Linux in beta to 4.19.67

coreosbot:
coreos/linux#351


bgilbert:
2191 will soon promote to stable, so we shouldn't take this now.

PR #3746, opened 2019-08-30
profiles/oem-aci: provide old python

ajeddeloh:
I don't know why this works. Any version older than the 2.7 version we
use works. Portage appears to be doing the opposite of what it should
be.


bgilbert:
Commit message needs a description of what this is fixing.

I'm okay with this if we also add a kola test to catch any reoccurrence.

ajeddeloh:
https://github.com/coreos/mantle/pull/1051

ajeddeloh:
updated commit message

bgilbert:
Addresses https://github.com/coreos/bugs/issues/2608.

PR #3747, opened 2019-08-30
[build-2191] profiles/oem-aci: provide old python

ajeddeloh:
This ensures the gce agent ACI has python. For unknown reasons including
python-3.6.5 in package.provided causes both python 3.x and 2.x to be
removed but we still need 2.x.

I don't know why this works. Any version older than the 2.7 version we
use works. Portage appears to be doing the opposite of what it should
be.


bgilbert:
Addresses https://github.com/coreos/bugs/issues/2608.

PR #3779, opened 2019-10-11
Bump etcd to 3.3.17

bgilbert:


bgilbert:
https://jenkins-container-linux.apps.ci.centos.org/job/os/job/manifest/356/cldsv/

arithx:
>is quite gnarly for other reasons.

Ah the classic golang dump when a `SIGQUIT` hits it :)

We can always bump the time limit on packet runs but I'm not sure there's much value if they're taking longer than 4 hours per run.

PR #3785, opened 2019-10-17
bump go to 1.13, pin rkt to 1.12

ajeddeloh:
https://jenkins-container-linux.apps.ci.centos.org/job/os/job/manifest/366/cldsv/


ajeddeloh:
Tests looks good, failures are unrelated, do/packet only, normal flakes.

PR #3786, opened 2019-11-04
port the CFS fixes forward for build 2247

jhohertz:
Sorry if this doesn't match expectations for PRs, but I wanted to put this out there for consideration. This is against the build-2247 branch.

This is a patch to address coreos/bugs#2623 , which is in turn to address kubernetes/kubernetes#67577

Long and short of it, is that there is a CFS bug that will result in high levels of throttling on kubernetes clusters unless this patch is in place.

Please note I am still in the midst of testing this build, however I have been running a similar patch against build-2135 for some time now, with great results.

Note there is also an effort to get this patch into the 4.19.x stable kernel tree, however this seems to be stalled at the moment. See the kubernetes issue for more details on that.


ajeddeloh:
This PR looks good, except we want to run this through the normal promotion process. Can you PR against `master` instead and we'll get an alpha out with it soon. Also, can you include 763a9ec06c409dcde2a761aac4bb83ff3938e0b3 as well?

jhohertz:
I will try and get this figured out @ajeddeloh , thanks. :)

bgilbert:
@jhohertz Please do PR the patches to coreos/linux as well, and base the PR on top of https://github.com/coreos/linux/pull/363 which should land soon.

Thanks for the PR!  We know our kernel bump process isn't exactly trivial.  :slightly_smiling_face:

jhohertz:
New WIP PRs here:

* coreos/linux/pull/364
* coreos/coreos-overlay#3790

bgilbert:
Closing in favor of https://github.com/coreos/coreos-overlay/pull/3790.

PR #3832, opened 2019-12-12
sys-apps/ignition: bump to v0.34.0

ajeddeloh:


sys-apps/ignition/ignition-9999.ebuild
```diff
@@ -40,11 +40,6 @@ RDEPEND="
 
 RDEPEND+="${DEPEND}"
 
-PATCHES=(
-	"${FILESDIR}/0001-providers-allow-FetchConfig-to-mutate-the-fetcher.patch"
```
lucab: I guess we should remove these two files from git as well, right?

ajeddeloh: Good catch, done

PR #3879, opened 2020-02-11
Fix name resolution in GCE OEM container

pothos:
- coreos-base/oem-gce: Add metadata.google.internal to hosts:
    When a custom DNS server is used
    coreos-metadata-sshkeys@core.service
    fails to resolve metadata.google.internal
    because only "metadata" is specified in /etc/hosts.
- coreos-base/oem-gce: Bind mount nsswitch.conf for /etc/hosts:
    The host's /etc/nsswitch.conf is a symlink to
    /usr/share/google-oslogin/nsswitch.conf
    but that is not present in the rkt container.
    Do not only bind-mount /etc but also the target
    of the symlink. With a broken nsswitch.conf
    any entries in /etc/hosts are not considered
    which makes problems when a custom DNS server
    is used.

Closes https://github.com/coreos/bugs/issues/2601


bgilbert:
Thanks for sending.  Please remember to bump the ebuild revision.

pothos:
Done

bgilbert:
https://jenkins-container-linux.apps.ci.centos.org/job/os/job/manifest/747/cldsv/

bgilbert:
Better late than never!  Apologies for the wait, and thanks for sending.

PR #3896, opened 2020-03-05
Upgrade Linux in alpha to 4.19.108

coreosbot:
coreos/linux#393


bgilbert:
106 -> 108 because the 107 run targeted the wrong alpha.

PR #3948, opened 2020-05-17
Update etcd to 3.3.20

bgilbert:


bgilbert:
https://jenkins-container-linux.apps.ci.centos.org/job/os/job/manifest/745/cldsv/

bgilbert:
It turns out that etcd 3.3.20 isn't available from Quay in ACI format, so this will need to happen after the switch to docker images.

